PR #16, opened 2026-03-09
docs: fix README Quick Example code errors

### LINKED ISSUE #15: [docs] README Quick Example has two errors that prevent it from running
**Describe the bug**
The Quick Example code in README.md has two issues that cause errors when copied and run directly:

1. **Connection order error**: Calling `db.connect()` before `load_builtin_dataset()` causes "There is already a read-write connection constructed" error
2. **Result access error**: `record['a.fName']` uses dict-style access, but the actual return is a list and should be accessed by index

**Execution Logs**
Issue 1 - Connection order error:
```python
import neug

db = neug.Database("./database") 
conn = db.connect()  # Create connection first

db.load_builtin_dataset("tinysnb")  # Error: internally calls connect() again
```

Error output:
```
RuntimeError: Transaction state conflict: There is already a read-write connection constructed. 
at /Users/neug/neug/src/main/connection_manager.cc:48 func: CreateConnection, Error Code: 4001
```

Issue 2 - Result access error:
```python
for record in result:
    print(f"{record['a.fName']}, {record['b.fName']}, {record['c.fName']} are mutual friends")
```

Error output:
```
TypeError: list indices must be integers or slices, not str
```

**Expected behavior**
The example code in README should be copy-paste runnable without errors.

**Suggested Fix**
Suggested modification to the Quick Example:

```python
import neug

# Step 1: Load and analyze data (Embedded Mode)
db = neug.Database("/path/to/database") 

# Load sample data (must load data before connect)
db.load_builtin_dataset("tinysnb")

# Then create connection to execute queries
conn = db.connect()

# Run analytics - find triangles in the graph
result = conn.execute("""
    MATCH (a:person)-[:knows]->(b:person)-[:knows]->(c:person),
          (a)-[:knows]->(c)
    RETURN a.fName, b.fName, c.fName
""")

# Access results by index (returns a list)
for record in result:
    print(f"{record[0]}, {record[1]}, {record[2]} are mutual friends")

# Step 2: Serve applications (Service Mode)  
conn.close()
db.serve(port=8080)
```

**Environment**
- NeuG version: v0.1
- Python version: 3.12
- OS: macOS

rouxiaomin:
## Related Issues
fix #15

## What does this PR do?
Fixes two errors in the README Quick Example that prevent it from running correctly.

## What changes in this PR?
- Fix connection order: Move `db.load_builtin_dataset()` before `db.connect()` to avoid "already a read-write connection" error
- Fix result access: Change `record['a.fName']` to `record[0]` since QueryResult returns a list for each row, not a dict

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,12 @@ import neug
 
 # Step 1: Load and analyze data (Embedded Mode)
 db = neug.Database("/path/to/database") 
-conn = db.connect()
 
-# Load sample data
+# Load sample data (must load data before creating connection)
 db.load_builtin_dataset("tinysnb")
+
+# Create connection to execute queries
+conn = db.connect()
 
 # Run analytics - find triangles in the graph
 result = conn.execute("""
@@ -48,8 +50,9 @@ result = conn.execute("""
     RETURN a.fName, b.fName, c.fName
 """)
 
+# Access results by index (QueryResult returns a list for each row)
 for record in result:
-    print(f"{record['a.fName']}, {record['b.fName']}, {record['c.fName']} are mutual friends")
+    print(f"{record[0]}, {record[1]}, {record[2]} are mutual friends")
 
 # Step 2: Serve applications (Service Mode)  
 conn.close()


longbinlai (approved):
Thank you for addressing the issue. The fix LGTM

PR #33, opened 2026-03-12
feat: Export Query Results to CSV Format File

### LINKED ISSUE #38: [Export] Export query results to CSV files
Query results need to be exported to external files in CSV format for use in Excel, data warehouses, or scripts.

- **Approach**: Query results are currently in QueryResponse PB; parse row-by-row directly from the columnar layout of QueryResponse and write to a CSV file.
- **Options**: Support `delimiter`, `header`, `quote_char`, `escape_char`, `batch_size`, etc.
- **Data types**: Support primitive, temporal, vertex, edge, path, list, and struct.

Assemble rows from the columnar storage and write, keeping consistency with the existing QueryResponse structure.

**Query Example:**
```cypher
COPY (MATCH (v:person) RETURN v.ID, v.fName, v.age) 
TO 'person.csv' (HEADER = true, DELIMITER = ',', BATCH_SIZE = 2048);
```

**Additional context**
Parent issue: Support exporting query results to external files (CSV / JSON / Parquet).

shirly121:
## Related issue number

Fixes #38

<h3>Greptile Summary</h3>

This PR replaces the previous Kùzu-derived CSV export pipeline with a new, self-contained implementation that routes all `COPY (...) TO` queries through the function catalog and a unified `ArrowCsvExportWriter`. The old `ExportWriterFactory` / `CsvExportWriter` / `ExportFuncSharedState` / `ExportFuncLocalState` machinery is removed and replaced by a single `write_exec_func_t` callback stored on `ExportFunction`, a new `WriteOptions` struct, and a `CSVStringFormatBuffer` that writes directly to an Arrow output stream in configurable batches. The default value of the `HEADER` option is also changed from `false` to `true`.

Key changes and concerns found during review:

- **Breaking proto wire-format change** — `DataExport` message fields 2/3/4 (`file_path`, `options`, `property_mappings`) are removed and replaced with fields 5/6 (`file_schema`, `entry_schema`). Any serialised plan or existing client using the old field numbers will silently produce an empty message; the old field numbers should be marked `reserved`.
- **Silent removal of `EXPORT_DATABASE` / `IMPORT_DATABASE`** — these statement types are dropped from the planner dispatch switch with no error message or deprecation notice, leaving users with undefined behaviour.
- **Fragile sink/data-export ordering** — a post-hoc `SwapElements` call is used to fix the physical plan order after the fact; this is brittle and should be addressed structurally.
- **Overly strict `COPY TO` child-operator check** — requiring the child to be `PROJECTION` or `AGGREGATE` could regress valid query shapes.
- **Possible regression for `\t` as a tab delimiter** — the new single-character validation rejects multi-character escape sequences; whether the parser converts `\t` to a real tab before reaching this code needs to be confirmed.
- **Silent error suppression** — when `ignore_errors = true`, invalid cell values are silently replaced with an empty string (a no-op write), making corrupted rows indistinguishable from rows with legitimately empty fields.
- **Test coverage gaps** — the 20+ Python integration tests only check file existence and row/header counts; no test verifies the actual serialised cell values.

<h3>Confidence Score: 2/5</h3>

- Not safe to merge — breaking proto wire-format change, silent removal of EXPORT_DATABASE/IMPORT_DATABASE, and fragile plan ordering need to be addressed first.
- Multiple high-severity issues: (1) the DataExport proto removes field numbers 2/3/4 without reserving them, making old and new clients wire-incompatible; (2) EXPORT_DATABASE and IMPORT_DATABASE statement types are silently dropped from the planner with no error; (3) the post-hoc sink/data_export swap is fragile. The core new writer logic is sound, but these structural issues prevent a safe merge.
- proto/cypher_dml.proto (breaking wire-format change), src/compiler/planner/planner.cpp (silently dropped statement types), src/compiler/gopt/g_query_converter.cpp (fragile plan reordering and overly strict child check)

| Filename | Overview |
|----------|----------|
| src/utils/writer/writer.cc | New core CSV writer implementation. Contains the CSVStringFormatBuffer and ArrowCsvExportWriter classes. A typo exists in a comment ("espace"), and the ignored-errors path silently drops values with no null marker. Overall logic is sound but these edge cases need attention. |
| proto/cypher_dml.proto | Breaking proto API change: old fields 2 (file_path), 3 (options), 4 (property_mappings) are removed and replaced by fields 5 (file_schema) and 6 (entry_schema). Old and new clients are wire-incompatible. |
| src/compiler/gopt/g_query_converter.cpp | convertCopyTo now only allows PROJECTION/AGGREGATE as COPY TO children (potential regression), and a post-hoc sink/data_export element swap was added as a fragile ordering fix. |
| src/compiler/planner/planner.cpp | EXPORT_DATABASE and IMPORT_DATABASE statement-type cases silently removed from the planner's main dispatch switch, making these statements no longer functional. |
| src/compiler/function/csv_export_function.cpp | New, clean replacement for the old export_csv_function.cpp. Correctly registers the writeExecFunc and bind function. The DELIMITER/DELIM normalisation helper properly validates single-char constraints. |
| src/execution/execute/ops/batch/data_export.cc | DataExportOpr refactored to use ExportFunction and FileSchema/EntrySchema. Holds a raw pointer to ExportFunction from the catalog; safe as long as catalog lifetime exceeds operator lifetime. |
| include/neug/utils/writer/writer.h | New header for the writer hierarchy. WriteOptions has reasonable defaults (batch_size=1024, header=true, delimiter='|'). API is clean and well-structured. |
| tools/python_bind/tests/test_export.py | Good breadth of integration tests. Tests only verify row counts and header column counts, not actual cell values. Missing content-level assertions may miss bugs in value serialization. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant User
    participant Binder
    participant Planner
    participant GQueryConvertor
    participant DataExportOpr
    participant ArrowCsvExportWriter

    User->>Binder: COPY (query) TO 'file.csv' (options)
    Binder->>Binder: getExportFunction(fileTypeInfo)<br/>→ looks up ExportCSVFunction from catalog
    Binder->>Binder: getExportFuncBindData()<br/>→ ExportFuncBindData{columnNames, fileName, options}
    Binder-->>Planner: BoundCopyTo{ExportFuncBindData, ExportFunction, query}

    Planner->>Planner: planCopyTo()<br/>→ LogicalCopyTo{bindData, exportFunc, child}

    Planner-->>GQueryConvertor: LogicalCopyTo

    GQueryConvertor->>GQueryConvertor: convertCopyTo()<br/>validates child is PROJECTION/AGGREGATE
    GQueryConvertor->>GQueryConvertor: convertDataExport()<br/>→ DataExport{file_schema, entry_schema}
    GQueryConvertor->>GQueryConvertor: post-hoc swap:<br/>ensure sink before data_export in plan
    GQueryConvertor-->>DataExportOpr: PhysicalPlan (DataExport + Sink ops)

    DataExportOpr->>DataExportOpr: Build()<br/>resolves ExportFunction* from GCatalog<br/>builds FileSchema + EntrySchema

    DataExportOpr->>DataExportOpr: Eval(ctx)<br/>calls exportFunction_->execFunc()

    DataExportOpr->>ArrowCsvExportWriter: writeExecFunc(ctx, schema, entry_schema, graph)
    ArrowCsvExportWriter->>ArrowCsvExportWriter: convertFileSchemaOptions()<br/>(DELIMITER→DELIM, validate single char)
    ArrowCsvExportWriter->>ArrowCsvExportWriter: Sink::sink_results() → QueryResponse
    ArrowCsvExportWriter->>ArrowCsvExportWriter: CSVStringFormatBuffer<br/>addHeader() + addValue() per cell
    ArrowCsvExportWriter->>ArrowCsvExportWriter: flush(stream) every batchSize rows
    ArrowCsvExportWriter-->>User: CSV file written to disk
```

<details open><summary><h3>Comments Outside Diff (22)</h3></summary>

1. `tests/storage/test_export.cc`, line 2142 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/tests/storage/test_export.cc#L2142)) 

   **Incorrect path in file-existence check**

   The condition checks for a file literally named `"c"` instead of `"/tmp/rel_a.csv"`. Because `"c"` almost never exists, the cleanup guard is effectively dead code and the stale test artifact is never removed before the test runs. This can cause the subsequent `EXPECT_TRUE` to pass spuriously when running tests in a dirty environment.

2. `src/compiler/planner/plan/plan_port_db.cpp`, line 1381 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/compiler/planner/plan/plan_port_db.cpp#L1381)) 

   **`planExportDatabase` unconditionally returns `nullptr`**

   The entire `EXPORT DATABASE` implementation was removed and replaced with `return nullptr`. Any caller that does not guard against a null `LogicalPlan*` will dereference null and crash at runtime. Even if `EXPORT DATABASE` is intentionally being disabled as part of this refactor, returning `nullptr` silently is dangerous. At minimum, the function should throw an appropriate "not implemented" or "not supported" exception so the failure is explicit and diagnosable.

3. `include/neug/utils/writer/writer.h`, line 652 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/include/neug/utils/writer/writer.h#L652)) 

   **Default for `has_header` contradicts documentation**

   `has_header` defaults to `true` in `WriteOptions`, but the documentation in `doc/source/data_io/export_data.md` explicitly documents the default as `false`:

   ```
   |`HEADER`|Whether to output a header row.|`false`|
   ```

   This discrepancy means a user who omits the `HEADER` option will get a header row in the output despite the docs saying otherwise. One of the two must be corrected.

4. `src/utils/writer/writer.cc`, line 2109 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/utils/writer/writer.cc#L2109)) 

   **Potential null dereference of `entry_schema_`**

   `ArrowCsvExportWriter` inherits from `ArrowExportWriter`, whose constructor accepts `entry_schema = nullptr` as a default. Here, `*entry_schema_` is dereferenced unconditionally — if `entry_schema_` is a null `shared_ptr`, this will crash.

   A null-check (or an assertion) should be added before dereferencing:

   ```cpp
   if (!entry_schema_) {
     return Status(StatusCode::ERR_INVALID_ARGUMENT, "entry_schema is null");
   }
   auto csvBuffer = CSVStringFormatBuffer(table, schema_, *entry_schema_);
   ```

5. `src/compiler/function/csv_export_function.cpp`, line 885 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/compiler/function/csv_export_function.cpp#L885)) 

   **`DELIMITER` silently ignored when `DELIM` is also set**

   `std::map::insert` (and its case-insensitive equivalent) does **not** overwrite an existing key. If a user specifies both `DELIM` and `DELIMITER` in their options (which may be unusual but not impossible), the value from `DELIMITER` is silently dropped and `DELIM` takes precedence without any warning or error. Consider using `operator[]` / `emplace` with `try_emplace` or logging a warning when both are supplied.

2. `tests/storage/test_export.cc`, line 2142 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/tests/storage/test_export.cc#L2142)) 

   **Truncated path in existence check**

   The file path `"/tmp/rel_a.csv"` was accidentally truncated to just `"c"` in the guard condition. This means `std::filesystem::remove("/tmp/rel_a.csv")` will never be called to clean up the stale file before the test runs — potentially causing the subsequent `EXPECT_TRUE` to operate on leftover data from a previous run.

3. `src/compiler/planner/plan/plan_port_db.cpp`, line 1346-1347 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/compiler/planner/plan/plan_port_db.cpp#L1346-L1347)) 

   **`planExportDatabase` unconditionally returns `nullptr`**

   The entire implementation of `planExportDatabase` was removed and replaced with `return nullptr`. Any caller that invokes this path (e.g., an `EXPORT DATABASE` statement) will receive a null `LogicalPlan*` and almost certainly crash with a null-pointer dereference downstream.

   If export-database support is intentionally being dropped or deferred, the function should at minimum throw an informative "not implemented" exception rather than silently returning null.

4. `tools/python_bind/tests/test_export.py`, line 2227-2236 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/tools/python_bind/tests/test_export.py#L2227-L2236)) 

   **Test name / query mismatch — queries `movies`, not `person`**

   `test_export_person_without_header` is named as if it exports `person` nodes, but both the count query and the `COPY TO` statement target `movies`. This is almost certainly a copy-paste error and will silently test the wrong label under the wrong name.

5. `src/utils/writer/writer.cc`, line 2010-2072 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/utils/writer/writer.cc#L2010-L2072)) 

   **`WriteOptions` reconstructed on every cell write**

   A new `WriteOptions` object is constructed and `schema_.options` is looked up (via `get()`) for every single cell written — once for `has_header`, once for `delimiter`, once for `ignore_errors`, once for `escape_char`, and once for `quote_char`. For large result sets this creates significant per-cell overhead.

   Consider constructing `WriteOptions` once (e.g., in the `CSVStringFormatBuffer` constructor) and caching the resolved option values as fields, rather than re-resolving them in every `addValue` call.

6. `src/utils/writer/writer.cc`, line 1910-1924 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/utils/writer/writer.cc#L1910-L1924)) 

   **`kStructArray` case missing null-validity check**

   Every other typed-array case in `formatValueToStr` calls `validateProtoValue(...)` before reading the value, but the `kStructArray` branch proceeds without any validity check:

   ```cpp
   case neug::Array::TypedArrayCase::kStructArray: {
       auto struct_arr = arr.struct_array();
       std::string list_val;
       list_val.append("[");
       for (int i = 0; i < struct_arr.fields_size(); ++i) {
           const auto& field = struct_arr.fields(i);
           ARROW_ASSIGN_OR_RAISE(auto elem, formatValueToStr(field, rowIdx));
   ```

   If `struct_arr` has a validity bitmap, this can silently produce output for null rows rather than returning an error or an empty string. Add the same validity check used by the other branches.

7. `src/compiler/function/csv_export_function.cpp`, line 88 ([link](https://github.com/alibaba/neug/blob/744892bab9fca79a3f6226b2174c3bd499dc07e6/src/compiler/function/csv_export_function.cpp#L88)) 

   **Missing newline at end of file**

   The file ends without a trailing newline (the diff shows `\ No newline at end of file`). This can cause compiler warnings and may confuse some tooling.

   Add a newline after the closing `}` of the namespace.

   <sub>Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!</sub>

8. `src/utils/writer/writer.cc`, line 317-336 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/utils/writer/writer.cc#L317-L336)) 

   **Output stream never closed after writing**

   `ArrowCsvExportWriter::writeTable` opens an `arrow::io::OutputStream` and writes all CSV data to it, but never calls `stream->Close()` before the function returns. Arrow's `LocalFileSystem` output stream uses buffered I/O, and without an explicit close the kernel buffers may not be fully flushed to disk. This means exported CSV files can be silently truncated or empty on many file systems.

9. `src/compiler/planner/plan/plan_port_db.cpp`, line 23-29 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/compiler/planner/plan/plan_port_db.cpp#L23-L29)) 

   **`EXPORT DATABASE` feature completely removed — breaking change**

   The entire `planExportDatabase` implementation has been replaced with an unconditional `THROW_NOT_SUPPORTED_EXCEPTION`. Any user who relied on the `EXPORT DATABASE` Cypher statement will receive a runtime error after this PR merges, with no migration path or deprecation notice. The previous implementation exported all tables to files and was apparently functional (it was tested via `test_load.py`).

   If this removal is intentional (e.g. because it depended on the old `ExportCSVBindData` approach), it should at minimum be documented and the related parser/binder code for `EXPORT DATABASE` should also be removed to avoid confusing compiler errors, rather than silently failing at runtime.

10. `src/compiler/gopt/g_query_converter.cpp`, line 1989-1997 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/compiler/gopt/g_query_converter.cpp#L1989-L1997)) 

    **Overly-restrictive child operator type check breaks valid query shapes**

    The new guard rejects any `COPY TO` whose direct child is not a `PROJECTION` or `AGGREGATE`, but valid Cypher queries often use `ORDER BY` (`SORT`), `LIMIT`, `DISTINCT`, or other operators as the outermost node:

    ```cypher
    COPY (MATCH (v:person) RETURN v ORDER BY v.ID LIMIT 10) TO 'out.csv'
    ```

    These will now throw a compile-time exception. The previous code made no such restriction — it only required at least one child. If the intent is to guarantee that column names are available, checking for the schema expression list (as the old code did) is a safer approach than an operator-type whitelist.

11. `src/utils/writer/writer.cc`, line 65-73 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/utils/writer/writer.cc#L65-L73)) 

    **Buffer capacity calculation can overflow `size_t`**

    ```cpp
    capacity_ = DEFAULT_CAPACITY * batchSize * response->arrays_size();
    ```

    `DEFAULT_CAPACITY` is `size_t` (64), `batchSize` is `size_t` (default 1024), and `arrays_size()` returns `int`. With the default values this is `64 * 1024 * N` which is safe for small column counts, but a user could set `BATCH_SIZE` to a very large number (e.g., `1 << 30`) and cause a silent `size_t` wrap-around, resulting in a wildly undersized allocation and subsequent heap corruption via the `memcpy` in `write()`.

    Consider saturating/clamping the capacity or using a checked multiplication:

12. `src/utils/writer/writer.cc`, line 238-250 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/utils/writer/writer.cc#L238-L250)) 

    **Header not emitted for empty result sets**

    The header row is written only inside `addValue` when `rowIdx == 0 && colIdx == 0`. If the query produces zero rows, `addValue` is never called and the output file contains no header line — even when the user explicitly passed `HEADER = true`. Many downstream tools (Pandas, Spark, etc.) expect a header-only file to be valid CSV.

    The header should be written once during initialisation (or at the start of `writeTable`) rather than lazily on the first cell.

13. `src/execution/execute/ops/batch/data_export.cc`, line 77-83 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/execution/execute/ops/batch/data_export.cc#L77-L83)) 

    **Raw pointer to catalog function — lifetime not guaranteed**

    `writeFunc` is a raw pointer obtained from the in-memory catalog:

    ```cpp
    auto func = gCatalog->getFunctionWithSignature(signatureName);
    auto writeFunc = func->ptrCast<function::ExportFunction>();
    ```

    `DataExportOpr` stores this as `function::ExportFunction* exportFunction_`. If the catalog invalidates or reloads function entries (e.g. on `LOAD EXTENSION`, schema reload, or multi-tenant scenarios), the operator will hold a dangling pointer. The pointer is then dereferenced in `Eval`, which runs later.

    Consider either storing the function by value (it's a small struct) or using a `shared_ptr`/`unique_ptr` copy so the lifetime is explicit.

14. `src/utils/writer/writer.cc`, line 195-215 ([link](https://github.com/alibaba/neug/blob/91376f558392eb126d4094a78c8923a767522a62/src/utils/writer/writer.cc#L195-L215)) 

    **`addEscapes` uses `sizeof(escape)` as a skip distance — misleading and fragile**

    ```cpp
    found = val.find(toEscape, found + sizeof(escape));
    ```

    `sizeof(escape)` for a `char` is always `1`, so the behaviour is correct: after inserting the escape prefix, the search advances past the current `toEscape` character so it isn't double-escaped. However, this looks as if the intent is to skip the newly-inserted escape character in the *output* string, which is incorrect conceptually. A future maintainer who changes `escape` to a multi-character type (e.g. `std::string`) would get a compile error or silent breakage.

15. `proto/cypher_dml.proto`, line 54-65 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/proto/cypher_dml.proto#L54-L65)) 

    **Breaking protobuf wire-format change**

    The PR removes fields `2` (`file_path`), `3` (`options`), and `4` (`property_mappings`) and introduces new fields `5` (`file_schema`) and `6` (`entry_schema`) in their place. In protobuf, field numbers are baked into the serialised wire format. Any existing client or stored plan that was serialised with the old field numbers will silently produce an empty `DataExport` message when decoded by the new server, because the old field numbers are now undefined.

    If there are any deployed services, stored query plans, or other consumers that serialise `DataExport` messages, they will break silently after this change. Consider either:
    - Keeping the old field numbers alive as `reserved` declarations so the proto compiler rejects accidental reuse, and using new numbers only for the new fields, or
    - Providing a migration path / version bump if a clean break is intentional.

    ```proto
    // Recommended: mark old numbers as reserved to prevent silent decode errors
    reserved 2, 3, 4;
    reserved "file_path", "options", "property_mappings";
    ```

16. `src/compiler/planner/planner.cpp`, line 112-118 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/src/compiler/planner/planner.cpp#L112-L118)) 

    **`EXPORT_DATABASE` / `IMPORT_DATABASE` silently broken**

    The cases for `StatementType::EXPORT_DATABASE` and `StatementType::IMPORT_DATABASE` are removed from the dispatch switch without a deprecation notice or error message. If a user submits either of those statement types, execution will fall through to whichever `default` or next-matching case exists in the switch, producing undefined or misleading behaviour instead of a clear error.

    If these features are intentionally removed in this PR, please add an explicit `default` (or specific case) that throws a `NotImplemented` / unsupported-statement exception so callers receive an actionable message:

    ```cpp
    case StatementType::EXPORT_DATABASE:
    case StatementType::IMPORT_DATABASE:
      THROW_EXCEPTION_WITH_FILE_LINE(
          "EXPORT/IMPORT DATABASE is no longer supported; "
          "use COPY (...query...) TO 'file.csv' instead.");
    ```

17. `src/compiler/gopt/g_query_converter.cpp`, line 91-97 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/src/compiler/gopt/g_query_converter.cpp#L91-L97)) 

    **Fragile post-hoc plan reordering**

    This block swaps the last two plan elements when they happen to be a `sink` followed by a `data_export`. This is a structural assumption that is easily broken: if any future operator is appended after the sink (e.g., a metrics or tracing wrapper), the swap will either not trigger or will move the wrong operators.

    A more robust fix would be to produce the correct order during `convertDataExport` itself — i.e., emit `data_export` first and then have the caller append the sink — rather than patching up the order after the fact. At minimum, please add a comment explaining *why* the swap is necessary and what invariant is being enforced.

18. `src/compiler/gopt/g_query_converter.cpp`, line 1248-1259 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/src/compiler/gopt/g_query_converter.cpp#L1248-L1259)) 

    **Overly strict child-operator check may regress valid queries**

    The new guard requires the direct child of a `COPY TO` plan node to be either `PROJECTION` or `AGGREGATE`. Previously no such restriction existed. While `RETURN` always produces one of those two operator types in practice, there could be edge cases — e.g., a COPY TO wrapping an `ORDER BY` / `LIMIT` operator, or a future optimisation rewrite that inserts a different operator — that hit this hard exception.

    Consider making this a softer warning (or removing the check entirely) unless there is a documented invariant that must hold here.

19. `src/utils/writer/writer.cc`, line 2098-2115 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/src/utils/writer/writer.cc#L2098-L2115)) 

    **Escaped character itself is never explicitly advanced past**

    The loop writes the `escape` prefix then searches for the *next* special character starting at `found + sizeof(escape)` (`found + 1`), leaving `i` still pointing at the current special character. The special character is therefore written "for free" by the next iteration's inner `while (i < found)` loop (or by the trailing `while`). This produces correct output, but the behaviour is subtle and non-obvious.

    Consider advancing `i` past the special character after writing the escape to make the intent explicit:

    ```cpp
    while (found != std::string::npos) {
      // write chars before the special char
      while (i < found) {
        write(reinterpret_cast<const uint8_t*>(&val[i]), sizeof(char));
        i++;
      }
      // write escape prefix, then the character itself
      write(reinterpret_cast<const uint8_t*>(&escape), sizeof(char));
      write(reinterpret_cast<const uint8_t*>(&val[i]), sizeof(char));
      i++;  // advance past the now-escaped character
      found = val.find_first_of(toEscape, i, 2);
    }
    ```

    Also, the comment on line 1984 reads `// espace special characters` — typo for "escape".

20. `src/utils/writer/writer.cc`, line 2161-2171 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/src/utils/writer/writer.cc#L2161-L2171)) 

    **Silent null emission on error gives no signal in the output**

    When `ignore_errors_` is `true` and `formatValueToStr` fails, the code writes `DEFAULT_NULL_STR` which is an empty string (`""`), and `strlen("") == 0`, so `write(…, 0)` is a no-op — the column value is completely absent in the output row. This makes bad rows indistinguishable from rows where the field legitimately has an empty string value.

    Consider writing a configurable null sentinel (e.g. `"NULL"` or `""`) and documenting the behaviour, or at least logging a warning when a value is silently dropped.

21. `src/compiler/function/csv_export_function.cpp`, line 912-925 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/src/compiler/function/csv_export_function.cpp#L912-L925)) 

    **`\t` tab delimiter no longer supported — potential regression**

    The old `CsvExportWriter::parse_csv_options` explicitly handled the `\t` escape sequence and converted it to a tab character. The new validation rejects any `DELIM` value whose `.size() != 1`, which means users who previously used `DELIM = '\t'` (two characters in the string if passed through as-is from the parser) will now receive an `INVALID_ARGUMENT` exception.

    If the Cypher parser already converts `'\t'` to the single tab character before it reaches this code, this is fine — but that should be verified and documented. If the string `\t` (2 bytes) can reach this function, the regression needs to be addressed.

22. `tools/python_bind/tests/test_export.py`, line 2379-2389 ([link](https://github.com/alibaba/neug/blob/fdda53661ec6ee0d475cc0a20c3b2fbc6dd321d4/tools/python_bind/tests/test_export.py#L2379-L2389)) 

    **Tests verify row/column counts but not actual cell values**

    All 20+ tests follow the same pattern: run a `COPY TO`, parse the CSV, and assert header length and row count. No test checks that the written values match what the graph actually contains. This means bugs in value serialisation (wrong encoding for dates, timestamps, floats, complex types, etc.) would go undetected.

    Consider adding at least one end-to-end test that reads back specific known values, for example exporting a small fixed dataset and asserting the exact CSV content.

</details>

<sub>Last reviewed commit: fdda536</sub>

### DIFF
--- a/include/neug/utils/spinlock.h
+++ b/include/neug/utils/spinlock.h
@@ -26,13 +26,10 @@ class SpinLock {
 
  public:
   void lock() {
-    while (locked.test_and_set(std::memory_order_acquire)) {
-      {
-      }
-    }
+    while (locked.test_and_set(std::memory_order_acquire)) {}
   }
 
-  void unlock() { locked.clear(std::memory_order_relaxed); }
+  void unlock() { locked.clear(std::memory_order_release); }
 };
 
 }  // namespace neug


shirly121:
@Greptile fixed in https://github.com/alibaba/neug/pull/33/changes/91376f558392eb126d4094a78c8923a767522a62

include/neug/utils/reader/schema.h
```diff
@@ -39,6 +39,7 @@ struct EntrySchema {
   virtual ~EntrySchema() = default;
   virtual EntrySchemaType type() const = 0;
   std::vector<std::string> columnNames;
+  // todo: support vertex, edge and path types
```
zhanglei1949: Resolved?

src/utils/writer/writer.cc
```diff
@@ -0,0 +1,393 @@
...
+    blob_.size = copy_len;
+    blob_.data = std::move(new_data);
+    data_ = blob_.data.get();
+  }
+
+  memcpy(data_ + blob_.size, buffer, len);
+  blob_.size += len;
+}
+
+void CSVStringFormatBuffer::addValue(int rowIdx, int colIdx) {
```
zhanglei1949, line 296: Both Arrow and neug query result are based on columnar format, so could we process one array at one time? rather than rowId +colId?

include/neug/utils/writer/writer.h
```diff
@@ -0,0 +1,163 @@
...
+
+class CSVStringFormatBuffer : public StringFormatBuffer {
+ public:
+  CSVStringFormatBuffer(const neug::QueryResponse* response,
+                        const reader::FileSchema& schema,
+                        const reader::EntrySchema& entry_schema);
+  ~CSVStringFormatBuffer() {}
+  void addValue(int rowIdx, int colIdx) override;
+  void addHeader();
+  arrow::Status flush(std::shared_ptr<arrow::io::OutputStream> stream) override;
```
zhanglei1949: - Status
- result<..>

shirly121: > * Status
> * result<..>

updated in https://github.com/alibaba/neug/pull/33/changes/d7dfcf8409a75fe2c9a9181830ca99d93f3c3d02

include/neug/utils/reader/schema.h
shirly121: the comment has been removed

longbinlai (approved):
LGTM

PR #79, opened 2026-03-18
feat: add row_num interface to `IRecordBatchSupplier`

liulx20:
Fixes

<h3>Greptile Summary</h3>

This PR adds a `row_num()` virtual method to `IRecordBatchSupplier` and all its implementations so that `insert_vertices_impl` can pre-allocate indexer capacity upfront rather than growing it incrementally per batch. The approach is sound in principle, but several correctness and performance issues remain across the implementation.

Key findings:

- **Capacity guard changed from `>=` to `>`** (`vertex_table.h` lines 281 and 304): the original per-batch check used `>=`, meaning it grew capacity before reaching the limit; the new code uses `>`, which skips growth when `new_size == capacity`. If the `IndexerType` does not permit `size == capacity`, this is an off-by-one that will overrun the allocated region.
- **`new_size` variable shadowing** (`vertex_table.h` line 303): the same name is declared twice in the same function — once at function scope and once inside the `while` loop — triggering `-Wshadow` warnings and reducing readability.
- **Double file read in `CSVStreamRecordBatchSupplier`** (`loader_utils.cc`): the file is opened and fully scanned a second time just to count rows, doubling I/O for CSV loading.
- **Scanner scanned twice in `batch_read`** (`reader.cc`): `CountRows()` performs a full pass before `ToRecordBatchReader()` starts a second pass; for non-Parquet sources this doubles read time.
- **`row_num()` counting failure is now a hard error, not a hint** (`loader_utils.cc`, `reader.cc`): both paths throw on any `CountRows` failure, meaning a transient I/O error during counting aborts the entire load even though the underlying data could still be read.
- **`SupplierWrapperWithFirstBatch::row_num()` omits `first_batch_` rows** (`loader_utils.h` line 115): when a `first_batch_` is provided, its rows are not added to the total, causing under-allocation.
- **`row_num_` not initialized in constructor initializer list** (`loader_utils.cc`): `CSVStreamRecordBatchSupplier` does not initialize `row_num_` in the initializer list; its value is indeterminate until the successful assignment at line 261.
- **`nullptr` passed as `arrow::StopToken`** (`loader_utils.cc` line 255): `StopToken` is a value type; relying on implicit conversion from `nullptr` may not compile across all Arrow versions. Prefer `arrow::StopToken{}`.
- **`GeneratedRecordBatchSupplier::row_num()` returns a shrinking count** (`tests/unittest/utils.h`): because `GetNextBatch()` pops from `batches_`, `row_num()` returns progressively smaller values after each consumed batch, which does not match the expected "total row count" contract.

<h3>Confidence Score: 1/5</h3>

- This PR is not safe to merge — multiple correctness issues (off-by-one in capacity guard, uninitialized member, incorrect row counts) combined with doubled I/O on every CSV and Parquet load need to be resolved first.
- The capacity guard condition change from `>=` to `>` is a potential off-by-one that could silently corrupt data or crash for indexers that don't permit `size == capacity`. Row-count failures are now hard errors rather than soft hints, making loads less resilient. The double file/scanner read significantly degrades performance. Several `row_num()` implementations return incorrect or unstable values (`first_batch_` exclusion, decreasing test count). Together these represent high-risk correctness and stability regressions on the core data-loading path.
- `include/neug/storages/graph/vertex_table.h` (off-by-one capacity guard, variable shadowing), `src/storages/loader/loader_utils.cc` (double file read, uninitialized member, hard failure on count), `src/utils/reader/reader.cc` (double scanner scan, hard failure on count).

| Filename | Overview |
|----------|----------|
| include/neug/storages/graph/vertex_table.h | Adds upfront capacity pre-allocation using `row_num()`; introduces variable shadowing of `new_size` and changes the capacity guard condition from `>=` to `>`, risking an off-by-one overrun. |
| include/neug/storages/loader/loader_utils.h | Adds `row_num()` to the `IRecordBatchSupplier` interface and all implementations; `SupplierWrapperWithFirstBatch` omits `first_batch_` rows from the count, `ArrowRecordBatchStreamSupplier` constructor signature is a breaking change requiring callers to pass `row_num`, and the `int64_t` return is implicitly narrowed from `size_t` in `ArrowRecordBatchArraySupplier`. |
| src/storages/loader/loader_utils.cc | Implements row counting in `CSVStreamRecordBatchSupplier` by opening and fully reading the CSV file a second time; `row_num_` is not in the constructor initializer list, and `nullptr` is passed for the `arrow::StopToken` value parameter. |
| src/utils/reader/reader.cc | Calls `scanner->CountRows()` (full scan) before `scanner->ToRecordBatchReader()` (second full scan), doubling I/O; the log says "Proceeding without row count" but immediately throws, making row count a hard requirement instead of a hint. |
| tests/unittest/utils.h | Implements `row_num()` by summing over the mutable `batches_` vector; because `GetNextBatch()` pops elements, repeated calls to `row_num()` after consumption return a shrinking, incorrect total. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant VT as insert_vertices_impl
    participant S as IRecordBatchSupplier
    participant CSV as CSVStreamRecordBatchSupplier
    participant FS as FileSystem

    VT->>S: row_num()
    S-->>VT: int64_t (pre-allocation hint)
    VT->>VT: new_size = indexer_.size() + row_nums
    alt new_size > capacity
        VT->>VT: grow cap (×1.25 loop)
        VT->>VT: EnsureCapacity(cap)
    end

    loop GetNextBatch()
        VT->>S: GetNextBatch()
        S-->>VT: RecordBatch (or nullptr)
        VT->>VT: per-batch fallback capacity check (new_size > capacity)
        VT->>VT: insert_primary_keys + set_properties_column
    end

    note over CSV,FS: CSVStreamRecordBatchSupplier constructor
    CSV->>FS: Open(file_path) → StreamingReader file
    CSV->>FS: Open(file_path) again → count file
    CSV->>FS: CountRowsAsync (full read)
    FS-->>CSV: row_num_
    CSV->>FS: StreamingReader::Make (for actual reads)
```

<details open><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `src/storages/loader/loader_utils.cc`, line 242-265 ([link](https://github.com/alibaba/neug/blob/90db25befae385c4de5dabb3dba37c731c9241a2/src/storages/loader/loader_utils.cc#L242-L265)) 

   <a href="#">P0</a> **`row_num_` left uninitialized on counting failure**

   The member `row_num_` is not initialized in the constructor's initializer list, and there are two code paths where it is never assigned a value:
   1. When `count_file_result.ok()` returns `false` (the outer `else` branch).
   2. When `count_result.ok()` returns `false` (the inner `else` branch).

   In both cases, `row_num_` retains an indeterminate value (undefined behavior in C++). The `LOG(WARNING)` message even says *"Proceeding with row_num_=0"* — but `row_num_` is never actually set to `0`.

   This means the subsequently computed `new_size` in `vertex_table.h` could be wildly large, causing either an out-of-memory `EnsureCapacity` call or overflowing `new_size`.

   Fix: initialize `row_num_` to `0` in the constructor initializer list:

2. `include/neug/storages/loader/odps_fragment_loader.h`, line 109-151 ([link](https://github.com/alibaba/neug/blob/99ab2d8c050791802ca4a1a6220ebdc81cddf1af/include/neug/storages/loader/odps_fragment_loader.h#L109-L151)) 

   <a href="#">P0</a> **ODPS suppliers missing `row_num()` — compile break**

   `row_num()` is now declared `pure virtual` in `IRecordBatchSupplier`, but neither `ODPSStreamRecordBatchSupplier` (line 109) nor `ODPSTableRecordBatchSupplier` (line 133) implement it. Both classes are therefore abstract and cannot be instantiated — any attempt to do so will fail at compile time.

   `ODPSTableRecordBatchSupplier` owns a `table_` member (`std::shared_ptr<arrow::Table>`), so a trivial fix is:
   ```cpp
   int64_t row_num() const override { return table_->num_rows(); }
   ```

   `ODPSStreamRecordBatchSupplier` does not load the table upfront, so a sensible fallback is:
   ```cpp
   int64_t row_num() const override { return 0; }  // row count not available for streaming ODPS supplier
   ```

   Both implementations must be added before this PR can compile.

3. `include/neug/storages/graph/vertex_table.h`, line 281-304 ([link](https://github.com/alibaba/neug/blob/ce826f1d6697111fdce6761f62461ead9416c3e1/include/neug/storages/graph/vertex_table.h#L281-L304)) 

   **Capacity guard condition changed from `>=` to `>` — potential off-by-one**

   The original per-batch guard used `if (new_size >= indexer_.capacity())`, meaning: "if inserting this batch would fill **or exceed** current capacity, grow first." The new pre-allocation block (line 281) and the retained per-batch fallback (line 304) both use `>`, meaning they only grow when the new size would *exceed* current capacity.

   If the underlying `IndexerType` treats `size == capacity` as full (i.e. requires strictly fewer elements than the allocated capacity), this change introduces an off-by-one: a batch that exactly fills remaining capacity will not trigger a resize, and the subsequent insert will overrun the allocated region. The previous `>=` guard was correct because it eagerly grew capacity before reaching the limit.

   Please verify that the indexer permits `size == capacity` without undefined behaviour. If it does not, restore `>=`:

   

   …and similarly on line 304:

   ```cpp
         if (batch_new_size >= indexer_.capacity()) {
   ```

</details>

<sub>Last reviewed commit: ["Add capacity checkin..."](https://github.com/alibaba/neug/commit/ce826f1d6697111fdce6761f62461ead9416c3e1)</sub>

### DIFF
--- a/src/execution/common/operators/retrieve/path_expand.cc
+++ b/src/execution/common/operators/retrieve/path_expand.cc
@@ -27,206 +27,213 @@ neug::result<Context> PathExpand::edge_expand_v(
     const StorageReadInterface& graph, Context&& ctx,
     const PathExpandParams& params) {
   std::vector<size_t> shuffle_offset;
+
   if (params.labels.size() == 1 &&
-      ctx.get(params.start_tag)->column_type() == ContextColumnType::kVertex &&
-      std::dynamic_pointer_cast<IVertexColumn>(ctx.get(params.start_tag))
-              ->vertex_column_type() == VertexColumnType::kSingle) {
-    auto& input_vertex_list =
-        *std::dynamic_pointer_cast<SLVertexColumn>(ctx.get(params.start_tag));
-    auto pair = path_expand_vertex_without_predicate_impl(
-        graph, input_vertex_list, params.labels, params.dir, params.hop_lower,
-        params.hop_upper);
-    ctx.set_with_reshuffle(params.alias, pair.first, pair.second);
-    return ctx;
-  } else {
-    if (params.dir == Direction::kOut) {
-      auto& input_vertex_list =
-          *std::dynamic_pointer_cast<IVertexColumn>(ctx.get(params.start_tag));
-      std::set<label_t> labels;
-      std::vector<std::vector<LabelTriplet>> out_labels_map(
-          graph.schema().vertex_label_frontier());
-      for (const auto& label : params.labels) {
-        labels.emplace(label.dst_label);
-        if (params.hop_lower == 0) {
-          labels.emplace(label.src_label);
-        }
-        out_labels_map[label.src_label].emplace_back(label);
+      params.labels[0].src_label == params.labels[0].dst_label &&
+      ctx.get(params.start_tag)->column_type() == ContextColumnType::kVertex) {
+    auto vertex_col =
+        dynamic_cast<const IVertexColumn*>(ctx.get(params.start_tag).get());
+    if (vertex_col->vertex_column_type() == VertexColumnType::kSingle) {
+      const auto& input_vertex_list =
+          dynamic_cast<const SLVertexColumn&>(*ctx.get(params.start_tag));
+      if (input_vertex_list.label() == params.labels[0].src_label) {
+        auto pair = path_expand_vertex_without_predicate_impl(
+            graph, input_vertex_list, params.labels, params.dir,
+            params.hop_lower, params.hop_upper);
+        ctx.set_with_reshuffle(params.alias, pair.first, pair.second);
+        return ctx;
       }
-
-      MLVertexColumnBuilderOpt builder(labels);
-      std::vector<std::tuple<label_t, vid_t, size_t>> input;
-      std::vector<std::tuple<label_t, vid_t, size_t>> output;
-      foreach_vertex(input_vertex_list,
-                     [&](size_t index, label_t label, vid_t v) {
-                       output.emplace_back(label, v, index);
-                     });
-      int depth = 0;
-      while (depth < params.hop_upper && (!output.empty())) {
-        input.clear();
-        std::swap(input, output);
-        if (depth >= params.hop_lower) {
-          for (auto& tuple : input) {
-            builder.push_back_vertex({std::get<0>(tuple), std::get<1>(tuple)});
-            shuffle_offset.push_back(std::get<2>(tuple));
-          }
-        }
-
-        if (depth + 1 >= params.hop_upper) {
-          break;
-        }
-
-        for (auto& tuple : input) {
-          auto label = std::get<0>(tuple);
-          auto v = std::get<1>(tuple);
-          auto index = std::get<2>(tuple);
-          for (const auto& label_triplet : out_labels_map[label]) {
-            auto oe_view = graph.GetGenericOutgoingGraphView(
-                label_triplet.src_label, label_triplet.dst_label,
-                label_triplet.edge_label);
-            auto oes = oe_view.get_edges(v);
-            for (auto it = oes.begin(); it != oes.end(); ++it) {
-              output.emplace_back(label_triplet.dst_label, it.get_vertex(),
-                                  index);
-            }
-          }
-        }
-        ++depth;
-      }
-      ctx.set_with_reshuffle(params.alias, builder.finish(), shuffle_offset);
-      return ctx;
-    } else if (params.dir == Direction::kIn) {
-      auto& input_vertex_list =
-          *std::dynamic_pointer_cast<IVertexColumn>(ctx.get(params.start_tag));
-      std::set<label_t> labels;
-      std::vector<std::vector<LabelTriplet>> in_labels_map(
-          graph.schema().vertex_label_frontier());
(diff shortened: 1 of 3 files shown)


include/neug/storages/loader/loader_utils.h
```diff
@@ -78,6 +78,7 @@ class IRecordBatchSupplier {
  public:
   virtual ~IRecordBatchSupplier() = default;
   virtual std::shared_ptr<arrow::RecordBatch> GetNextBatch() = 0;
+  virtual int64_t row_num() const = 0;
```
zhanglei1949: What about use `NumRows` as the method name

include/neug/storages/loader/loader_utils.h
```diff
@@ -166,6 +180,16 @@ class ArrowRecordBatchArraySupplier : public IRecordBatchSupplier {
 
   std::shared_ptr<arrow::RecordBatch> GetNextBatch() override;
 
+  int64_t row_num() const override {
```
zhanglei1949: Same

PR #117, opened 2026-03-24
refactor: Remove dummy method  `copy_to_tmp` for Column and Table 

### LINKED ISSUE #116: Method `copy_to_tmp` could be deleted since not used anymore
The method is not used currently.
https://github.com/alibaba/neug/blob/main/include/neug/utils/property/column.h#L62

`init` is also not used.

zhanglei1949:
Fix #116 

We copy files under checkpoint to tmp_dir in a lazy manner implemented by ensure_writable, and `copy_to_tmp` are deprecated but seems not deleted.

Related issue: #118

### DIFF
--- a/include/neug/utils/property/column.h
+++ b/include/neug/utils/property/column.h
@@ -59,8 +59,6 @@ class ColumnBase {
 
   virtual size_t size() const = 0;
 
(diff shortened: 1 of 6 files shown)


liulx20 (approved):

lnfjpt (approved):
LGTM

PR #209, opened 2026-04-10
fix: Maintain the edge count in the CSR

### LINKED ISSUE #200: It may be necessary to maintain the EdgeNum counter to improve the efficiency of statistics updates
Currently, obtaining the edge count requires traversing all edges.
```cpp
  size_t edge_num() const override {
    size_t res = 0;
    auto* adj_lists =
        reinterpret_cast<const nbr_t* const*>(adj_list_buffer_->GetData());
    auto* degrees =
        reinterpret_cast<const std::atomic<int>*>(adj_list_size_->GetData());
    for (size_t i = 0; i < vertex_capacity(); ++i) {
      auto begin = adj_lists[i];
      for (int j = 0; j < degrees[i].load(); ++j) {
        if (begin[j].timestamp.load() !=
            std::numeric_limits<timestamp_t>::max()) {
          res++;
        }
      }
    }
    return res;
  }

```

liulx20:
Fixes

### DIFF
--- a/include/neug/storages/csr/immutable_csr.h
+++ b/include/neug/storages/csr/immutable_csr.h
@@ -58,27 +58,7 @@ class ImmutableCsr : public TypedCsrBase<EDATA_T> {
                : 0;
   }
 
-  size_t edge_num() const override {
-    size_t ret = 0;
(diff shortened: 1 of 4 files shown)


zhanglei1949 (approved):
LGTM

PR #231, opened 2026-04-15
refactor: Refactor serialization for VertexTimeStamp

### LINKED ISSUE #232: No need to dump vertex_timestamp to two files
Previously we dump VertexTimeStamp to `.ts` and `.meta`. However, one file is ok, since we use Inarchive to store the data.

zhanglei1949:
Fix #232 
Also remove some dummy methods.


lnfjpt (approved):
LGTM

PR #234, opened 2026-04-16
fix: Temporarily disable the multiprocessing test case

### LINKED ISSUE #233: [BUG] multiple processes cann't access the same database simultaneously
The documentation now states that the database can be opened simultaneously by multiple processes or threads in read-only mode. However, the current reality is that concurrent access to the same database by multiple threads or processes is unsafe.
Image
Even when opened in read-only mode, operations such as dump or close still modify the database state.

For instance, a dump operation modifies the snapshot and deletes certain files.

```cpp
void PropertyGraph::Dump(bool reopen) {
  // First dump to the  temp dir, then move to the checkpoint dir
  std::string target_dir = temp_checkpoint_dir(work_dir_);
  if (std::filesystem::exists(target_dir)) {
    remove_directory(target_dir);
  } else {
    std::filesystem::create_directories(target_dir);
  }

  std::error_code errorCode;
  std::filesystem::create_directories(target_dir, errorCode);
  if (errorCode) {
    std::stringstream ss;
    ss << "Failed to create snapshot directory: " << target_dir << ", "
       << errorCode.message();
    LOG(ERROR) << ss.str();
    THROW_RUNTIME_ERROR(ss.str());
  }
```

liulx20:
Fixes

### DIFF
--- a/tools/python_bind/tests/test_db_init.py
+++ b/tools/python_bind/tests/test_db_init.py
@@ -276,6 +276,7 @@ def open_db(db_path, mode):
     db.close()
 
 
+@pytest.mark.skip(reason="https://github.com/alibaba/neug/issues/233")
 def test_open_with_multiple_process(tmp_path):
     db_dir = tmp_path / "multi_process_db"
     shutil.rmtree(db_dir, ignore_errors=True)  # Ensure clean state


zhanglei1949 (approved):
LGTM

PR #256, opened 2026-04-20
refactor: Disable `compact_on_close` by default

### LINKED ISSUE #255: Disable compact_on_close by default
Proposal: Default compact_on_close to False

Enabling compaction on close introduces unnecessary overhead when closing the database, which is especially costly for large databases.

A more fine-grained alternative would be to track whether any insertions or updates have been made and trigger compaction only when necessary. However, acquiring and maintaining such state is non-trivial in our current architecture. Therefore, disabling compact_on_close by default is the more practical choice.

Note:

- This configuration is NEVER exposed to USER(via the Python client).
- Compaction will still be triggered during WAL ingestion, prior to encountering an updated WAL segment.

zhanglei1949:
Fix #255

### DIFF
--- a/include/neug/config.h.in
+++ b/include/neug/config.h.in
@@ -98,7 +98,7 @@ struct NeugDBConfig {
         enable_auto_compaction(false),
         memory_level(MemoryLevel::kInMemory),
         compact_csr(true),
-        compact_on_close(true),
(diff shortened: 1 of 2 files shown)


zhanglei1949:
closed since not planned.

PR #258, opened 2026-04-20
fix: adding MD5 checksums for string column

### LINKED ISSUE #251: The md5 for items buffer are not calculated for StringColumn
We may skip the md5 calculation for data_array, but we should calculate the md5 for the items buffer.

liulx20:
Fix #251 
This pull request improves data integrity for string columns by adding MD5 checksums to the output files in the `TypedColumn<std::string_view>` class. The main changes ensure that both the data and item files now include MD5 hashes, which can be used for later verification.


zhanglei1949 (approved):
LGTM

PR #261, opened 2026-04-21
test: Add test to verify memory growth is bounded

### LINKED ISSUE #202: [BUG] Service-mode `serve()` / `stop_serving()` cycles cause near-linear RSS growth and retain about 1 GB after `close()`
**Describe the bug**
Repeatedly starting and stopping the NeuG service on the same in-memory `Database` object grows process RSS almost linearly, even when no queries are executed. The memory does not return near baseline after `db.close()`, `gc.collect()`, or `malloc_trim(0)`.

**Execution Logs**
The script below reproduces the issue and records RSS before `close()`, after `close()`, and after `malloc_trim(0)`.

```python

import ctypes
import gc
import time

def rss_mb() -> float:
    with open("/proc/self/status", "r", encoding="utf-8") as f:
        for line in f:
            if line.startswith("VmRSS:"):
                return round(int(line.split()[1]) / 1024.0, 2)
    raise RuntimeError("VmRSS not found in /proc/self/status")


libc = ctypes.CDLL("libc.so.6")
base = rss_mb()
db = Database("", "w")

for i in range(1, 301):
    db.serve(17000 + (i % 20), "127.0.0.1", False)
    db.stop_serving()

before_close = rss_mb()
db.close()
gc.collect()
time.sleep(1)
after_close = rss_mb()
libc.malloc_trim(0)
time.sleep(1)
after_trim = rss_mb()

print(base, before_close, after_close, after_trim)
```

In a longer 10-minute run of the same workflow, RSS continued to grow almost linearly for the full duration.

**Expected behavior**
Starting and stopping the service without workload should not produce large persistent RSS growth. After `db.close()`, memory should return close to baseline apart from small allocator noise.

**Error Message**
Observed measurements from reproduced runs:

```text
100 cycles: 61.62 MB -> 97.32 MB
after db.close(): 97.32 MB
after malloc_trim(0): 91.29 MB

300 cycles: 61.61 MB -> 109.56 MB
after db.close(): 106.19 MB
after malloc_trim(0): 100.63 MB
```

10-minute run of the same restart loop:

```text
baseline RSS: 70.41 MB
before close: 1205.90 MB
after db.close(): 1104.29 MB
after malloc_trim(0): 1095.01 MB
total growth before close: +1135.49 MB
retained above baseline after trim: +1024.60 MB
elapsed time: 603.71 s
restart cycles completed: 52025
average RSS growth: about 112.85 MB/min
last-quarter growth rate: about 102.57 MB/min
linear fit on sampled RSS curve: R^2 = 0.99957
```

zhanglei1949:
Fix #202

### DIFF
--- a/tools/python_bind/tests/test_db_init.py
+++ b/tools/python_bind/tests/test_db_init.py
@@ -437,7 +437,9 @@ def test_memory_growth(tmp_path):
 
     ITERATIONS = 100
     PORT_BASE = 10020  # avoid collision with other tests' ports
-    TOLERANCE_MB = 15  # empirical: ~7.82 MB observed over 100 rounds; 15 MB covers jitter
+    TOLERANCE_MB = (
+        15  # empirical: ~7.82 MB observed over 100 rounds; 15 MB covers jitter
+    )
 
     def rss_mb() -> float:
         ru = resource.getrusage(resource.RUSAGE_SELF)
--- a/tools/python_bind/tests/test_db_init.py
+++ b/tools/python_bind/tests/test_db_init.py
@@ -418,3 +418,57 @@ def test_memory_level_invalid(tmp_path):
     with pytest.raises(Exception) as excinfo:
         Database(db_path=str(db_dir), mode="w", buffer_strategy="invalid_level")
         assert str(ERR_INVALID_ARGUMENT) in str(excinfo.value)
+
+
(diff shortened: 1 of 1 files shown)


zhanglei1949:
Closed, as these tests are unrelated to the root cause.

PR #276, opened 2026-04-22
fix: Fix Bugs in Duplicated Column Projections

### LINKED ISSUE #243: [BUG] Valid node-property parameter predicate fails with `_Map_base::at` when ordered by projected alias
**Describe the bug**
Hi, we ran into a query-level internal error that looks like a real engine bug rather than a syntax problem.

A valid query with a node-property parameter predicate works in simpler nearby shapes, but fails once the result is ordered by a projected alias. On the reproduced build, the query returns:

```text
Error code: 1012
Error Message: ERR_INTERNAL_ERROR: _Map_base::at
```

This is not a process crash in our local repro. The Python process and embedded database exit normally; the failure is at query level.

**Reproduction**
The complete minimal reproducer script is:

```python

import shutil
import tempfile
from pathlib import Path

from neug import Database


def run_query(conn, query: str, *, parameters: dict | None = None) -> list[list]:
    return list(conn.execute(query, parameters=parameters))


def main() -> None:
    db_path = Path(tempfile.mkdtemp(prefix="neug_bug_order_alias_"))
    print(f"Using database path: {db_path}", flush=True)

    db = Database(str(db_path), mode="w")
    conn = db.connect()

    try:
        run_query(conn, "CREATE NODE TABLE L0(id INT64, p0_2 INT64, PRIMARY KEY(id))")
        run_query(conn, "CREATE (:L0 {id: 1, p0_2: 643})")

        controls = [
            (
                "MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id",
                {"v": 643},
            ),
            (
                "MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id ORDER BY n.id LIMIT 100",
                {"v": 643},
            ),
            (
                "MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id AS node_id, n.id AS selected_id",
                {"v": 643},
            ),
        ]
        for query, parameters in controls:
            print("CONTROL>", query, flush=True)
            print(list(conn.execute(query, parameters=parameters)), flush=True)

        failing_query = (
            "MATCH (n:L0) "
            "WHERE n.p0_2 = $v "
            "RETURN n.id AS node_id, n.id AS selected_id "
            "ORDER BY node_id LIMIT 100"
        )
        print("TARGET>", failing_query, flush=True)
        print(list(conn.execute(failing_query, parameters={"v": 643})), flush=True)
    finally:
        try:
            conn.close()
        except Exception:
            pass
        try:
            db.close()
        except Exception:
            pass
        shutil.rmtree(db_path, ignore_errors=True)


if __name__ == "__main__":
    main()
```

You can also run the locally archived reproducer directly from our archive copy:

```bash
python reproduce.py
```

**Expected behavior**
The target query should return one row:

```text
[[1, 1]]
```

**Actual behavior**
The controls above succeed, but the target query fails with:

```text
Error code: 1012
ERR_INTERNAL_ERROR: _Map_base::at
```

Observed trigger boundary from local confirmation:

- succeeds:
  - `MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id`
  - `MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id ORDER BY n.id LIMIT 100`
  - `MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id AS node_id, n.id AS selected_id`
- fails:
  - `MATCH (n:L0) WHERE n.p0_2 = $v RETURN n.id AS node_id, n.id AS selected_id ORDER BY node_id LIMIT 100`

So this seems narrower than “all node-property predicates fail”: it currently looks like the combination of

- node-property parameter predicate
- duplicated projection from the same source expression under aliases
- `ORDER BY` on the projected alias

is what triggers the internal error.

**Notes**
We checked whether this could be promoted to a crash bug by running the same target query repeatedly in a subprocess loop. In our local confirmation, it stayed a stable query-level internal error and did not turn into a process abort.

If it helps, I can also provide the exact local archive bundle and the narrower control matrix we used while shrinking the trigger shape.

shirly121:
Committed-by: Xiaoli Zhou from Dev container

## Related issue number

Fixes #243


longbinlai (approved):
LGTM

PR #280, opened 2026-04-22
fix: Support List Expression to Fix Bugs in Unwind

### LINKED ISSUE #279: UNWIND fails on mixed list literals (lowered to ToTuple)
**Describe the bug**

`UNWIND` on a Cypher list literal with mixed element kinds, such as `[n.id, 12, n.id]`, does not work as a list. The plan lowers that expression to `ToTuple` instead of a list, so the `unfold` (UNWIND) path cannot treat it as iterable the same way as a homogeneous list.

**Execution logs**

1. Cypher

```cypher
MATCH (n:person)
UNWIND [n.id, 12, n.id] AS id
RETURN id
```

2. Python

```python
import neug

db = neug.Database("/tmp/modern_graph")
conn = db.connect()

res = conn.execute("""
MATCH (n:person)
UNWIND [n.id, 12, n.id] AS id
RETURN id
""")
```

**What we see in the plan (summary)**

Under `unfold` → `input_expr`, the implementation uses `to_tuple` with `fields` for the property, the constant `12`, and the property again. That matches a tuple-like shape, not a list, which explains why `UNWIND` cannot step through it as a list. (A full plan dump is available in the thread if needed; omitted here to avoid noise.)

**Expected behavior**

`UNWIND` should iterate the three list elements in order, binding each value to `id`, same as for a list literal with uniform element types.

**Parent issue**

#122 (link as sub-issue when created).

shirly121:
Committed-by: Xiaoli Zhou from Dev container

Committed-by: Xiaoli Zhou from Dev container

## Related issue number

Fixes #279


longbinlai (dismissed):
LGTM

shirly121:
@copilot resolve the merge conflicts in this pull request

lnfjpt (approved):

PR #283, opened 2026-04-23
fix: disable resizing during the INSERT transaction

### LINKED ISSUE #282: [BUG] Can't resizing during the INSERT transaction.
Currently, string columns trigger a resize during an insert transaction, which can cause the system to core dump.
```cpp
  void set_any(size_t idx, const Property& value, bool insert_safe) override {
    if (idx >= size_) {
      THROW_RUNTIME_ERROR("Index out of range");
    }
    auto dst_value = value.as_string_view();
    if (pos_.load() + dst_value.size() > data_buffer_->GetDataSize()) {
      size_t new_avg_width = (pos_.load() + idx) / (idx + 1);
      size_t new_len =
          std::max(size_ * new_avg_width, pos_.load() + dst_value.size());
      data_buffer_->Resize(new_len);
    }
    set_value(idx, dst_value);
  }
```
Additionally, there is no need to set default parameters at every layer of the interface.
```cpp
class PropertyGraph {
  Status AddVertex(label_t label, const Property& id,
                   const std::vector<Property>& props, vid_t& vid,
                   timestamp_t ts, bool insert_safe = false);
};
class VertexTable {
  bool AddVertex(const Property& id, const std::vector<Property>& props,
                 vid_t& vid, timestamp_t ts = 0, bool insert_safe = false)
};
class Table {
  void insert(size_t index, const std::vector<Property>& values,
              bool insert_safe = false);
};
class ColumnBase {
  virtual void set_any(size_t index, const Property& value,
                       bool insert_safe = false);
};
```

liulx20:
This pull request refactors several method signatures in the storage layer to require explicit specification of the `insert_safe` parameter, removing its default value. This change enforces clarity and consistency when inserting or updating data in tables and columns, and ensures that all call sites and tests explicitly specify whether insert operations should be performed safely (with buffer resizing or checks). Additionally, error handling is improved for unsafe insertions when buffer space is insufficient.

Fixes

### DIFF
--- a/tests/unittest/test_indexer.cc
+++ b/tests/unittest/test_indexer.cc
@@ -126,7 +126,7 @@ TEST_F(LFIndexerTest, SupportsCoreMutableInterfacesInMemory) {
   EXPECT_GE(indexer.capacity(), 8U);
 
   std::vector<int64_t> values = {7, 11, 13, 17, 19, 23, 29, 31, 37, 41};
-  EXPECT_EQ(indexer.insert(Property::from_int64(values[0])), 0U);
+  EXPECT_EQ(indexer.insert(Property::from_int64(values[0]), false), 0U);
   for (size_t i = 1; i < values.size(); ++i) {
     EXPECT_EQ(indexer.insert(Property::from_int64(values[i]), true),
               static_cast<uint32_t>(i));
@@ -278,7 +278,7 @@ TEST_F(LFIndexerTest, VarcharReserveEnablesNonSafeInsert) {
   std::vector<std::string> values = {"alpha",   "beta", "gamma", "delta",
                                      "epsilon", "zeta", "eta",   "theta"};
   for (const auto& v : values) {
-    indexer.insert(Property::from_string_view(v));
+    indexer.insert(Property::from_string_view(v), false);
   }
   ExpectStringValues(indexer, values);
   indexer.close();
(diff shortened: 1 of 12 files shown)


tests/unittest/test_indexer.cc
```diff
@@ -577,7 +587,7 @@ TEST_F(LFIndexerTest, VarcharShortDumpReopenReserveThenInsertLong_SyncToFile) {
     long_values.push_back(std::string(30, static_cast<char>('p' + i)));
   }
   for (const auto& v : long_values) {
-    indexer.insert(Property::from_string_view(v));
+    indexer.insert(Property::from_string_view(v), true);
```
liulx20, line 590: needs resize here? @zhanglei1949

tests/unittest/test_indexer.cc
```diff
@@ -482,7 +490,7 @@ TEST_F(LFIndexerTest, VarcharShortDumpReopenReserveThenInsertLong_InMemory) {
     long_values.push_back(std::string(60, static_cast<char>('d' + i)));
   }
   for (const auto& v : long_values) {
-    indexer.insert(Property::from_string_view(v));
+    indexer.insert(Property::from_string_view(v), true);
```
liulx20, line 493: needs resize here? @zhanglei1949

zhanglei1949: Yes.

tests/unittest/test_indexer.cc
zhanglei1949: Yes

zhanglei1949 (approved):
LGTM

PR #289, opened 2026-04-28
refactor: refactors the graph schema modification interfaces 

### LINKED ISSUE #288: Refactor some APIs to use config structs instead of long positional signatures
We currently rely on long positional parameter signatures for schema mutation APIs (for example, create edge type and related methods). 
```cpp
Status StorageTPUpdateInterface::CreateEdgeType(
    const std::string& src_type, const std::string& dst_type,
    const std::string& edge_type,
    const std::vector<std::tuple<DataType, std::string, Property>>& properties,
    bool error_on_conflict, EdgeStrategy oe_edge_strategy,
    EdgeStrategy ie_edge_strategy)
```
The same signature pattern is repeated across multiple layers, and this is becoming costly to maintain, adding one new option requires touching many places.

Proposed solution
Adopt config structs for schema mutation operations, for example:

CreateVertexTypeConfig
CreateEdgeTypeConfig
Add/Rename/Delete property operation configs

```cpp
  virtual Status CreateEdgeType(const CreateEdgeTypeConfig& config, bool error_on_conflict);
```

liulx20:
This pull request refactors the graph schema modification interfaces to use strongly-typed configuration objects for all vertex and edge type/property operations, replacing the previous approach that relied on multiple primitive parameters. This makes the API more consistent, extensible, and easier to use, especially for complex schema changes. It also updates documentation and usage examples to reflect the new builder-based configuration pattern.


include/neug/storages/graph/operation_configs.h
```diff
@@ -0,0 +1,541 @@
...
+#include <tuple>
+#include <vector>
+#include "neug/utils/property/property.h"
+#include "neug/utils/property/types.h"
+
+namespace neug {
+class InArchive;
+class OutArchive;
+
+class CreateVertexTypeConfig {
```
zhanglei1949: Maybe naming like `CreateVertexTypeParam` is better? Config sounds like a configuration, but the entity here is just a wrapper of all required params.

include/neug/storages/graph/operation_configs.h
```diff
@@ -0,0 +1,541 @@
...
+
+  void Serialize(InArchive& arc) const;
+  static CreateVertexTypeConfig Deserialize(OutArchive& arc);
+};
+
+class CreateVertexTypeConfigBuilder {
+ public:
+  CreateVertexTypeConfig config;
+  CreateVertexTypeConfigBuilder() = default;
+  CreateVertexTypeConfigBuilder& WithVertexTypeName(
```
zhanglei1949: For the builders, we could consider dropping the `with` prefix from all methods, following the same convention as OkHttpClient.

zhanglei1949 (approved):
LGTM

PR #309, opened 2026-05-06
fix: Support Returning Edge Properties Immediately After Edge Creation

### LINKED ISSUE #318: [Bug] Segfault when accessing edge properties after CREATE edge
**Describe the bug**
Accessing edge properties immediately after `CREATE` causes a segmentation fault.

**Reproduce**

```cypher
MATCH (u1:User {name: 'Adam'}), (u2:User {name: 'marko'})
CREATE (u1)-[e:follows {date: 2012}]->(u2)
RETURN e.date;
```

**Root cause**
In the `create_edge` path, edge properties are not stored after `AddEdge`. The property pointer (`prop_ptr`) is set to `nullptr`, causing a segfault when subsequent operations try to access edge properties.

```cpp
// TODO: store edge properties
builder.push_back_opt(v1.vid_, v2.vid_, nullptr);  // nullptr causes segfault on property access
```

**Expected behavior**
Edge properties should be persisted and accessible immediately after creation within the same query.

liulx20:
This pull request refactors the edge insertion API throughout the codebase to return additional information when adding edges. Instead of returning only a status or offset, the relevant `AddEdge` and `put_edge` methods now return a pair containing the previous offset and a pointer to the edge data. This change provides more context to the caller and enables more advanced use cases, such as direct access to the newly inserted edge's data.


include/neug/transaction/insert_transaction.h
```diff
@@ -194,10 +194,12 @@ class StorageTPInsertInterface : public StorageInsertInterface {
...
-  inline bool AddEdge(label_t src_label, vid_t src, label_t dst_label,
-                      vid_t dst, label_t edge_label,
-                      const std::vector<Property>& properties) override {
-    return txn_.AddEdge(src_label, src, dst_label, dst, edge_label, properties);
+  inline const void* AddEdge(label_t src_label, vid_t src, label_t dst_label,
+                             vid_t dst, label_t edge_label,
+                             const std::vector<Property>& properties) override {
+    bool ok =
+        txn_.AddEdge(src_label, src, dst_label, dst, edge_label, properties);
+    return ok ? reinterpret_cast<const void*>(1) : nullptr;
```
zhanglei1949: What is the expected behavior for dealing with the returned reinterpret_cast<const void*>(1)

include/neug/storages/graph/graph_interface.h
```diff
@@ -521,9 +521,9 @@ class StorageUpdateInterface : public StorageReadInterface,
   virtual bool AddVertex(label_t label, const Property& id,
                          const std::vector<Property>& props,
                          vid_t& vid) override = 0;
-  virtual bool AddEdge(label_t src_label, vid_t src, label_t dst_label,
-                       vid_t dst, label_t edge_label,
-                       const std::vector<Property>& properties) override = 0;
+  virtual const void* AddEdge(
```
zhanglei1949: Is there any better choice to wrap the returned edge, rather than returning a implementation-specific void* pointer?

include/neug/transaction/insert_transaction.h
liulx20: Placeholder, cann't attempt to read this property value during insert-transaction.

tools/python_bind/tests/test_db_query.py
```diff
@@ -642,6 +642,33 @@ def test_insert_edge(tmp_path):
...
+def test_create_edge_return_edge_property(tmp_path):
+    db_dir = tmp_path / "create_edge_return_edge_property"
+    shutil.rmtree(db_dir, ignore_errors=True)
+    db_dir.mkdir()
+    db = Database(db_path=str(db_dir), mode="w")
+    conn = db.connect()
+
+    conn.execute("CREATE NODE TABLE person(id INT64, PRIMARY KEY(id));")
+    conn.execute(
+        "CREATE REL TABLE knows(FROM person TO person, since INT64, MANY_MANY);"
```
zhanglei1949: should be `MANT_TO_MANY`

zhanglei1949 (approved):
LGTM

PR #312, opened 2026-05-06
feat: Support `MERGE` Clause to Ensure Pattern Existence

### LINKED ISSUE #164: [BUG] Valid MERGE query fails in planner with unsupported alias-management operator 31
**Describe the bug**
NeuG accepts `MERGE` syntax and plans the query as an update, but a valid `MERGE` query fails during planning with:

`Unsupported operator type for alias management: 31`

Statement delta reduced the original relationship `MERGE` reproducer to a single-node `MERGE`, so this does not appear to be specific to relationship writes.

**Minimal reproducer**

```python

import shutil
import tempfile
from pathlib import Path

from neug import Database


def main() -> None:
    db_path = Path(tempfile.gettempdir()) / "neug_bug_004"
    shutil.rmtree(db_path, ignore_errors=True)

    db = Database(str(db_path), mode="w")
    conn = db.connect()

    setup = [
        ("CREATE NODE TABLE L0(id INT64, k0 INT64, PRIMARY KEY(id));", "schema"),
    ]

    for idx, (query, mode) in enumerate(setup, start=1):
        print(f"STEP {idx}: {query}")
        conn.execute(query, access_mode=mode)

    query = "MERGE (:L0 {id: 2})"
    print(f"STEP {len(setup) + 1}: {query}")
    conn.execute(query, access_mode="update")


if __name__ == "__main__":
    main()
```

**Observed output**

```text
STEP 2: MERGE (:L0 {id: 2})
RuntimeError: Failed to execute query: MERGE (:L0 {id: 2}). Error code: 3000, Error Message: Unsupported operator type for alias management: 31 at /project/src/compiler/gopt/g_alias_manager.cpp:121 func: extractSingleOpGAliasNames, Error Code: 9999
```

**Expected behavior**
The `MERGE` query is valid and should execute successfully.

**Why this looks like a bug instead of an unsupported feature**
- `MERGE` is part of the parser and planner pipeline.
- The error is not a parser or binder rejection.
- Operator type `31` maps to `MERGE`, so the failure happens because planner alias-management does not handle the `MERGE` logical operator.

shirly121:
## Related issue number

Fixes #164


src/execution/execute/ops/insert/merge_edge.cc
```diff
@@ -0,0 +1,374 @@
...
+      merged[it->second].second = std::move(o.second);
+    } else {
+      pos[o.first] = merged.size();
+      merged.push_back(std::move(o));
+    }
+  }
+  return merged;
+}
+
+static Property value_to_edge_property(const Value& value) {
```
liulx20: This interface appears to overlap with the value_to_property functionality and should likely be merged

shirly121: > This interface appears to overlap with the value_to_property functionality and should likely be merged

fixed in https://github.com/alibaba/neug/pull/312/changes/bfeafae3e0746d3d20c0007df0c39548eb26288c

liulx20 (approved):

PR #331, opened 2026-05-08
fix: Ensure the atomicity of insert transaction

### LINKED ISSUE #315: [Bug] Insert transaction: no rollback for partial vertex insertions during commit
**Describe the bug**
During the commit phase of `insert_transaction`, the main write operations to the graph are performed. When inserting multiple vertices, if the first vertex is inserted successfully (`graph.AddVertex` returns success) but a subsequent vertex insertion fails (returns false or throws an exception), the already-committed vertex is not rolled back.

**Expected behavior**
All vertex insertions within a single transaction should be atomic — either all succeed or all are rolled back.

**Root cause**
No rollback mechanism is implemented for partially completed `AddVertex` operations during the commit phase.

zhanglei1949:
Fix #315

### DIFF
--- a/include/neug/transaction/transaction_utils.h
+++ b/include/neug/transaction/transaction_utils.h
@@ -39,7 +39,9 @@ enum class OpType : uint8_t {
   kDeleteVertexProp = 12,
   kDeleteEdgeProp = 13,
   kDeleteVertexType = 14,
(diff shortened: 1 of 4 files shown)


zhanglei1949:
Will be resolved by #338

PR #332, opened 2026-05-09
fix: Fix the performance issue of query ic5.

### LINKED ISSUE #333: The main branch exhibits significant performance degradation in LDBC interactive.
On the SF100 dataset, throughput is limited to approximately 20,000 QPS.

liulx20:
This pull request updates the logic in the `tc_fusable` function to correctly handle the structure of the select operator's predicate. The main change is adjusting the expected number and positions of operators in the predicate, ensuring the function matches the updated query plan structure.

Predicate structure validation:

* Changed the check for the number of operators in `select_opr.predicate()` from 7 to 3, and updated the indices used to access `var`, `within`, and `v_set` accordingly. This ensures compatibility with the new predicate structure in the query plan.

Fixes


lnfjpt (approved):
LGTM

PR #336, opened 2026-05-09
fix: fix bugs of copy to after load from

### LINKED ISSUE #334: [BUG] COPY (LOAD FROM ... RETURN ...) TO fails with 'Child of CopyTo should be a projection or aggregate'
**Describe the bug**

`COPY (LOAD FROM ... RETURN ...) TO` fails when the LOAD FROM subquery produces a result that is not directly a projection or aggregate node. The compiler's `convertCopyTo` function expects the child of CopyTo to be a projection or aggregate, but when using LOAD FROM with a RETURN clause inside COPY TO, the plan structure does not meet this expectation.

**Execution Logs**

```cypher
COPY (LOAD FROM "/path/to/vPerson.json" RETURN fName, age) TO '/tmp/export_array.json';
```

**Error Message**

```
Failed to execute query: COPY (LOAD FROM ".../vPerson.json" RETURN fName, age) TO '/tmp/export_array.json';
Error code: 3000
Error Message: ERR_COMPILATION: Child of CopyTo should be a projection or aggregate
  at /project/src/compiler/gopt/g_query_converter.cpp:2061 func: convertCopyTo, Error Code: 9999
```

**Expected behavior**

`COPY (LOAD FROM ... RETURN ...) TO` should work correctly, allowing users to export LOAD FROM results directly to a file.

shirly121:
## Related issue number

Fixes #334


zhanglei1949 (approved):
LGTM

PR #344, opened 2026-05-12
refactor: Use Value as the underlying type for default value store in Schema

### LINKED ISSUE #339: Default value held in Schema should be of type Value, not Property
https://github.com/alibaba/neug/blob/ddf509ba9e61702aea4047f5de6cc84f6f9083c5/include/neug/storages/graph/schema.h#L172
Currently, we held default value for Vertex/Edge Property as `Property`. However, it should be value.

zhanglei1949:
Fix #339 

Previously we use Property + std::string to hold the default value.


include/neug/storages/graph/operation_params.h
```diff
@@ -28,14 +29,14 @@ class OutArchive;
 class CreateVertexTypeParam {
  private:
   std::string vertex_label_name;
-  std::vector<std::tuple<DataType, std::string, Property>> properties;
+  std::vector<std::tuple<DataType, std::string, execution::Value>> properties;
```
liulx20: Is it necessary to store dataType twice here？

include/neug/execution/common/types/value.h
```diff
@@ -633,6 +633,7 @@ bool Value::ApplyComparisonOp(const Value& lhs, const Value& rhs) {
 
 Property value_to_property(const Value& value);
 Value property_to_value(const Property& property);
+Value property_to_value(const Property& property, const DataType& type);
```
liulx20: why is this interface need here

include/neug/storages/graph/operation_params.h
liulx20: The Value itself contains type information.

zhanglei1949: Well, maybe we should remove the `DataType` field.

include/neug/execution/common/types/value.h
zhanglei1949: Deleted after discussion.

liulx20 (approved):

PR #358, opened 2026-05-15
fix: Avoid copying during edge filtering

### LINKED ISSUE #357: Edge loading in BatchAddEdges buffers entire dataset in memory and can cause OOM on large graphs
BatchAddEdges currently accumulates all edge endpoints and property batches in memory before inserting into CSR/Table, which causes high peak memory usage and may lead to OOM on large datasets.
```cpp
  std::vector<vid_t> src_lid, dst_lid;
  std::vector<std::shared_ptr<arrow::RecordBatch>> data_batches;
  while (true) {
    auto batch = supplier->GetNextBatch();
    if (batch == nullptr) {
      break;
    }
    auto src_array = batch->column(0);
    auto dst_array = batch->column(1);
    parse_endpoint_column(src_indexer, src_array, src_lid);
    parse_endpoint_column(dst_indexer, dst_array, dst_lid);
    if (batch->num_columns() > 2) {
      batch = batch->RemoveColumn(0).ValueOrDie()->RemoveColumn(0).ValueOrDie();
      data_batches.push_back(batch);
    }
  }
```

liulx20:
Fixes


zhanglei1949 (approved):
LGTM

PR #383, opened 2026-05-20
fix: disable libunwind in `glog`

### LINKED ISSUE #382: [BUG] try/catch fails to catch exceptions — process aborts due to libunwind/libgcc_s _Unwind_* symbol conflict
C++ `try/catch` blocks fail to catch standard exceptions (e.g., `IOException` thrown by `MMapContainer::Resize` on mmap failure). Instead of being caught, the exception causes the process to abort immediately. The root cause is a **symbol conflict between `libunwind.so.8` and `libgcc_s.so.1`** — both export `_Unwind_*` symbols, and the dynamic linker resolves to `libunwind.so.8` (which cannot dispatch C++ exceptions via `__gxx_personality_v0`), so stack unwinding aborts before reaching the matching catch block.

## To Reproduce

1. Build neug with brpc (bthread tracer enabled, which links `-lunwind -lunwind-x86_64`) and glog (which also links `unwind::unwind`).
2. Trigger a scenario where `MMapContainer::Resize` fails (e.g., allocate a very large capacity that exhausts virtual memory).
3. Observe that the process aborts instead of the `catch (const std::exception& e)` block in `VertexTable::insert_vertices_impl` firing.

Minimal reproduction path:
```
VertexTable::insert_vertices_impl<long>
  → EnsureCapacity(large_cap)
    → indexer_.reserve() / table_->resize() / v_ts_.Reserve()
      → buffer_->Resize(huge_size)
        → MMapContainer::Resize()
          → mmap(...) returns MAP_FAILED
          → THROW_IO_EXCEPTION(...)   // exception thrown here
          → unwinder resolves to libunwind.so.8, can't find C++ personality → abort()
```

Log output immediately before crash:
```
I20260520 19:18:13.837088 59558 mmap_container.cc:111] mmap returned address: 0xffffffffffffffff
```

Stack trace:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7de9859 in __GI_abort () at abort.c:79
#2  0x00007fff8e045212 in ?? () from /lib/x86_64-linux-gnu/libunwind.so.8
#3  0x00007fff8e7fd546 in neug::VertexTable::EnsureCapacity (this=0x160cf90, capacity=<optimized out>)
    at /usr/include/c++/13/bits/allocator.h:184
#4  0x00007fff8ed1f7dd in neug::VertexTable::insert_vertices_impl<long> (...)
    at /data/0423/neug/include/neug/storages/graph/vertex_table.h:289
#5  0x00007fff8ed1ca04 in neug::VertexTable::insert_vertices (...)
    at /usr/include/c++/13/ext/atomicity.h:111
```

## Root Cause

Both `libunwind.so.8` and `libgcc_s.so.1` export overlapping `_Unwind_*` symbols:
- `_Unwind_RaiseException`
- `_Unwind_Resume`
- `_Unwind_GetGR` / `_Unwind_SetGR`
- `_Unwind_GetIP` / `_Unwind_SetIP`
- etc.

The dynamic linker resolves these to whichever library is loaded first. In our binary, brpc (`-lunwind -lunwind-x86_64` for bthread tracer) and glog (`target_link_libraries(glog PRIVATE unwind::unwind)`) both pull in `libunwind.so.8` early in the load order, so it wins symbol resolution over `libgcc_s.so.1`. Since libunwind's `_Unwind_RaiseException` doesn't know how to call `__gxx_personality_v0` for C++ LSDA-based exception dispatch, it fails and calls `abort()`.

### Where `libunwind` is linked

1. **brpc** (`third_party/brpc/CMakeLists.txt:343`):
   ```cmake
   set(DYNAMIC_LIB ${DYNAMIC_LIB} ${LIBUNWIND_LIB} ${LIBUNWIND_X86_64_LIB})
   ```

2. **glog** (`third_party/glog/CMakeLists.txt:627`):
   ```cmake
   target_link_libraries (glog PRIVATE unwind::unwind)
   ```

liulx20:
This pull request updates the `build_glog_as_third_party` function in the `cmake/BuildGlogAsThirdParty.cmake` file to address a symbol conflict issue between `libunwind` and `libgcc_s` when building glog as a third-party dependency. The most significant change is the explicit disabling of `libunwind` support in glog to prevent runtime errors related to exception handling.

Dependency build configuration:

* Added `set(WITH_UNWIND OFF ...)` to disable `libunwind` when building glog, preventing symbol conflicts with `libgcc_s` that could cause C++ exceptions to abort instead of being caught. This ensures reliable stack traces without introducing symbol resolution issues.


zhanglei1949 (approved):
LGTM

zhanglei1949:
From my point of view, it is ok to disable libunwind, since we don't really rely on the embedding stack trace printing from libunwind.

PR #394, opened 2026-05-22
fix: correct the unsorted boundary for csr

liulx20:
This pull request introduces improvements to the handling of the `unsorted_since_` flag in both `ImmutableCsr` and `MutableCsr` classes to ensure correct tracking of edge sorting states after various mutations. It also enables automatic compaction by default in the database configuration and adds a new Python test to verify correct neighbor sorting and compaction behavior. 

Fixes


zhanglei1949 (approved):
LGTM

Issue #398, opened 2026-05-22
Refactor: tighten DROP TABLE / checkpoint isolation contract


zhanglei1949:
## Motivation

Issue #293 reported two visible bugs from DROP TABLE: stale data after recreate, and a crash on the next CHECKPOINT. A first fix (PR #306) made the symptoms go away but introduced changes that, on review, weaken the storage layer's invariants and add footguns for future maintainers. This issue tracks the cleaner refactor.

## Architectural concerns with the first attempt (PR #306)

1. **`IDataContainer::Drop()` would silently corrupt checkpoints under non-default memory modes.** `container_utils::OpenContainer` only routes `path_` through `tmp_dir` for `kSyncToFile`. In `kInMemory` / `kHugePagePreferred`, the container's `path_` points directly at a **checkpoint** file. A generic Drop()-deletes-path_ contract is therefore safe today only by accident; a future override added under either of those modes would unlink checkpoint state.
2. **`bool load_from_checkpoint = true` leaks an implementation detail onto the public API**, with a default that silently re-introduces #293 if any new caller forgets to pass `false`.
3. **`DeleteEdgeType` did not call `EdgeTable::Drop()`** before erasing the map entry; cleanup happened only by checkpoint-time `copy_directory(overwrite=true)` GC. This was asymmetric with `DeleteVertexType` and obscured the lifecycle.

## Proposed design

- **`Drop()` releases in-memory owning state only; never touches the filesystem.** `IDataContainer` does not gain a `Drop()` method — the only filesystem exit at the container layer remains destruction-time `Close()`.
- **Three explicit physical-cleanup gates** for `tmp_dir`:
  - process startup: `PropertyGraph::Open` wipes `tmp_dir`
  - next CREATE for the same label/triple: `VertexTable::Initialize` / `EdgeTable::Initialize` sweeps stale tmp files keyed on the label NAME (not the numeric `label_t`, which schema reuses)
  - next CHECKPOINT: `PropertyGraph::Dump` rewrites `checkpoint_dir` from scratch
- **`Open()` is split into `Open()` (restore from checkpoint) and `Initialize()` (fresh table)** — no boolean parameter.
[...]

PR #306, opened 2026-04-29
fix: Introduce Drop() for MMapContainer and fix checkpoint related bugs

zhanglei1949:
Fix #293

PR #399, opened 2026-05-22
fix: DROP TABLE no longer leaks across checkpoint boundary (#293, #398)

### LINKED ISSUE #293: [BUG] DROP TABLE does not clear checkpointed data; stale nodes visible after CREATE TABLE IF NOT EXISTS
**Describe the bug**
After `CHECKPOINT`, `DROP TABLE` does not properly clear persisted data. When the same table schema is recreated with `CREATE TABLE IF NOT EXISTS`, the old nodes are still queryable as if they were never deleted.

Additionally, calling `CHECKPOINT` after `DROP TABLE` crashes with `Invalid vertex label id`, suggesting internal metadata inconsistency after the drop.

**Execution Logs**
Minimal repro script (pure neug 0.1.1 Python API, no dependencies):

```python
import neug, tempfile, os, shutil

d = tempfile.mkdtemp(prefix='neug_bug_')
db = neug.Database(os.path.join(d, 'graph.db'))
conn = db.connect()

# 1. Create table, insert data, checkpoint
conn.execute('CREATE NODE TABLE IF NOT EXISTS Person(id STRING PRIMARY KEY);')
conn.execute("CREATE (p:Person {id: 'alice'});")
conn.execute('CHECKPOINT')

rows = list(conn.execute('MATCH (p:Person) RETURN p.id;'))
print(f'1. After insert: {rows}')  # [['alice']]

# 2. Drop table
conn.execute('DROP TABLE IF EXISTS Person;')

# 3. Recreate table with same schema
conn.execute('CREATE NODE TABLE IF NOT EXISTS Person(id STRING PRIMARY KEY);')

# 4. Query — BUG: old data still visible
rows = list(conn.execute('MATCH (p:Person) RETURN p.id;'))
print(f'2. After DROP+reCREATE: {rows}')  # [['alice']] ← should be []

# 5. Insert new data
conn.execute("CREATE (p:Person {id: 'bob'});")
rows = list(conn.execute('MATCH (p:Person) RETURN p.id;'))
print(f'3. After new insert: {rows}')  # [['alice'], ['bob']] ← alice is stale

del conn; del db
shutil.rmtree(d, ignore_errors=True)
```

Output:
```
1. After insert: [['alice']]
2. After DROP+reCREATE: [['alice']]      ← BUG: should be []
3. After new insert: [['alice'], ['bob']] ← 'alice' is stale data
```

**Secondary issue — CHECKPOINT after DROP crashes:**

If you call `CHECKPOINT` after `DROP TABLE`, the following error occurs:
```
Error code: 1012, Error Message: ERR_INTERNAL_ERROR: Invalid vertex label id: 1
at /project/src/storages/graph/property_graph.cc:791 func: Open
```

**Expected behavior**
1. After `DROP TABLE` + `CREATE TABLE IF NOT EXISTS`, the table should be empty — old checkpointed nodes should not be visible.
2. `CHECKPOINT` after `DROP TABLE` should succeed without crashing.

**Environment**
- neug version: 0.1.1
- Python: 3.10
- OS: Linux (Ubuntu 22.04)

zhanglei1949:
## Summary

Fixes #293 and implements the refactor tracked in #398. Replaces the earlier PR #306 approach with a tighter Drop/checkpoint contract that keeps file-system mutation out of the container layer.

- **Drop()** at every layer releases in-memory state only; `checkpoint_dir` is touched exclusively by `Dump()`. This preserves the implicit crash-rollback semantics for DROP-before-CHECKPOINT.
- **`Open()` split into `Open()` (restore from checkpoint) and `Initialize()` (fresh table).** Replaces the leaky `bool load_from_checkpoint = true` parameter.
- **`Initialize()` sweeps `tmp_dir` for files keyed on the same label name/triple** before opening fresh containers. This is what unsticks the same-session DROP+CREATE stale-data path — schema reuses the numeric `label_t` for a same-name CREATE, but tmp filenames key on the string name.
- **`DeleteVertexType` and `DeleteEdgeType` call `Drop()` before erasing**, making the two paths symmetric.
- **`IDataContainer::Drop()` is intentionally NOT added.** In `kInMemory` / `kHugePagePreferred` modes a container's `path_` points at the checkpoint file, so a generic Drop-deletes-path contract would silently corrupt checkpoint state. See #398 for the full rationale.

### DIFF
--- a/src/storages/graph/edge_table.cc
+++ b/src/storages/graph/edge_table.cc
@@ -59,8 +59,7 @@ void cleanup_edge_label_tmp_files(const std::string& work_dir,
       edata_prefix(src_label_name, dst_label_name, edge_label_name),
       statistics_file_prefix(src_label_name, dst_label_name, edge_label_name),
   };
-  for (const auto& entry :
-       std::filesystem::directory_iterator(tmp_dir_path)) {
+  for (const auto& entry : std::filesystem::directory_iterator(tmp_dir_path)) {
     if (!entry.is_regular_file()) {
       continue;
     }
@@ -631,12 +630,14 @@ void EdgeTable::openImpl(const std::string& work_dir, MemoryLevel memory_level,
                                  ? ""
                                  : checkpoint_dir_path + "/" + oe_prefix_path);
   } else if (memory_level == MemoryLevel::kHugePagePreferred) {
-    in_csr_->open_with_hugepages(
-        checkpoint_dir_path.empty() ? ""
-                                    : checkpoint_dir_path + "/" + ie_prefix_path);
-    out_csr_->open_with_hugepages(
-        checkpoint_dir_path.empty() ? ""
-                                    : checkpoint_dir_path + "/" + oe_prefix_path);
+    in_csr_->open_with_hugepages(checkpoint_dir_path.empty()
+                                     ? ""
+                                     : checkpoint_dir_path + "/" +
+                                           ie_prefix_path);
+    out_csr_->open_with_hugepages(checkpoint_dir_path.empty()
+                                      ? ""
+                                      : checkpoint_dir_path + "/" +
+                                            oe_prefix_path);
   } else {
     THROW_INVALID_ARGUMENT_EXCEPTION(
         "unsupported memory level: " +
--- a/src/storages/graph/edge_table.cc
+++ b/src/storages/graph/edge_table.cc
@@ -38,6 +38,49 @@
 
 namespace neug {
 
+namespace {
+
+// Remove any tmp_dir files belonging to a previously dropped edge label of
+// the given (src, edge, dst) triple. tmp filenames key on label strings
+// (not numeric label_t), and Schema reuses the same numeric ids when a same-
+// triple edge is re-created, so without this sweep an in-process DROP+CREATE
+// would re-mmap the prior edge label's leftover data.
+void cleanup_edge_label_tmp_files(const std::string& work_dir,
+                                  const std::string& src_label_name,
+                                  const std::string& dst_label_name,
+                                  const std::string& edge_label_name) {
+  std::string tmp_dir_path = tmp_dir(work_dir);
+  if (!std::filesystem::exists(tmp_dir_path)) {
+    return;
+  }
+  const std::string prefixes[] = {
+      ie_prefix(src_label_name, dst_label_name, edge_label_name),
+      oe_prefix(src_label_name, dst_label_name, edge_label_name),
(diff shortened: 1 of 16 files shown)


src/storages/graph/edge_table.cc
```diff
@@ -592,16 +661,50 @@ void EdgeTable::Open(const std::string& work_dir, MemoryLevel memory_level) {
...
-    if (table_cap != capacity_.load()) {
-      THROW_INVALID_ARGUMENT_EXCEPTION(
-          "capacity in statistic file not match actual table capacity, maybe "
-          "the graph is not dumped properly");
+    if (!checkpoint_dir_path.empty()) {
+      load_statistic_file(work_dir, meta_->src_label_name,
+                          meta_->dst_label_name, meta_->edge_label_name,
+                          capacity_, table_idx_);
+      if (table_cap != capacity_.load()) {
+        THROW_INVALID_ARGUMENT_EXCEPTION(
```
liulx20: is INVALID_ARGUMENT_EXCEPTION appropriate here

src/storages/graph/edge_table.cc
```diff
@@ -592,16 +661,50 @@ void EdgeTable::Open(const std::string& work_dir, MemoryLevel memory_level) {
...
+//   - next CREATE for the same triple: Initialize sweeps stale tmp files
+//   - next CHECKPOINT: PropertyGraph::Dump rewrites checkpoint_dir from scratch
+//
+// checkpoint_dir is never modified outside of CHECKPOINT — this is what gives
+// DROP its implicit crash-rollback semantics (a crash between DROP and the
+// next CHECKPOINT re-reads the prior checkpoint, effectively undoing the
+// DROP). Do not make this function delete files; doing so requires writing
+// DROP to the WAL first or that rollback invariant will break.
+void EdgeTable::Drop() {
+  out_csr_.reset();
```
liulx20: is there a difference in usage between `Drop` and `Close`?

src/storages/graph/edge_table.cc
zhanglei1949: Replace with Internal error

src/storages/graph/edge_table.cc
zhanglei1949: It seems that for `LFIndexer`, `Table`, `VertexTable`, `EdgeTable`, the logic of `drop` and `close` are actually the same. Maybe we could unify two method into one.

liulx20 (dismissed):

liulx20 (approved):

Issue #86, opened 2026-03-19
Refactor UpdateTransaction


zhanglei1949:
Currently UpdateTransaction holds a global lock and blocks any other insertion/read since the UpdateTransaction starts.
However, it is possible to split UpdateTransaction into two stages.
First stage, UpdateTransaction maintains all changes copy on write, which will not effect the view for snapshots before this timestamp. So both insert transaction and read transaction could be parallelized with UpdateTransaction.
Second stage, UpdateTransaction begins to commit. At this time, a global lock is acquired, and no other transactions are allowed.

In addition, the transaction system currently lacks rollback capability when errors occur during the commit phase. This leads to partial commits that violate transaction atomicity.

Two specific scenarios are affected:

Insert transaction: Partial vertex insertions cannot be rolled back when a subsequent insert fails.
Update transaction: Schema drop operations cannot be undone once data is deleted, because bulk operations (drop/copy) lack WAL support.
See sub-issues #315 and #316 for detailed descriptions of each problem.

PR #401, opened 2026-05-25
refactor: introduce GraphView facade for snapshot-style PropertyGraph access

### LINKED ISSUE #400: Refactor: extract read-side view layer over PropertyGraph storage
## Background

`PropertyGraph` couples mutation API and read accessors. Every caller — read query, insert transaction, update transaction, batch loader — operates directly on the live `vertex_tables_` / `edge_tables_` / `Csr` state via the same handle.

## Problem

This direct-access model blocks the COW UpdateTransaction work in parent issue #86 in two concrete ways:

1. **No stable observation point.** Once UpdateTransaction starts forking storage components copy-on-write, callers that hold raw `PropertyGraph&` will see partial mid-fork state. There is no place to express "the graph as observed at timestamp T" — `PropertyGraph` is the only handle to the data, and it always reflects the latest mutation.

2. **Read vs. insert vs. update paths are not separable at the type level.** `StorageReadInterface` / `StorageTPInsertInterface` / `StorageAPUpdateInterface` all take a `PropertyGraph&` and reach into whichever accessor they need. A read path and a write path are indistinguishable at the boundary, and there is no way to hand two transactions a shared snapshot.

These together prevent the COW splitting described in #86 from being implemented cleanly.

zhanglei1949:
Fixes #400.

## Summary

Introduce `GraphView` — a thin handle wrapping `PropertyGraph*` that exposes the read-side surface (and a couple of mutators) needed for the COW `UpdateTransaction` work in #86. No callers are migrated in this PR; the view is added alongside `PropertyGraph` so follow-up changes to `ReadTransaction` / `StorageReadInterface` can route through a single stable observation point.

This PR also lands two small fixes uncovered while wiring up the new test suite.

## What's in this PR

**`GraphView` (new):**
- `include/neug/storages/graph/graph_view.h` + `src/storages/graph/graph_view.cc`
- Vertex-side reads: `GetVertexSet`, `get_lid`, `GetOid`, `IsValidLid`, `GetVertexPropertyColumn` (by name and by id).
- Edge-side reads: `GetGenericOutgoingView` / `GetGenericIncomingView` (returning `CsrView`), `GetEdgeDataAccessor` (by id and by name).
- Mutators kept on the view for now: `AddVertex`, `AddEdge`. These will move off `GraphView` once the read/write split is finalized.
- Holds only a `PropertyGraph*` — copyable/movable, no ownership.

**`csr_view.h` touch-ups:**
- `TypedCsrView<T, kMultipleMutable>` gets a default constructor and zero-initialized members so it can be returned by value safely (e.g. from the error path of `get_typed_view`).
- `CsrView::get_edges` now guards against a null `adjlists_[v]` in the multi-edge case instead of producing a `[nullptr, nullptr + degree*stride)` range.
- `get_typed_view` now logs the offending `CsrViewType` and explicitly instantiates the supported branch.

**`EdgeTable::get_edge_data_accessor(int col_id)`:**
- Out-of-range `col_id` now throws `InvalidArgumentException` instead of indexing past `meta_->properties`.
- The bundled-edge path rejects non-zero `col_id` with a clear error instead of silently returning a sentinel `(kEmpty, nullptr)` accessor.

**Tests / CI:**
- `tests/storage/test_graph_view.cc` (13 tests): construction, schema access, `get_lid` / `GetOid` / `IsValidLid`, vertex property column by name (incl. primary-key path) and by id (incl. the "id 0 is `name`, not `id`" asymmetry inherited from `VertexTable`), outgoing+incoming edge traversal, edge data accessor by id + by name, and error paths for missing properties / invalid triplets / out-of-range column ids.
- `tests/storage/CMakeLists.txt` registers `graph_view_test`; `.github/workflows/neug-test.yml` runs it.


src/storages/graph/edge_table_view.cc
```diff
@@ -0,0 +1,81 @@
...
+  return out_csr_ ? out_csr_->get_generic_view(ts) : CsrView();
+}
+
+CsrView EdgeTableView::get_incoming_view(timestamp_t ts) const {
+  return in_csr_ ? in_csr_->get_generic_view(ts) : CsrView();
+}
+
+EdgeDataAccessor EdgeTableView::get_edge_data_accessor(int col_id) const {
+  if (!edge_schema_) {
+    return EdgeDataAccessor();
```
liulx20: What does it mean when edge_meta is null? Is returning this value meaningful?

liulx20 (approved):

PR #415, opened 2026-05-27
fix: make httpfs/parquet load cleanly from OSS 

### LINKED ISSUE #414: [BUG] parquet extension fails to LOAD with undefined symbol: OCSP_basic_verify
**Describe the bug**

After enabling `ARROW_S3=ON` in the project-wide Arrow build (required by the `httpfs` extension to provide the OSS / S3 filesystem), the `parquet` extension fails to load at runtime with:

```
Failed to load extension library: libparquet.neug_extension.
Error: undefined symbol: OCSP_basic_verify
```

This is a regression that surfaces on any environment whose system OpenSSL does not coincidentally already export the OCSP / EC / BIO symbol set required by the AWS C++ SDK. Inside the manylinux build container and our test `ubuntu:22.04` container the failure is fully reproducible after `INSTALL parquet; LOAD parquet`.

Root cause:

* `ARROW_S3=ON` causes `arrow_bundled_dependencies.a` to drag in AWS C++ SDK object files.
* Those `.o` files reference 14+ OpenSSL symbols (`OCSP_basic_verify`, `EC_KEY_*`, `BIO_new`, `OPENSSL_init_ssl`, ...).
* The `parquet` extension does not perform any S3 / network work itself, so its previous `CMakeLists.txt` did not link OpenSSL.
* The transitive references arrive through Arrow's `INTERFACE_LINK_LIBRARIES`, but `libcrypto.a` appears earlier in the link line than `libaws-cpp-sdk-core.a`. GNU ld's single-pass demand-driven scan therefore elides the OCSP / `init_ssl` `.o` files (their symbols are not yet referenced when ld processes libcrypto), leaving them unresolved by the time it reaches the AWS archives.

**Execution Logs**

Reproduction (using `neug` 0.1.2 from TestPyPI inside an `ubuntu:22.04` container):

```bash
pip install --index-url https://test.pypi.org/simple/ \
            --extra-index-url https://pypi.org/simple/ neug==0.1.2

python3 -c "
import neug
db = neug.Database('/tmp/db', read_only=False)
conn = db.connect()
conn.execute('INSTALL parquet')   # downloads from OSS, sha256 OK
conn.execute('LOAD parquet')      # <-- fails here
"
```

**Expected behavior**

`LOAD parquet` succeeds and the extension's functions register, allowing `LOAD FROM "...parquet"` queries to run.

**Error Message**

```
RuntimeError: Failed to load extension library:
  /usr/local/lib/python3.10/dist-packages/extension/parquet/libparquet.neug_extension.
Error: /usr/local/lib/python3.10/dist-packages/extension/parquet/libparquet.neug_extension:
  undefined symbol: OCSP_basic_verify
```

`nm -D --undefined-only` on the broken `.so` additionally shows:

```
U OCSP_basic_verify
U EC_KEY_new_by_curve_name
U EC_KEY_set_public_key
U BIO_new
U BIO_new_mem_buf
U OPENSSL_init_ssl
U OPENSSL_init_crypto
... (14 total)
```

**Environment**

- Build container: `neug/neug-manylinux:x86_64`
- Test container: `ubuntu:22.04`
- `neug` wheel: 0.1.2 (TestPyPI)
- Affected extension: `parquet` (`linux_x86_64`)
- `httpfs` extension is not affected (already statically links OpenSSL via the cURL static-link work on `fix_httpfs`).

**Proposed fix**

Mirror the link-time treatment used by `httpfs`: in `extension/parquet/CMakeLists.txt`, `find_package(OpenSSL REQUIRED)` and link `OpenSSL::SSL` / `OpenSSL::Crypto` wrapped in `-Wl,--whole-archive ... -Wl,--no-whole-archive` so every `.o` (including `ocsp_vfy.o`) is force-included. Cost: +3 MB stripped binary (40 to 43 MB). The AWS code paths are never executed at runtime; only the linker needs the definitions present at load time.

### LINKED ISSUE #416: [BUG] httpfs extension is not portable: dynamic libcurl + missing TLS CA bundle break OSS HTTPS
**Describe the bug**

The `httpfs` extension shipped via `INSTALL httpfs` cannot reliably load OSS / HTTPS URLs on a freshly-installed `neug` wheel. Two concrete failure modes are observed depending on the host:

1. **Dynamic-link portability failure** — `dlopen` of `libhttpfs.neug_extension` fails on hosts whose system `libcurl.so.4` is missing, an incompatible ABI, or built against a different SSL backend than the one used inside the `neug-manylinux` build container. The extension was previously linked against the *system* `libcurl` of the build image, so the produced `.so` is not portable across distros.

2. **TLS certificate verification failure at runtime** — even on hosts where `libcurl.so.4` happens to be ABI-compatible, every HTTPS request through Arrow's S3 / HTTP filesystem fails with:

   ```
   curl error: SSL peer certificate or SSH remote key was not OK
   unable to get local issuer certificate
   ```

   Because the bundled cURL has no CA bundle baked in, and Arrow's `S3GlobalOptions` was not given a CA file/path, every `LOAD FROM "oss://..."` and every `iceberg://` request through the OSS filesystem hits the verification failure.

The end result: a clean `pip install neug==0.1.2` on `ubuntu:22.04` cannot read OSS Parquet, even though the OSS bucket is publicly accessible.

**Execution Logs**

Reproduction on a clean `ubuntu:22.04` container (without preinstalled `libcurl4-openssl-dev`):

```bash
pip install --index-url https://test.pypi.org/simple/ \
            --extra-index-url https://pypi.org/simple/ neug==0.1.2

python3 -c "
import neug
db = neug.Database('/tmp/db', read_only=False)
conn = db.connect()
conn.execute('INSTALL httpfs')
conn.execute('LOAD httpfs')                                   # may dlopen-fail here
conn.execute('LOAD FROM \"oss://graphscope/neug/vPerson.parquet\"')   # or fail here with CA error
"
```

**Expected behavior**

`LOAD httpfs` succeeds on any reasonable Linux host (no need for the user to preinstall `libcurl4-openssl-dev` or set `CURL_CA_BUNDLE`), and Arrow's S3/OSS HTTPS calls verify certificates against the host's system trust store automatically.

**Error Message**

Variant 1 — dlopen failure on a host without matching libcurl:

```
RuntimeError: Failed to load extension library:
  /usr/local/lib/python3.10/dist-packages/extension/httpfs/libhttpfs.neug_extension.
Error: libcurl.so.4: cannot open shared object file: No such file or directory
```

Variant 2 — CA verification failure on a host that does have libcurl:

```
arrow::Status NotImplemented: When initializing S3:
  curl error: SSL peer certificate or SSH remote key was not OK
  unable to get local issuer certificate
```

**Environment**

- Build container: `neug/neug-manylinux:x86_64`
- Test container: `ubuntu:22.04` (clean image, no `libcurl4-openssl-dev`)
- `neug` wheel: 0.1.2 (TestPyPI)
- Affected extension: `httpfs` (`linux_x86_64`)

**Proposed fix**

Make the `httpfs` extension self-contained at the binary level and CA-aware at runtime:

1. **Build cURL statically inside the manylinux image** — extend `scripts/install_deps.sh` to compile cURL from source as a static archive into `/opt/neug` with the OpenSSL backend, no system-CA hard-coding, and minimal feature surface.
2. **Statically link cURL into `httpfs`** — change `extension/httpfs/CMakeLists.txt` from `find_library(CURL ...)` to an explicit `find_library(... libcurl.a PATHS /opt/neug ...)`, and append `OpenSSL::SSL` / `OpenSSL::Crypto` / `ZLIB::ZLIB` / `pthread` / `dl` to the link line so the static cURL has all of its transitive deps resolved.
3. **Resolve the TLS CA bundle at runtime** — in `src/utils/s3_options.cc`, probe the standard system CA paths (`/etc/ssl/certs/ca-certificates.crt`, `/etc/pki/tls/certs/ca-bundle.crt`, `/etc/ssl/ca-bundle.pem`, ...) and pass the first hit into `arrow::fs::S3GlobalOptions{ca_file_path, ca_dir_path}` when the user did not supply one explicitly. This makes HTTPS verification work out-of-the-box on Debian/Ubuntu/RHEL/Alpine without any user setup.

Together these changes make the `httpfs` extension portable across distros and remove the need for users to set `CURL_CA_BUNDLE` / `SSL_CERT_FILE`.

BingqingLyu:
## Related Issues

- fix #416, #414 

## What does this PR do?

Fixes two distinct extension-loading regressions that prevent OSS / S3 Parquet
ingestion from working out-of-the-box on a freshly-installed `neug` wheel:

1. **`httpfs` extension was not portable (#416)** — it dynamically linked the
   build container's `libcurl.so.4`, so on any host without a matching system
   `libcurl` (or with a different SSL backend) `LOAD httpfs` failed at
   `dlopen` time, and even when it loaded, OSS HTTPS calls hit
   `unable to get local issuer certificate` because the bundled cURL had no
   CA bundle compiled in and Arrow's `S3GlobalOptions` was never given one.
2. **`parquet` extension failed to load with `undefined symbol: OCSP_basic_verify` (#414)**
   after enabling `ARROW_S3=ON` project-wide. AWS C++ SDK objects pulled in
   via `arrow_bundled_dependencies.a` reference 14+ OpenSSL symbols
   (`OCSP_basic_verify`, `EC_KEY_*`, `BIO_new`, `OPENSSL_init_ssl`, ...) but
   the parquet extension itself never linked OpenSSL, and GNU ld's
   single-pass scan elided the `.o` files containing those definitions.

The combined effect: end-to-end `INSTALL httpfs; INSTALL parquet; LOAD httpfs;
LOAD parquet; LOAD FROM "oss://..."` now works on a clean `ubuntu:22.04`
container with only the `neug` wheel installed.

## What changes in this PR?

**Commit 1 — `fix(httpfs): statically link cURL and resolve TLS CA bundle at runtime` (`3cc49d0`, fixes #416)**

- `scripts/install_deps.sh`: build cURL from source as a static library
  inside the manylinux container with OpenSSL backend and no system-CA
  hard-coding, installed under `/opt/neug`.
- `extension/httpfs/CMakeLists.txt`: switch from `find_library(CURL ...)` to
  an explicit static-archive search (`libcurl.a` under `/opt/neug`), and
  link `OpenSSL::SSL` / `OpenSSL::Crypto` / `ZLIB::ZLIB` / `pthread` /
  `dl` alongside the static cURL.
- `cmake/BuildArrowAsThirdParty.cmake`: ensure `ARROW_S3=ON` is enabled
  early enough that the AWS / Azure SDK pieces are present when extensions
  are configured.
- `src/utils/s3_options.cc`: at runtime auto-detect the system CA bundle
  (`/etc/ssl/certs/ca-certificates.crt` on Debian/Ubuntu, `/etc/pki/...`
  on RHEL, etc.) and pass it to Arrow's S3 filesystem global options, so
  the bundled cURL can verify HTTPS certificates without any environment
  setup on the user's side.

**Commit 2 — `fix(parquet ext): force-link OpenSSL to resolve OCSP symbols pulled by Arrow's bundled AWS SDK` (`29b7503`, fixes #414)**

- `extension/parquet/CMakeLists.txt`: `find_package(OpenSSL REQUIRED)` and
  link `OpenSSL::SSL` / `OpenSSL::Crypto` wrapped in
  `-Wl,--whole-archive ... -Wl,--no-whole-archive`. This force-includes
  every `.o` (notably `ocsp_vfy.o` and the OPENSSL init objects) into the
  final extension `.so`, so the AWS-SDK references that arrive via
  Arrow's `INTERFACE_LINK_LIBRARIES` are all resolved at link time.
- Cost: ~3 MB stripped binary growth on `parquet` (40 → 43 MB). The AWS
  code paths are never executed inside `parquet`; only the linker needs
  the definitions present.

## Verification

End-to-end test on a clean `ubuntu:22.04` container with `neug==0.1.2`
installed from TestPyPI and the rebuilt extensions uploaded to OSS at
`v0.1.2/linux_x86_64/{httpfs,parquet}/`:

```
INSTALL httpfs        ✓  download from OSS + sha256 verify OK
INSTALL parquet       ✓  download from OSS + sha256 verify OK
LOAD httpfs           ✓  schemes registered: s3, oss, http, https
LOAD parquet          ✓  no undefined symbols, functions registered
LOAD FROM "oss://graphscope/neug/vPerson.parquet"  →  8 rows × 16 cols
LOAD FROM "oss://graphscope/neug/eMeets.parquet"   →  7 rows × 5  cols
```

`nm -D --undefined-only libparquet.neug_extension` after the fix:
no remaining `OCSP_*` / `OPENSSL_init_*` / `EC_KEY_*` / `BIO_*`
unresolved symbols.

## Risks / Backwards compatibility

- `parquet` extension grows by ~3 MB stripped — acceptable for a graph
  database extension binary.
- `httpfs` no longer relies on the host `libcurl`, removing a class of
  per-distro portability bugs.
- No public API changes, no schema or storage format changes.
- No changes to the main `neug` wheel — only the two extension `.so`s
  on the OSS distribution path are affected, so existing users who do
  not call `INSTALL httpfs` / `INSTALL parquet` are unaffected.


lnfjpt (approved):
LGTM

PR #370, opened 2026-05-18
refactor: switch UpdateTransaction to COW snapshot isolation

zhanglei1949:
## Resolved issues

- Fixes #330: COW for UpdateTransaction.
- Fixes #548: checkpoint-only UpdateTransaction commits are counted.
- Fixes #549: CompactTransaction timestamp release race.

## Summary

- Adds workspace/checkpoint-backed graph snapshot storage, based on #148.
- Refactors UpdateTransaction to publish COW graph snapshots with lazy write materialization.

PR #422, opened 2026-05-28
refactor: validate label / edge-triplet IDs at PropertyGraph API boundary

### LINKED ISSUE #421: PropertyGraph: harden label / edge-triplet validation at API boundary
## Problem

Several `PropertyGraph` public APIs trust caller-supplied `label_t` / edge-triplet IDs without validation:

- **Mutating ops** (`BatchAddVertices`, `BatchDeleteVertices`, `DeleteVertex` ×2, `DeleteEdge`, `BatchDeleteEdges` ×2, `AddVertex`, `UpdateVertexProperty`, `UpdateEdgeProperty`) either `assert(...)` (compiled out in Release) or rely on the called sub-component to catch the bad ID. In Release builds, an invalid `label_t` becomes out-of-bounds vector access on `vertex_tables_[label]` → undefined behavior.
- **Non-Status accessors** (`get_vertex_table` ×2, `GetVertexPropertyColumn` ×2, `GetVertexSet`, `LidNum`, `VertexNum`, `IsValidLid`, `get_lid`, `GetOid`) index `vertex_tables_[label]` with no check — same UB risk.
- `BatchAddEdges` already does the right thing, but inline (three separate `schema_.ensure_*` calls) — could reuse a single helper.
- `GetVertexPropertyColumn(label, col_id)` does not bounds-check `col_id` either; a stray id triggers a vector OOB inside `vertex_tables_[label].GetPropertyColumn(col_id)`.
- The existing private check helpers are inconsistent: `vertex_label_check(string)` does `LOG(ERROR)` and uses awkward `"Vertex label[X] does not exists."` wording, while a `label_t` overload doesn't exist; `edge_triplet_check(string,string,string)` returns a single generic `"Edge [E] from [S] to [D] does not exist"` even when the actual problem is one of the vertex labels.

## Proposed fix

1. Add `label_t` overloads of the existing private helpers and make all four overloads `const`:
   ```cpp
   Status vertex_label_check(const std::string&) const;
   Status vertex_label_check(label_t) const;
   Status edge_triplet_check(const std::string&, const std::string&,
                             const std::string&) const;
   Status edge_triplet_check(label_t, label_t, label_t) const;
   ```
   Both `edge_triplet_check` overloads delegate per-vertex validation to `vertex_label_check` so they share logic and produce equally granular errors. Drop the `LOG(ERROR)` calls (caller decides).
2. Mutating APIs: replace `assert(...)` / missing checks with `RETURN_IF_NOT_OK(...check(...))` — invalid IDs return `ERR_INVALID_ARGUMENT` instead of UB/crash.
3. Non-Status accessors: call `schema_.ensure_vertex_label_valid(label)` (throws `InvalidArgumentException`) — only viable signal channel for `vid_t` / `bool` / `Property` / `T&` returns.
4. `GetVertexPropertyColumn(label, col_id)`: also bounds-check `col_id` against `schema_.get_vertex_properties(label).size()`.
5. Unify error wording across the string / `label_t` overloads (`"Vertex label '<name>' is not valid"` / `"Vertex label id <N> is not valid"`; same `"Edge triplet <...> is not valid"` shape).

## Why now

While reviewing the `zl/update-txn-cow` branch (PR #370 split work), these validation changes were sitting alongside workspace/checkpoint and COW changes in `property_graph.{h,cc}` with no semantic dependency on either. Splitting them out gives those downstream PRs a cleaner diff and lets this hardening land independently.

## Scope

- Only `include/neug/storages/graph/property_graph.h` and `src/storages/graph/property_graph.cc`.
- No behavior change for valid inputs.
- For invalid inputs: previously UB → now `Status` error (mutating ops) or `InvalidArgumentException` (accessors).
- Error wording for the existing string-based `vertex_label_check` / `edge_triplet_check` changes shape; verified no tests assert on the old strings.

zhanglei1949:
## What do these changes do?

Harden `PropertyGraph`'s label-id and edge-triplet validation so that bad caller-supplied IDs no longer become undefined behavior in Release builds, and unify the existing private check helpers so the string and `label_t` overloads behave consistently.

- Add `label_t` overloads of the private helpers and make all four overloads `const`. Both `edge_triplet_check` overloads now delegate per-vertex validation to `vertex_label_check`, so they share logic and produce equally granular errors:
  ```cpp
  Status vertex_label_check(const std::string&) const;
  Status vertex_label_check(label_t) const;
  Status edge_triplet_check(const std::string&, const std::string&,
                            const std::string&) const;
  Status edge_triplet_check(label_t, label_t, label_t) const;
  ```
- **Mutating APIs** (`BatchAddVertices`, `BatchDeleteVertices`, `DeleteVertex` ×2, `DeleteEdge`, `BatchDeleteEdges` ×2, `AddVertex`, `UpdateVertexProperty`, `UpdateEdgeProperty`) now `RETURN_IF_NOT_OK(vertex_label_check(...) / edge_triplet_check(...))` instead of `assert(...)` or relying on the callee to catch the bad ID.
- **Non-Status accessors** (`get_vertex_table` ×2, `GetVertexPropertyColumn` ×2, `GetVertexSet`, `LidNum`, `VertexNum`, `IsValidLid`, `get_lid`, `GetOid`) now call `schema_.ensure_vertex_label_valid(...)` (throws `InvalidArgumentException`) — the only viable signal channel for non-`Status` returns.
- `GetVertexPropertyColumn(label, col_id)` additionally bounds-checks `col_id` against the label's property count.
- Unify error wording across overloads: `"Vertex label '<name>' is not valid"` / `"Vertex label id <N> is not valid"`; same `"Edge triplet <...> is not valid"` shape. Drop the `LOG(ERROR)` calls from the existing string-based helpers (caller decides).

No behavior change for valid inputs. Verified no tests assert on the old error wording.

## Background

These changes were originally bundled inside the `zl/update-txn-cow` branch (PR #370 split work) but have no semantic dependency on the workspace/checkpoint or COW refactor. Splitting them out keeps those downstream PRs focused and lets this hardening land independently against `main`.

## Related issue number

Fixes #421

### DIFF
--- a/include/neug/storages/graph/property_graph.h
+++ b/include/neug/storages/graph/property_graph.h
@@ -329,10 +329,12 @@ class PropertyGraph {
       const std::vector<std::pair<vid_t, int32_t>>& ie_edges);
 
   inline VertexTable& get_vertex_table(label_t vertex_label) {
+    schema_.ensure_vertex_label_valid(vertex_label);
     return vertex_tables_[vertex_label];
   }
 
   inline const VertexTable& get_vertex_table(label_t vertex_label) const {
+    schema_.ensure_vertex_label_valid(vertex_label);
     return vertex_tables_[vertex_label];
   }
 
@@ -556,16 +558,25 @@ class PropertyGraph {
   void loadSchema(const std::string& filename);
   inline std::shared_ptr<RefColumnBase> GetVertexPropertyColumn(
       uint8_t label, int32_t col_id) const {
+    schema_.ensure_vertex_label_valid(label);
+    auto props = schema_.get_vertex_properties(label);
+    if (col_id < 0 || static_cast<size_t>(col_id) >= props.size()) {
+      THROW_INVALID_ARGUMENT_EXCEPTION(
+          "Vertex property column id out of range: " + std::to_string(col_id) +
+          " (label has " + std::to_string(props.size()) + " properties)");
+    }
     return vertex_tables_[label].GetPropertyColumn(col_id);
   }
 
   inline std::shared_ptr<RefColumnBase> GetVertexPropertyColumn(
       uint8_t label, const std::string& prop) const {
+    schema_.ensure_vertex_label_valid(label);
     return vertex_tables_[label].GetPropertyColumn(prop);
   }
 
   inline VertexSet GetVertexSet(label_t label,
                                 timestamp_t ts = MAX_TIMESTAMP) const {
+    schema_.ensure_vertex_label_valid(label);
     return vertex_tables_[label].GetVertexSet(ts);
   }
 
@@ -587,11 +598,14 @@ class PropertyGraph {
                                       const std::vector<std::string>& props,
                                       std::vector<std::string>& valid_props);
 
+  Status vertex_label_check(const std::string& vertex_type_name) const;
+  Status vertex_label_check(label_t label) const;
+
   Status edge_triplet_check(const std::string& src_type_name,
                             const std::string& dst_type_name,
-                            const std::string& edge_type_name);
-
-  Status vertex_label_check(const std::string& vertex_type_name);
+                            const std::string& edge_type_name) const;
+  Status edge_triplet_check(label_t src_label, label_t dst_label,
+                            label_t edge_label) const;
 
   void compact_schema();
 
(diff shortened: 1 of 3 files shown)


include/neug/storages/graph/property_graph.h
```diff
@@ -556,16 +558,25 @@ class PropertyGraph {
   void loadSchema(const std::string& filename);
   inline std::shared_ptr<RefColumnBase> GetVertexPropertyColumn(
       uint8_t label, int32_t col_id) const {
+    schema_.ensure_vertex_label_valid(label);
```
zhanglei1949, line 561: returning null or throw exception?

liulx20 (approved):

PR #436, opened 2026-06-01
fix: centralize neug path setup in conftest to prefer local package

### LINKED ISSUE #429: [BUG] Old data still exists after drop and re-create
This test failed on Linux (not dev container):
```
def test_drop_and_recreate_node_table_no_stale_data(tmp_path):
    """After DROP + re-CREATE of a node table, old rows must not reappear."""
    db_dir = tmp_path / "drop_recreate_stale"
    db = Database(db_path=str(db_dir), mode="w")
    conn = db.connect()

    conn.execute("CREATE NODE TABLE IF NOT EXISTS Person(id STRING PRIMARY KEY);")
    conn.execute("CREATE (p:Person {id: 'alice'});")
    conn.execute("CHECKPOINT")
    assert list(conn.execute("MATCH (p:Person) RETURN p.id;")) == [["alice"]]

    # Drop and re-create with the same schema
    conn.execute("DROP TABLE IF EXISTS Person;")
    conn.execute("CREATE NODE TABLE IF NOT EXISTS Person(id STRING PRIMARY KEY);")

    # Old data must be gone
    assert list(conn.execute("MATCH (p:Person) RETURN p.id;")) == []  //  <---- still ['alice']

    # Only newly inserted data should be visible
    conn.execute("CREATE (p:Person {id: 'bob'});")
    assert list(conn.execute("MATCH (p:Person) RETURN p.id;")) == [["bob"]]

    conn.close()
    db.close()
```

Spockkk0225:
## Related Issues
fix #429

## What does this PR do?
统一在 conftest.py 中配置 neug 包路径，确保优先使用本地 neug/ 而非 site-packages 中安装的版本。

## What changes in this PR?
- 在 tests/conftest.py 中添加 `sys.path.insert(0, ...)` 将 tools/python_bind/ 置于 sys.path 最前面
- 解决了 conftest.py 先于测试文件加载，导致 append 无效的问题
- 删除所有测试文件中冗余的 `sys.path.append` 调用（共 25 处）
- 解决了 append 优先级低于 site-packages，本地包被 pip 版本遮蔽的问题

## Why the old code was ineffective
- `sys.path.append` in test files runs AFTER conftest.py is loaded, so it cannot help conftest.py's imports
- `append` places the local path AFTER site-packages, so pip-installed neug takes priority
- `conftest.py` is pytest's auto-loaded file that runs before any test module in the same directory

### DIFF
--- a/tools/python_bind/tests/conftest.py
+++ b/tools/python_bind/tests/conftest.py
@@ -16,6 +16,11 @@
 # limitations under the License.
 #
 
+import os
+import sys
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
 import pytest
 
 from neug.session import Session
--- a/tools/python_bind/tests/test_alter_property.py
+++ b/tools/python_bind/tests/test_alter_property.py
@@ -22,8 +22,6 @@ import sys
 import time
 import unittest
 
-sys.path.append(os.path.join(os.path.dirname(__file__), "../"))
-
 from neug.database import Database
 
 logger = logging.getLogger(__name__)
--- a/tools/python_bind/tests/test_batch_loading.py
+++ b/tools/python_bind/tests/test_batch_loading.py
@@ -23,8 +23,6 @@ import sys
 import time
 import unittest
 
-sys.path.append(os.path.join(os.path.dirname(__file__), "../"))
-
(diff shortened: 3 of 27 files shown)


lnfjpt (approved):
LGTM

PR #458, opened 2026-06-03
feat: support CSV array (LIST) type in LOAD FROM via CAST

### LINKED ISSUE #441: CSV: Array (LIST) type not supported in load_data
## Parent Issue

Part of #221

## Problem

`load_data` with CSV format does not support Array (LIST) type columns. The following gaps exist:

1. **Arrow CSV Reader does not support LIST types**: Arrow's CSV Reader only supports scalar target types (numeric, string, boolean, timestamp, etc.). Even if `column_types` is set to `arrow::list(...)`, the reader cannot parse CSV cells containing array-like text (e.g., `[1,2,3]` or `1;2;3`) into a `ListArray`.
2. **Storage layer**: `set_properties_column()` lacks a `kList` case, so LIST data cannot be written into graph storage.

shirly121:
## Summary

- Implement CSV array type support: Arrow CSV Reader reads list-typed columns as strings, then `ArrowTypeCaster` converts them to Arrow ListArrays via post-processing (string→list parsing for `[1, 2, 3]` format)
- Add list/fixed-size-list handling in `ArrowArrayContextColumn::get_elem()` so the execution layer can extract values from Arrow ListArrays
- Fix `Database.close()` double-close guard to prevent `ValueError: I/O operation on closed file` during interpreter shutdown

## Test plan

- [x] 11 Python end-to-end tests (`test_load_array.py`): FLOAT[], DOUBLE[], INT32[], INT64[], STRING[], nested INT32[][], nested STRING[][], DATE[], TIMESTAMP[], INTERVAL[], multiple array columns
- [x] C++ unit tests (`test_reader.cc`, `test_sniffer.cc`): ArrowTypeCaster string→list conversion, CSV-safe schema projection, batch/full read with list columns, sniffer list column inference
- [x] All 292 existing C++ tests pass
- [x] `test_sniffer.py::test_csv_cast_list` now passes (xfail removed)

Closes #441

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/neug/utils/reader/arrow_type_cast.h
+++ b/include/neug/utils/reader/arrow_type_cast.h
@@ -1,0 +1,121 @@
+/** Copyright 2020 Alibaba Group Holding Limited.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * 	http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <arrow/api.h>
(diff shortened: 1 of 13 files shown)


longbinlai:
Need doc for such feature

shirly121:
> Need doc for such feature

这个要不等到 storage 支持 list PR 合进去了，再统一加一下？

PR #462, opened 2026-06-03
refactor: extract VarLenColumn base from TypedColumn<string_view>

### LINKED ISSUE #463: refactor: extract VarLenColumn base class from TypedColumn<string_view>
## Background

`TypedColumn<std::string_view>` in `include/neug/utils/property/column.h` carries all the dual-buffer storage machinery (items buffer + data buffer + atomic write frontier `pos_`) inline in the specialization. Concretely, ~270 lines mix three concerns into one class:

1. **Generic variable-length storage** — `items_buffer_` / `data_buffer_` lifecycle, `open` / `close` / `dump` (with MD5 dedup), `resize`, `ingest`, `init_pos`, `available_space`, the grow-or-throw policy on writes, and the resize-and-replicate-default-blob path.
2. **`var_len_item` (currently `string_item`) accounting** — offset/length entries, average-size computation, raw blob access.
3. **String-specific concerns** — `width_`, `type_` (`kVarchar`), UTF-8 truncation.

Any future variable-length column type (lists, JSON blobs, binary, etc.) would have to either duplicate (1) and (2) or hack around them via a string-only specialization. The grow-or-throw policy in `set_any` is also re-implemented across paths inside the class.

## Proposal

Extract a non-template `VarLenColumn : public ColumnBase` that owns concerns (1) and (2), and reduce `TypedColumn<std::string_view>` to a thin specialization that overrides only the string-specific bits.

- Rename `struct string_item` → `struct var_len_item` (offset + length entry, unchanged layout).
- Introduce `class VarLenColumn : public ColumnBase` that owns `items_buffer_`, `data_buffer_`, `size_`, `pos_`, and all the dual-buffer lifecycle methods (`open` / `close` / `dump` / `resize` / `ingest`).
- Provide protected helpers:
  - `ensure_capacity_or_throw(idx, blob_size, insert_safe)` — single grow-or-throw policy for writes.
  - `resize_with_replicated_default(new_size, default_blob)` — resize + write default blob once + replicate item entry for new rows.
  - `set_value_internal`, `get_raw_view`, `get_item`, `set_item`, `avg_size`.
- Add a protected move ctor (implicit move is deleted because of `std::atomic<size_t> pos_`) so subclass move ctors can delegate.
- `TypedColumn<std::string_view>` inherits from `VarLenColumn` and shrinks to ~70 lines holding only `width_` / `type_`, `truncate_if_needed`, and the public `set_value` / `set_any` / `get_view` / `get_prop` / `set_prop` / resize-with-default overrides.
- Promote `ColumnBase::set_prop` to pure virtual (currently defaults to `LOG(FATAL)`); all three existing subclasses already override it, so no caller behavior changes.

No behavior change for the public `StringColumn` API.

## Motivation

This is the storage-layer foundation needed by upcoming list-column support — the list column shares the entire dual-buffer + atomic-frontier machinery with strings and differs only in (a) the per-row size hint and (b) the blob-payload semantics. Extracting the base now means the list PR can focus on the new type rather than on `column.h` restructuring.

It's also independently meaningful as a cleanup of `TypedColumn<std::string_view>` even without the list work: the grow-or-throw policy and the resize-default path are clearer as named helpers than inlined into every caller.

## Scope

- `include/neug/utils/property/column.h` only. ~+238 / −167 lines, single file, no new headers, no new dependencies, no caller updates needed.

## Out of scope

- New variable-length column types (list, etc.) — follow-up PRs.
- Header hygiene cleanup of `column.h` (`#include <mutex>`, `<shared_mutex>`, `<unordered_map>`, `<unordered_set>`, `<vector>`, `file_names.h`, `likely.h`) — these are kept as-is to avoid breaking transitive includes from downstream headers; a focused cleanup PR can drop them after auditing dependents.

zhanglei1949:
## Summary

Lift the dual-buffer storage logic (items + data + atomic write frontier) out of `TypedColumn<std::string_view>` into a non-template base class `VarLenColumn`. `TypedColumn<std::string_view>` now inherits from `VarLenColumn` and keeps only varchar-specific concerns (width, UTF-8 truncation).

This is the storage-layer foundation needed by upcoming list-column support, but it's independently meaningful as a cleanup of `TypedColumn<std::string_view>`. Split out so the list PR can focus on the new type rather than on `column.h` restructuring.

## What changes in `column.h`

- Rename struct `string_item` → `var_len_item`.
- Introduce `class VarLenColumn : public ColumnBase` with all shared storage members (`items_buffer_`, `data_buffer_`, `size_`, `pos_`) plus `open` / `close` / `dump` / `resize` / `ingest`, `init_pos`, `available_space`, and the protected helpers `get_item` / `set_item` / `get_raw_view` / `avg_size` / `set_value_internal`.
- Add protected helpers `ensure_capacity_or_throw()` and `resize_with_replicated_default()` to centralize the grow-or-throw policy and the resize-and-replicate-default-blob path so subclasses don't reimplement them.
- Add a protected move ctor on `VarLenColumn` (implicit move is deleted because of `std::atomic`) for subclass move ctors to delegate to.
- `TypedColumn<std::string_view>` shrinks to ~70 lines: `width_` / `type_`, `truncate_if_needed`, and the public `set_value` / `set_any` / `get_view` / `get_prop` / `set_prop` / resize-with-default overrides.
- `ColumnBase::set_prop` becomes pure virtual instead of a `LOG(FATAL)` default; all three existing subclasses (`TypedColumn<T>`, `TypedColumn<EmptyType>`, `TypedColumn<std::string_view>`) already override it, so no caller behavior changes.
- Minor C++ hygiene: `= default` on trivial dtors, defaulted `ColumnBase` dtor.

## Behavior changes (intentional, all in the failure / over-allocation path)

The public success path for `StringColumn` is unchanged. A few corner-cases differ:

- **Capacity-exceeded exception type and message.** When a write to a `VarLenColumn` exceeds `data_buffer_` size, `ensure_capacity_or_throw` now throws `THROW_STORAGE_EXCEPTION` with `"Not enough space in buffer for new value, and insert_safe is false. Current buffer size: ..., current position: ..., new value size: ..."`. The old `TypedColumn<std::string_view>::set_value` threw `THROW_RUNTIME_ERROR("Index out of range or not enough space in buffer")` for the combined OOR + OOM case. Tests asserting on the exact wording or exception type would need updating.
- **`set_any` truncation order.** Old `set_any` did the capacity check on the *untruncated* string size, then truncated inside `set_value`. New `set_any` truncates first via `truncate_if_needed`, then does the capacity check on the truncated size. Net effect: less over-allocation when input exceeds `width_`. Identical results for inputs within `width_`.
- **`resize(size, default_value)` data-buffer growth heuristic.** Old path pre-allocated `pos_ + new_items * width_` bytes (worst case, as if every new row had a distinct max-length default). New path goes through `VarLenColumn::resize` (`max(size * default_avg_size(), pos_)`) and then grows once more inside `resize_with_replicated_default` if the single replicated default blob doesn't fit. Net effect: slightly less aggressive pre-allocation when expanding with a default; same on-disk output.
- **`ingest` adds an explicit capacity guard.** `VarLenColumn::ingest` now calls `ensure_capacity_or_throw(idx, sv.size(), /*insert_safe=*/false)` before writing. This preserves the old safety guarantee (`set_value`'s "not enough space" throw) at the new abstraction layer; without this guard, an undersized `data_buffer_` would have silently corrupted memory in release builds. Exception type/message is the new `STORAGE_EXCEPTION` form noted above.

Fix #463

### DIFF
--- a/include/neug/utils/property/column.h
+++ b/include/neug/utils/property/column.h
@@ -415,8 +415,8 @@ class VarLenColumn : public ColumnBase {
   // Ensure data_buffer_ has room for `blob_size` more bytes starting at pos_.
   // If not, grow when `insert_safe` is true (caller holds external lock),
   // otherwise throw. The `idx` is used only as a hint for the grow heuristic.
-  void ensure_capacity_or_throw(size_t idx, size_t blob_size,
-                                bool insert_safe) {
+  inline void ensure_capacity_or_throw(size_t idx, size_t blob_size,
+                                       bool insert_safe) {
     size_t cur_pos = pos_.load();
     if (cur_pos + blob_size <= data_buffer_->GetDataSize()) {
       return;
@@ -461,7 +461,7 @@ class VarLenColumn : public ColumnBase {
 
   // Append blob to data buffer at current pos_ and record item at idx.
   // Caller must ensure sufficient space in data_buffer_.
-  void set_value_internal(size_t idx, std::string_view blob) {
+  inline void set_value_internal(size_t idx, std::string_view blob) {
     assert(idx < size_);
     size_t offset = pos_.fetch_add(blob.size());
     if (!blob.empty()) {
@@ -473,7 +473,7 @@ class VarLenColumn : public ColumnBase {
   }
 
   // Return raw view of the blob at idx.
-  std::string_view get_raw_view(size_t idx) const {
+  inline std::string_view get_raw_view(size_t idx) const {
     assert(idx < size_);
     const auto item = get_item(idx);
     assert(item.offset + item.length <= data_buffer_->GetDataSize());
--- a/include/neug/utils/property/column.h
+++ b/include/neug/utils/property/column.h
@@ -273,8 +273,7 @@ class VarLenColumn : public ColumnBase {
   }
 
   void open_in_memory(const std::string& name) override {
-    items_buffer_ =
-        OpenContainer(name + ".items", "", MemoryLevel::kInMemory);
+    items_buffer_ = OpenContainer(name + ".items", "", MemoryLevel::kInMemory);
     data_buffer_ = OpenContainer(name + ".data", "", MemoryLevel::kInMemory);
     size_ = items_buffer_->GetDataSize() / sizeof(var_len_item);
     init_pos(name + ".pos");
--- a/include/neug/utils/property/column.h
+++ b/include/neug/utils/property/column.h
@@ -50,7 +50,7 @@ std::string_view truncate_utf8(std::string_view str, size_t length);
 
 class ColumnBase {
  public:
-  virtual ~ColumnBase() {}
+  virtual ~ColumnBase() = default;
 
   virtual void open(const std::string& name, const std::string& snapshot_dir,
                     const std::string& work_dir) = 0;
@@ -79,9 +79,7 @@ class ColumnBase {
 
   virtual Property get_prop(size_t index) const = 0;
 
-  virtual void set_prop(size_t index, const Property& prop) {
-    LOG(FATAL) << "Not implemented";
-  }
+  virtual void set_prop(size_t index, const Property& prop) = 0;
 
   virtual void ingest(uint32_t index, OutArchive& arc) = 0;
 };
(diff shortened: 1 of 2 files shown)


zhanglei1949:
Closed since not planned.

PR #468, opened 2026-06-03
build: share root cmake build across bindings; preserve env-var-driven options

### LINKED ISSUE #467: build: bindings reimplement core build instead of sharing root cmake tree
## Problem

`tools/python_bind/setup.py` configures and builds its own copy of the entire neug tree at `tools/python_bind/build/`, separate from the root `<repo>/build/` produced by a plain `cmake -S . -B build`. Consequences:

1. **Double-compile**: developers who run `cmake --build build` to iterate on C++ then run `make build` from `tools/python_bind/` end up re-compiling core + all third-party deps (arrow, brpc, protobuf, abseil, ...) inside `tools/python_bind/build/neug_py_bind/`.
2. **Not reusable**: any future binding (Node, Rust, Wasm) would have to replicate the same self-contained build setup, or invent its own.
3. **Loader is path-coupled**: `neug/__init__.py:get_build_lib_dir()` only recognizes setuptools' `lib.<platform>-...-cpython-XY/` directory naming, so `cmake --build build --target neug_py_bind` from repo root produces an artifact that the Python package can't find.
4. **Make targets are inconsistent**: `make build` was removed at some point (breaks CI which still calls it); `dev` doesn't bootstrap when cache is missing; `cmake --build -j` runs with unbounded parallelism.
5. **Python detection is implicit**: cmake auto-detects Python which can pick a different interpreter than the one running pytest, producing a wrong-ABI `neug_py_bind.cpython-XYZ-darwin.so` that the running Python can't import.

## Proposal

- Make `<repo>/build/` the single build tree shared by core and all bindings.
- Have each binding's CMakeLists.txt place its `.so` at a predictable `<repo>/build/tools/<binding>/` and link dynamically against `<repo>/build/src/libneug.{dylib,so}`.
- Update `setup.py` to consume the root build (with a fallback for fresh cibuildwheel containers) while still honoring all env vars CI sets.
- Update the Python loader to discover the .so at the root build path.
- Layer Makefile targets clearly: `build` = env-var-driven (CI), `dev`/`dev-full` = cmake-direct (fast).
- Bundle `libneug` into wheels via `@loader_path` / `$ORIGIN` RPATH so wheels are self-contained.

zhanglei1949:
## Summary

- Make `<repo>/build/` the single source of truth for all bindings; the Python binding consumes those artifacts instead of running its own full cmake configure.
- Preserve env-var-driven build options (BUILD_HTTP_SERVER, BUILD_TEST, ...) via `make build` → `setup.py` path; CI is unaffected.
- Bundle `libneug.{dylib,so}` into the wheel via `@loader_path` / `$ORIGIN` RPATH.
- New root targets `cpp-build` / `cpp-test` for C++-only workflows.
- JOBS auto-detected (no more `-j` unbounded); PYTHON locked to current `python3` to avoid wrong-ABI .so.
- Update GitHub workflows to reference the new `<repo>/build/` paths (commit `a6ad126`).

Fixes #467

## CI compatibility note

`make build` in `tools/python_bind/` is preserved as the env-var-honoring entry that CI uses. `setup.py` always re-runs cmake configure so env changes (BUILD_EXECUTABLES, BUILD_HTTP_SERVER, BUILD_TEST, ...) take effect on every invocation.

The 4 affected workflow files (`neug-test.yml`, `neug-extension-test.yml`, `benchmark.yml`, `ldbc-nightly-benchmark.yml`) had references to the old `tools/python_bind/build/neug_py_bind/...` cmake build tree. All updated in commit `a6ad126`:

- `./build/neug_py_bind/examples/cpp/simple_example` → `${GITHUB_WORKSPACE}/build/examples/cpp/simple_example`
- `./build/neug_py_bind/tools/utils/bulk_loader` → `${GITHUB_WORKSPACE}/build/tools/utils/bulk_loader`
- `cd .../tools/python_bind/build/neug_py_bind` (ctest, cmake --install, make coverage) → `cd ${GITHUB_WORKSPACE}/build`
- `tools/python_bind/build/neug_py_bind/coverage_filtered.info` → `build/coverage_filtered.info`
- PYTHONPATH: dropped stale `tools/python_bind/build/lib` (neug loader now auto-discovers the .so under `<repo>/build/tools/python_bind/`).

### DIFF
--- a/.github/issue-triage-config.yml
+++ b/.github/issue-triage-config.yml
@@ -1,0 +1,150 @@
+# Issue triage configuration.
+#
+# Each umbrella's scope_in/scope_out is the **LLM context**. Keep them concise
+# and disjoint — that's what makes the classifier accurate. Do NOT paste the
+# umbrella's full GitHub body here (it usually contains a child-issue list
+# which would leak labels during backtesting).
+
+milestone: "NeuG v0.2"
+
+# Minimum confidence required to auto-link as sub-issue (without human approval).
+# "high" = unambiguous fit. Recommended starting value: never auto-link
+# (set to "never") for the first 1-2 weeks. After backtest accuracy >= 80%,
+# raise to "high".
+auto_link_threshold: "high"   # one of: never | high | medium
+
+umbrellas:
+  - number: 49
+    title: "Bundle some almost ready features"
+    scope_in: |
+      Mid-to-large features that are nearly ready and being bundled into the v0.2
+      release. Typically external-contributor or experimental features being shaped
+      into a shippable bundle.
+    scope_out: |
+      Small bug fixes, refactors, or single-PR features. Targeted ingestion / storage
+      / transaction work (those have dedicated umbrellas).
+
+  - number: 86
+    title: "Refactor UpdateTransaction"
+    scope_in: |
+      Transaction layer: COW (copy-on-write), WAL (write-ahead log), rollback,
+      concurrency isolation, atomicity. Includes UpdateTransaction concurrency
+      design, WAL mechanics, transaction rollback for inserts/updates, schema
+      forkability for COW, IngestWal logic, AND API shape changes for the
+      insert / update / write-side surface (parameter types, method signatures,
+      Value-vs-positional API). Even small / one-line API tweaks belong here
+      if they live on the UpdateTransaction surface.
+    scope_out: |
+      Persistence / checkpoint / dump / load / mmap (those go to #272).
+
+  - number: 97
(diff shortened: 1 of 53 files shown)


Spockkk0225 (dismissed):
LGTM. I will update nodejs binding after this pr is merged.

zhanglei1949:
The CI failure is not related to this PR: https://github.com/alibaba/neug/actions/runs/26934266474/job/79461492639?pr=468, related issue #479 
But we still fix it in this PR.

Spockkk0225 (approved):

PR #473, opened 2026-06-03
refactor: WAL append returns false on I/O failure instead of throwing (#385)

### LINKED ISSUE #385: Writing WAL should return bool, rather than throw exceptions.
During WAL append operations, file write failures should not trigger exceptions. The function should return false to propagate the error to higher-level components (e.g., UpdateTransaction) for proper handling

zhanglei1949:
## Summary
- `LocalWalWriter::append` declared `bool` but threw `IOException` from `ftruncate` / `write` / `fsync` failure paths. Its callers (`UpdateTransaction::Commit`, `InsertTransaction::Commit`, `CompactTransaction::Commit`) only branch on the bool return and do not catch exceptions, so a real I/O failure unwound past the `Abort()` cleanup and left timestamps un-invalidated.
- Replace the `THROW_IO_EXCEPTION` calls in `append()` with `LOG(ERROR) + return false` so the existing caller error paths run.
- Drop a dead non-Apple branch under `F_FULLFSYNC` that contained a `strerrno` typo (`F_FULLFSYNC` is Apple-only, so that branch was unreachable).

Fixes #385.

### DIFF
--- a/src/transaction/wal/local_wal_writer.cc
+++ b/src/transaction/wal/local_wal_writer.cc
@@ -81,8 +81,8 @@ bool LocalWalWriter::append(const char* data, size_t length) {
   if (expected_size > file_size_) {
     size_t new_file_size = (expected_size / TRUNC_SIZE + 1) * TRUNC_SIZE;
     if (ftruncate(fd_, new_file_size) != 0) {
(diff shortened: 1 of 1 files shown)


zhanglei1949:
Closed due to lack of consensus.

PR #475, opened 2026-06-04
refactor: Switch d0_set from vertex_array_t<bool> in EdgeExpand::tc

### LINKED ISSUE #474: replace bitset with flat_hash_set in EdgeExpand::tc
The triangle-counting kernel in EdgeExpand::tc previously used a vertex_array_t<bool> sized to |V(d0_nbr_label)| to mark first-hop matches.  This bitset was reset (re-init or per-element clear) for every input vertex, which has two costs on large graphs:

  1. Memory: O(|V(d0_nbr_label)|) bits resident per worker thread, even when the actual matched-set is sparse (typical LDBC SNB  fanouts emit only a few hundred to a few thousand matches per source).

  2. Cache: every membership test (d0_set[u]) and clear-loop touches  bytes scattered across a multi-MB region, missing L1/L2 on virtually every probe and polluting the cache lines used by the CSR traversal in the same inner loop.

liulx20:
Fixes


zhanglei1949 (approved):
LGTM

PR #485, opened 2026-06-04
chore: automatically detect openssl on macos

### LINKED ISSUE #484: `Homebrew` has various installation path on macos_x86_64 and macos_arm64
brpc finds openssl in `/usr/local/opt/openssl`, which is a macos_x86_64 path:
```
    if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND NOT OPENSSL_ROOT_DIR)
        set(OPENSSL_ROOT_DIR
            "/usr/local/opt/openssl" # Homebrew installed OpenSSL
        )
    endif()
```

But in macos_arm64, it often appears under `/usr/local/Cellar/openssl@3/`, which is not mentioned in the document.

Use `brew --prefix openssl@3` to automatically detect and write the directory to `OPENSSL_ROOT_DIR`.

Spockkk0225:
While building on macos, it will detect real OpenSSL path and set OPENSSL_ROOT_DIR to brpc.
```
# macOS: auto-detect Homebrew OpenSSL path (arm64 uses /opt/homebrew, x86_64 uses /usr/local)
if (APPLE AND NOT OPENSSL_ROOT_DIR)
    execute_process(
        COMMAND brew --prefix openssl@3
        OUTPUT_VARIABLE BREW_OPENSSL_PREFIX
        OUTPUT_STRIP_TRAILING_WHITESPACE
        ERROR_QUIET
    )
    if (BREW_OPENSSL_PREFIX AND EXISTS "${BREW_OPENSSL_PREFIX}")
        set(OPENSSL_ROOT_DIR "${BREW_OPENSSL_PREFIX}")
        message(STATUS "Using Homebrew OpenSSL: ${OPENSSL_ROOT_DIR}")
    endif()
endif()
```

fix #484


CMakeLists.txt
```diff
@@ -470,6 +470,21 @@ if(HAVE_COPY_FILE_RANGE_COMPILES)
 endif()
 
 find_package(Threads REQUIRED)
+
+# macOS: auto-detect Homebrew OpenSSL path (arm64 uses /opt/homebrew, x86_64 uses /usr/local)
```
zhanglei1949: There seems to be multiple `find_package(OpenSSL)` in our CMakeLists, maybe we should replace all `find_package(OpenSSL)` with a new macro, which ensure OpenSSL in correctly found. 
https://github.com/search?q=repo%3Aalibaba%2Fneug+find_package%28OpenSSL&type=code

zhanglei1949 (changes requested):

zhanglei1949 (dismissed):
LGTM

CMakeLists.txt
```diff
@@ -470,6 +470,21 @@ if(HAVE_COPY_FILE_RANGE_COMPILES)
...
+if (APPLE AND NOT OPENSSL_ROOT_DIR)
```
zhanglei1949: Maybe we should do find_package(OpenSSL) if `OPENSSL_ROOT_DIR` 1is not specified, then set the prefix to OPENSSL_ROOT_DIR. This should be done in `BuildBrpcxxx.cmake`

zhanglei1949 (approved):
LGTM

PR #499, opened 2026-06-05
fix: Reset timestamp and update batch sort arguments

liulx20:
Fixes

### DIFF
--- a/src/storages/graph/edge_table.cc
+++ b/src/storages/graph/edge_table.cc
@@ -925,16 +925,16 @@ void EdgeTable::Compact(bool compact_csr,
     out_csr_->compact();
     in_csr_->compact();
   }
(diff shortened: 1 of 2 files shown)


src/storages/graph/edge_table.cc
```diff
@@ -925,16 +925,16 @@ void EdgeTable::Compact(bool compact_csr,
     out_csr_->compact();
     in_csr_->compact();
   }
+  out_csr_->reset_timestamp();
```
zhanglei1949, line 930: Could a comment be added here, to reminding developer before changing calling the order of `rest_timestamp and `batch_sort_by_edge_data`.

zhanglei1949 (dismissed):
LGTM

zhanglei1949 (approved):
LGTM

PR #524, opened 2026-06-10
fix: Add future annotations import to setup.py

### LINKED ISSUE #523: [BUG] setup.py uses Python 3.9+ type annotation syntax, breaks on Python 3.8
```bash
$ make build
Traceback (most recent call last):
File "setup.py", line 98, in <module>
class CMakeBuild(build_ext):
File "setup.py", line 190, in CMakeBuild
def collectcmake_args(self) -> tuple[list[str], list[str]]:
TypeError: 'type' object is not subscriptable
```

Environment
Python: 3.8.10
OS: Ubuntu 20.04

liulx20:
Fixes #523


zhanglei1949 (approved):
LGTM

PR #536, opened 2026-06-11
refactor: Avoid dump to new file when nbr_list not modified

### LINKED ISSUE #535: perf: skip full nbr_list serialization in MutableCsr::Dump when no writes occurred
`MutableCsr::Dump` currently performed full manual serialization of `nbr_list` on every checkpoint — iterating all vertices, writing adjacency data, and computing an MD5 checksum — even when no edges had been modified. Meanwhile, `degree_list_` and `cap_list_` already benefited from the container-level `IsDirty()` check in `Checkpoint::Commit()`.

The `nbr_list` cannot rely on the container-level dirty check alone because `put_edge` may allocate new buffers via `Allocator` (pointer escaping the mmap region), making the container appear clean even though the logical data has changed. An application-level flag bridges this gap.

zhanglei1949:
Fix #535

### DIFF
--- a/src/storages/csr/mutable_csr.cc
+++ b/src/storages/csr/mutable_csr.cc
@@ -84,7 +84,27 @@ void MutableCsr<EDATA_T>::Open(Checkpoint& ckp,
         " but degree list implies " + std::to_string(edge_count) +
         ", desc: " + descriptor.ToJsonString());
   }
-  nbr_data_dirty_ = false;
+}
+
+template <typename NBR_T>
+bool is_nbr_list_unmodified(MD5_CTX& ctx, FileHeader& header,
+                            const IDataContainer* nbr_container,
+                            const NBR_T* const* adj_lists, const int* cap_arr,
+                            size_t vnum) {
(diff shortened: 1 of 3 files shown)


src/storages/csr/mutable_csr.cc
```diff
@@ -221,6 +233,7 @@ void MutableCsr<EDATA_T>::resize(vid_t vnum) {
       cap_arr[i] = 0;
     }
     locks_ = std::make_unique<SpinLock[]>(vnum);
+    nbr_data_dirty_ = true;
   } else {
     adj_list_buffer_->Resize(vnum * sizeof(nbr_t*));
```
liulx20, line 245: Should the dirty bit be reset in this else branch?

zhanglei1949: if v_num <= old_num, we should not change nbr_data_dirty_

liulx20 (dismissed):

liulx20 (approved):

PR #555, opened 2026-06-16
refactor: Throw exception rather than LOG(FATAL) when encounter IO error

### LINKED ISSUE #514: [BUG] mmap fails during WAL recovery - process aborts via LOG(FATAL)
**Describe the bug**
`LocalWalParser::open()` calls `mmap()` to map WAL files into memory. When `mmap` returns `MAP_FAILED`, the code calls `LOG(FATAL)`, which may trigger `abort()` and terminate the process. The caller wraps recovery logic in `try/catch`, but `abort()` is not a C++ exception and cannot be caught. This could make the database unable to start after a single `mmap` failure.

**To Reproduce**
1. Create a database and perform some writes.
2. Corrupt a WAL file (e.g. `truncate` it).
3. Call `NeugDB::Open()` to reopen.
4. The process may crash with `SIGABRT`.

**Expected behavior**
`mmap` failure should be handled — throw an exception or return an error, rather than aborting the process.

**Code location**

`src/transaction/wal/local_wal_parser.cc`

```cpp
if (mmapped_buffer == MAP_FAILED) {
  LOG(FATAL) << "mmap failed...";
}
```

 - NeuG version: [master]
 - OS: [Independent]
 - Python Version [3.12]

zhanglei1949:
Fix #514

### DIFF
--- a/src/transaction/wal/local_wal_parser.cc
+++ b/src/transaction/wal/local_wal_parser.cc
@@ -19,6 +19,8 @@
 #include <sys/mman.h>
 #include <unistd.h>
 #include <algorithm>
+#include <cerrno>
+#include <cstring>
 #include <filesystem>
 #include <ostream>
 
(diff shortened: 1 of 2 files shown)


src/transaction/wal/local_wal_parser.cc
```diff
@@ -47,9 +48,16 @@ void LocalWalParser::open(const std::string& wal_uri) {
       continue;
     }
     int fd = ::open(path.c_str(), O_RDONLY);
-    void* mmapped_buffer = mmap(NULL, file_size, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (fd == -1) {
+      close();
+      THROW_IO_EXCEPTION("Failed to open wal file: " + path);
```
liulx20: can we expose error information using operations like strerror(errno)？

zhanglei1949: resolved.

liulx20 (approved):

PR #557, opened 2026-06-16
fix: Fix swallowed overflow error for label creation

### LINKED ISSUE #556: [BUG] Fail to handle the overflow of label_t for label creation
Currently, the overflow of label_t is swallow by NeuG. We should throw exception.

zhanglei1949:
Fix #556


longbinlai (approved):
LGTM

PR #563, opened 2026-06-17
fix: propagate test failures in npm-build-test workflow

### LINKED ISSUE #562: [BUG] macOS ARM64 tests fail with BigInt mismatch and CI does not propagate test failures
**Describe the bug**
The Node.js binding test suite has two issues that cause macOS ARM64 CI to fail while silently passing on Linux:

1. **CI test runner does not propagate failures**: The test loop is:
   ```bash
   for f in tests/test_*.js; do echo "=== $f ===" && node "$f"; done
   ```
   Bash `for` loop returns the exit code of the **last** iteration. If any middle iteration fails while the last one passes, the error is silently swallowed.

2. **BigInt type mismatch in test assertions**: Node.js numbers are float numbers. The binding correctly returns `BigInt` (e.g. `1n`) for INT64 columns, but the tests use plain `Number` (e.g. `1`) in `assert.deepEqual`, causing `1n !== 1` failures.

**Error Message**
```
not ok - test_auto_enable_checkpoint
  AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  [
    [
+     1n,
-     1,
      'Alice',
      30
    ],
    [
+     2n,
-     2,
      'Bob',
      25
    ]
  ]
```

Spockkk0225:
## Summary

The `for` loop in the Test step of `npm-build-test.yml` returns the exit code of the **last** iteration only:

```bash
for f in tests/test_*.js; do echo "=== $f ===" && node "$f"; done
```

If a middle test file (e.g. `test_db_transaction.js`) fails while the last one (`test_query_result.js`) passes, the overall exit code is 0 and the CI step is incorrectly marked as success.

## Fix

Track failures with a `FAILED` variable and exit with its value:

```bash
FAILED=0; for f in tests/test_*.js; do echo "=== $f ===" && node "$f" || FAILED=1; done; exit $FAILED
```

This applies to all three platform jobs: linux-x86_64, linux-arm64, and macos-aarch64.

fix #562


lnfjpt (approved):
LGTM

PR #574, opened 2026-06-20
feat: add shortest path return to BFS and SSSP algorithms

longbinlai:
## Summary

Add shortest path return to GDS BFS and SSSP algorithms, plus path serialization optimizations that benefit all path-returning queries.

### Commit 1: feat(gds): add shortest path return to BFS and SSSP algorithms
- `path` output column via YIELD clause (pure YIELD-based, no config options)
- Predecessor tracking in BFS/SSSP/BFSPred/SSSPPred (O(V) memory)
- Real edge data lookup from CSR in path reconstruction
- Struct-wrapped DataType for kPath compatibility with type converter
- Zero overhead when path is not YIELDed
- All changes confined to `extension/gds/` — no system code modified

### Commit 2-3: perf: optimize path JSON serialization
- Eliminate serialize→parse→copy string round-trip in `convert_path_to_json` (sink.cc)
- Add direct rapidjson build helpers: `build_vertex_json_value` / `build_edge_json_value`
- Backward compatible: `convert_vertex_to_json` / `convert_edge_to_json` unchanged for standalone callers

### Commit 4: feat(gds): add path_properties configuration
- **Lightweight mode (default)**: Only _ID, _LABEL, PK for nodes; _ID, _LABEL, _SRC_ID, _DST_ID for edges
  - Performance: ~0.54s on LDBC SNB SF10 (65K nodes, 1.9M edges, 62K paths)
- **Full mode** (`path_properties: 'full'`): All vertex and edge properties
  - Performance: ~1.26s (2.34x slower)
- Thread-local encoding flag in sink.cc with public API in sink.h

## Usage

```cypher
-- Distance only (backward compatible, zero overhead)
CALL bfs('g', {source: '0'}) YIELD node, distance RETURN node.id, distance;

-- With path return (lightweight, default)
CALL bfs('g', {source: '0'}) YIELD node, distance, path RETURN node.id, distance, path;

-- With path return (full properties)
CALL bfs('g', {source: '0', path_properties: 'full'}) YIELD node, distance, path RETURN node.id, distance, path;
```

## Testing

| Suite | Tests | Result |
|-------|-------|--------|
| GDS path return (test_gds_path.py) | 21 | ✅ all pass |
| GDS path encoding modes | 5 | ✅ all pass |
| LDBC SF10 path tests (test_gds_path_ldbc.py) | 18 | ✅ all pass |
| Existing GDS tests (test_gds.py) | 36 | ✅ all pass |
| LDBC Graphalytics conformance (test_graphalytics.py) | 24 | ✅ all pass |
| Other test suites (test_db_cases, test_query, etc.) | 300+ | ✅ no regressions |

## Performance (LDBC SNB SF10: 65K nodes, 1.9M edges, 62K paths)

| Configuration | Time | vs Baseline |
|--------------|------|-------------|
| BFS no path | 0.025s | — |
| BFS + path (lightweight) | 0.54s | 6.5x from no-path |
| BFS + path (full) | 1.26s | 2.34x from lightweight |

## Files Changed

- `extension/gds/` — 12 files (path return implementation + configuration)
- `include/neug/execution/common/operators/retrieve/sink.h` — encoding mode API
- `src/execution/common/operators/retrieve/sink.cc` — serialization optimization
- `tools/python_bind/tests/` — 3 new test files, 21 new tests


extension/gds/src/impl/bfs_impl.cc
```diff
@@ -82,6 +94,9 @@ void BFS::compute() {
               if (__atomic_compare_exchange_n(&distances_[dst], &expected,
                                               level, false, __ATOMIC_RELAXED,
                                               __ATOMIC_RELAXED)) {
+                if (return_path_) {
```
liulx20: I'm not sure whether there are multi-threading concurrency issues here. A more reasonable approach would be: after the BFS completes, re-traverse from each node—for example, starting from nodes at depth d to find their neighbors at depth d-1, continuing until reaching the target node.

extension/gds/src/impl/sssp_pred_impl.cc
```diff
@@ -118,6 +129,9 @@ void SSSPPred::compute() {
       double cand = dist + weight;
       if (distances_[w] < 0 || cand < distances_[w]) {
         distances_[w] = cand;
+        if (return_path_) {
```
liulx20: Similar to BFS, it may be difficult to trace back predecessor nodes in SSSP. A simple solution is to implement a single-threaded version to maintain the predecessors.

liulx20: consider routing the case that returns a path through the predicate branch, since it is inherently single-threaded.

extension/gds/src/impl/bfs_impl.cc
longbinlai: Thanks for pointing this out. You are right that the dense pull mode had a race condition on `predecessors_`.

**Fix applied** (commit 3dbc7588): In the dense pull mode, I added an atomic CAS on `distances_[dst]` before writing `predecessors_[dst]`. The flow is now:
1. Check if the vertex is unvisited (`distances_[dst] == max`)
2. Scan neighbors for a predecessor at `level - 1`
3. If reachable, CAS `distances_[dst]` from `max` to `level` — only the winning thread writes `predecessors_[dst]`

This ensures only one thread can claim each vertex and set its predecessor, eliminating the race. The sparse push mode was already safe (CAS on distance was already there), so no change was needed there.

extension/gds/src/impl/sssp_pred_impl.cc
longbinlai: Good point. Adopted your suggestion — when path return is requested (`return_path=true`), the dispatcher in `sssp.cc` now routes to `SSSPPred` (single-threaded Dijkstra) regardless of whether predicates are present. This completely avoids the concurrency hazard of maintaining predecessor chains in the parallel SSSP, where iterative relaxation by multiple threads makes predecessor tracking non-trivial.

longbinlai: Exactly what I did — the dispatcher in `sssp.cc` now checks `return_path` and routes to the predicate branch (`SSSPPred`, single-threaded Dijkstra) when path return is requested. This is inherently safe since there is no concurrent access to the predecessor array. All 95 Python tests pass.

tools/python_bind/tests/test_gds_path_ldbc.py
```diff
@@ -0,0 +1,694 @@
...
+import tempfile
+import time
+
+import pytest
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from neug.database import Database
+
+DATA_DIR = os.path.expanduser("~/Downloads/social_network_10_interactive")
```
liulx20: Should we avoid using absolute paths here?

src/execution/common/operators/retrieve/sink.cc
```diff
@@ -36,6 +36,19 @@
 namespace neug {
 namespace execution {
 
+// Path encoding mode: full (default) or lightweight
+// - full: encode all properties (backward compatible default)
+// - lightweight: only encode _ID, _LABEL, PK for vertices; structural info for
+// edges
+// GDS extension can call set_path_full_encoding(false) for lightweight mode
+static thread_local bool path_full_encoding_enabled = true;
```
liulx20: What is the scope of this, and can it be declared outside of sink.cc?

extension/gds/src/sssp.cc
```diff
@@ -99,37 +114,43 @@ execution::Context SSSPFunction::exec(const function::CallFuncInputBase& input,
...
   function::function_set func_set;
-  std::vector<common::DataTypeId> input_types = {
-      common::DataTypeId::kVarchar, common::DataTypeId::kUnknown};
+  std::vector<common::DataTypeId> input_types = {common::DataTypeId::kVarchar,
+                                                 common::DataTypeId::kUnknown};
   function::call_output_columns output_columns = {
       {"node", common::DataTypeId::kVertex},
-      {"distance", common::DataTypeId::kDouble}};
+      {"distance", common::DataTypeId::kDouble},
+      {"path", common::DataTypeId::kPath}};
```
liulx20, line 133: The function declaration here returns the column for the path, yet it doesn't return anything when return path is false, which seems odd. Is there a better way to handle this?

extension/gds/include/impl/bfs_impl.h
```diff
@@ -27,10 +27,12 @@ namespace gds {
...
  public:
   BFS(const StorageReadInterface& graph, label_t vertex_label,
-      label_t edge_label, vid_t source, bool directed, int concurrency);
+      label_t edge_label, vid_t source, bool directed, int concurrency,
+      bool return_path = false);
 
   void compute();
-  void sink(execution::Context& ctx, int node_alias, int distance_alias);
+  void sink(execution::Context& ctx, int node_alias, int distance_alias,
+            int path_alias = -1);
```
liulx20: Actually, -1 is a valid alias, so we probably shouldn't use -1 to indicate whether the path should be output.

src/execution/common/operators/retrieve/sink.cc
liulx20: this variable now affects the sink behavior of other queries, which is unreasonable.

extension/gds/src/sssp.cc
```diff
@@ -99,37 +114,43 @@ execution::Context SSSPFunction::exec(const function::CallFuncInputBase& input,
...
+                  sssp_input.edge_pred.get(), sssp_input.return_path);
     sssp.compute();
-    sssp.sink(ret, sssp_input.node_alias, sssp_input.distance_alias);
+    sssp.sink(ret, sssp_input.node_alias, sssp_input.distance_alias,
+              sssp_input.path_alias);
   } else {
     SSSP sssp(graph, sssp_input.vertex_label, sssp_input.edge_label, source_vid,
               sssp_input.directed, sssp_input.edge_weight,
-              sssp_input.concurrency);
+              sssp_input.concurrency, sssp_input.return_path);
```
liulx20: Since the else branch does not handle return_path, the logic related to return_path can be removed.

extension/gds/src/impl/bfs_impl.cc
longbinlai: Good suggestion! We've adopted exactly this approach in the latest commit. Predecessor arrays are now completely removed from BFS computation. Paths are reconstructed post-hoc in `sink()` by walking the distance array: for vertex v at distance d, scan neighbors to find one at distance d-1, repeat until reaching source. This eliminates the concurrency hazard entirely — no CAS needed, no predecessor writes during parallel computation. The dense pull mode has been reverted to the original simple non-atomic `distances_[dst] = level`.

extension/gds/src/impl/sssp_pred_impl.cc
longbinlai: Addressed. SSSPPred no longer stores predecessors during computation. Path reconstruction is done post-hoc in `sink()` by walking the distance array: for vertex v with distance D, find a neighbor u where `dist(u) + weight(u→v) == dist(v)`. This is single-threaded and avoids the difficulty of maintaining predecessor chains during parallel relaxation.

longbinlai: Yes, we route return_path cases through SSSPPred (single-threaded Dijkstra) as suggested. SSSPPred now also uses post-hoc path reconstruction instead of a predecessor array, so the routing is kept for the predicate case and the path return case. The parallel SSSP is only used when no predicates and no path return.

tools/python_bind/tests/test_gds_path_ldbc.py
longbinlai: Fixed. Replaced the hardcoded `~/Downloads/...` path with an environment variable `NEUG_LDBC_DATA_DIR`. The test auto-skips when the variable is not set.

src/execution/common/operators/retrieve/sink.cc
longbinlai: The global variable has been completely removed. The encoding mode is now carried per-column: `PathColumnBuilder` has a `set_full_encoding(bool)` setter, `PathColumn` carries the flag, and `sink.cc`'s `add_column()` reads it from the column when serializing. GDS sink() methods receive `full_encoding` as a parameter and set it on the builder. Non-GDS path columns default to `true` (full encoding), so other queries are never affected.

extension/gds/src/sssp.cc
longbinlai: The 3-column output declaration (node, distance, path) represents the function's schema — the maximum set of columns the function CAN produce. The YIELD clause selects which columns are actually returned. When the user doesn't YIELD path, `return_path` is false and the path column is simply not populated in the output Context. This is standard Cypher CALL behavior — the output columns declare what's available, not what's always produced. We kept this design as changing it would require modifications to the function binding infrastructure.

extension/gds/include/impl/bfs_impl.h
longbinlai: Fixed. The `path_alias >= 0` check has been removed. All sink() methods now use `return_path_` as the sole indicator for whether to build the path column. When `return_path_` is true, `path_alias` is always valid (set by the binder).

src/execution/common/operators/retrieve/sink.cc
longbinlai: Fully resolved. The `static thread_local bool path_full_encoding_enabled` variable has been completely removed from `sink.cc`. The encoding mode is now carried as a per-column member on `PathColumn`/`PathColumnBuilder`, so GDS path queries with `path_properties: 'lightweight'` only affect their own path column. Other queries' path columns always default to full encoding and are never affected.

extension/gds/src/sssp.cc
longbinlai: Done. The else branch no longer passes `return_path` or `path_alias` to the parallel SSSP. The `return_path_` and `predecessors_` members have been completely removed from the SSSP class, along with the dead `if (return_path_)` branches in `compute()` and `sink()`. The parallel SSSP constructor no longer takes a `return_path` parameter.

doc/source/extensions/load_gds.md
```diff
@@ -170,6 +170,7 @@ RETURN node, distance;
 |---|---|---|---|
 | `source` | STRING | *(required)* | The source node's primary key value |
 | `directed` | BOOL | `false` | Whether to follow edges in their stored direction only |
+| `path_properties` | STRING | `"full"` | Path encoding mode: `"full"` (all properties, default) or `"lightweight"` (structure only) |
```
liulx20: The `path_properties` actually describes sink behavior and is unrelated to the algorithm; therefore, it seems inappropriate to include it as an algorithm option.

shirly121:
`wrapTableBindFuncWithPathFix` and `buildPathDataType` have been removed. Path type is built during compiling phase, which is implemented in commit https://github.com/alibaba/neug/pull/574/changes/5065a102fe6fcd715ac3dd4c1c4e9cdc8a1b25de

longbinlai:
## Crash fix: prevent segfault when graph is modified after project_graph

Added a fix for issue #599 — GDS algorithms crash the process with a segfault when the graph schema is modified after `CALL project_graph()`.

### Root cause

`project_graph` only stores table-name strings, not a snapshot. When the schema changes (drop/recreate edge tables, alter properties), the projected graph becomes stale. The GDS `exec()` function did not validate that labels still exist, leading to:
- Null pointer dereferences in `GetVertexSet` / `GetGenericOutgoingGraphView`
- `LOG(FATAL)` (→ `abort()`) in `EmptyCsr::get_generic_view()` when the edge CSR is replaced
- Uncaught exceptions from the storage layer

### Fix (commit 80792a9c)

1. **`validate_projected_graph()`** — checks vertex/edge labels exist in the current schema before executing the algorithm. Returns false + logs error if labels are out of range.
2. **try-catch in `exec()`** — wraps BFS and SSSP execution in try-catch to gracefully handle storage layer exceptions. Returns empty Context on error.
3. **Graceful degradation** — the query fails with an error message instead of killing the process.

### Known limitation

`LOG(FATAL)` in `EmptyCsr::get_generic_view()` (`mutable_csr.h:403`) still calls `abort()` and cannot be caught by try-catch. This is a separate issue in the core storage layer (#599).

liulx20 (dismissed):
LGTM

liulx20 (approved):

PR #586, opened 2026-06-22
refactor: replace Arrow CSV/JSON IO with native reader framework

luoxiaojian:


src/utils/reader/csv_reader.cc
```diff
@@ -0,0 +1,639 @@
...
+  };
+  auto is_interval_token = [](const std::string& s) -> bool {
+    // Simple heuristic: contains "year" or "month" or "day" or "hour"
+    std::string lower;
+    lower.reserve(s.size());
+    for (char c : s)
+      lower.push_back(static_cast<char>(std::tolower(c)));
+    return lower.find("year") != std::string::npos ||
+           lower.find("month") != std::string::npos ||
+           lower.find("day") != std::string::npos ||
```
longbinlai: **Bug: `is_interval_token` heuristic is too broad — weekday names are misclassified as interval type**

The substring check matches any string containing "year", "month", "day", or "hour". This causes false positives for common string values:

- `"Monday"` contains `"day"` → matches
- `"Tuesday"` contains `"day"` → matches
- `"Wednesday"` contains `"day"` → matches
- `"hourly"` contains `"hour"` → matches
- `"birthday"` contains `"day"` → matches

**Reproduction:**

Create a CSV file `weekdays.csv`:
```csv
id,weekday
1,Monday
2,Tuesday
3,Wednesday
4,Thursday
5,Friday
```

Then in schema-flexible mode:
```cypher
COPY weekday_test FROM (LOAD FROM "weekdays.csv" RETURN *);
```

The `weekday` column will be inferred as `kInterval` instead of `kVarchar`, because all values match the `is_interval_token` heuristic. This causes parse errors or silent data corruption when the interval parser tries to interpret "Monday" as an interval value.

**Suggested fix:** Use a stricter pattern, e.g., ISO 8601 duration format (`PnYnMnDTnHnMnS`) or require the token to start with `P` followed by digits and unit specifiers.

src/storages/loader/loader_utils.cc
```diff
@@ -15,66 +15,420 @@
...
+  csv::CSVFormat csv_format;
+  csv_format.delimiter(config.delimiter);
+  if (config.quoting) {
+    csv_format.quote(config.quote_char);
   } else {
-    read_options.skip_rows = 0;
+    csv_format.quote(false);
+  }
+  csv_format.variable_columns(csv::VariableColumnPolicy::KEEP);
+  csv_format.no_header();
```
longbinlai, line 67: **Bug: escaping + quoting combination is broken — `escape_char` is not passed to csv-parser**

`build_csv_format()` does not configure `escape_char` on `csv::CSVFormat`. The csv-parser library does not have an `escape(char)` API, so backslash escaping inside quoted fields is not handled at the parser level. The post-parse `unescape_token()` cannot fix this because csv-parser has already mis-split the fields.

**Reproduction:**

Create `escaped.csv` with quoting + backslash escaping enabled:
```csv
id,message
1,"hello\"world"
2,"say \"hi\""
```

Run with `quoting=true` and `escaping=true`:
```cypher
COPY test FROM (LOAD FROM "escaped.csv" (quoting=true, escaping=true, quote=", escape=\) RETURN *);
```

**Expected:** `message` column contains `hello"world` and `say "hi"`.

**Actual:** csv-parser sees `\"` inside the quoted field and closes the quote at the `"` after `\`. The field is parsed as `hello\` instead of `hello"world`. The subsequent `unescape_token()` call cannot recover the lost data.

**Root cause:** Arrow's CSV reader handled quoting and escaping together at the parser level. The csv-parser library only supports double-quote escaping (`""` = literal `"`), not backslash escaping inside quoted fields.

**Suggested fix:** Either:
1. Document that backslash escaping inside quoted fields is not supported (only double-quote escaping works), or
2. Pre-process the file to replace `\"` with `""` before passing to csv-parser when both quoting and escaping are enabled.

src/utils/reader/csv_reader.cc
luoxiaojian: 一种可能是不推断interval类型，必须显示指定（对齐duckdb）
另一种是用严格的Interval::fromCString try-parse 验证
我倾向于前者，你觉得呢？

src/storages/loader/loader_utils.cc
luoxiaojian: arrow对于escaping和qouting也有bug：https://github.com/apache/arrow/issues/33405

luoxiaojian: 建议按方案1，我在escaping和qouting都enable的场景下加一个warning，你觉得呢？

src/utils/reader/csv_reader.cc
longbinlai: @shirly121 这就是早上和你讨论的 类似 interval 这种类型的推断问题了。你也review 看看哈

shirly121: 我感觉可以通过显示指定，这个语法上可以通过 CAST 支持，LOAD FROM xxx.csv Return CAST(interval_column, 'INTERVAL'); 
但有几个问题：
1. 需要定义具体边界，哪些类型可以 try-parse，哪些类型必须显示指定
2. 之前 arrow 版本在支持 interval 时，存在一些隐式转换，interval 在 arrow 层推断为 string，schema 里指定为 INTERVAL，在存储层会隐式的将 string -> interval，这种隐式转换还需要保留吗？

src/utils/reader/csv_reader.cc
```diff
@@ -0,0 +1,639 @@
...
+  }
+
+  ReadOptions readOpts;
+  const bool autogenerate =
+      readOpts.autogenerate_column_names.get(sharedState_->schema.file.options);
+
+  if (config.column_names.empty() && !autogenerate) {
+    config.column_names = read_header(paths[0], config);
+  } else if (config.column_names.empty() && autogenerate) {
+    std::ifstream input(paths[0]);
```
shirly121, line 610: 之前 csv/json reader 依赖 arrow 支持了远程读取，我们现在的方案是只支持本地？

luoxiaojian: 这个在后续的pr里会增加对httpfs、s3的读取支持，这个改动的体量已经足够大了

src/utils/reader/csv_reader.cc
```diff
@@ -0,0 +1,639 @@
...
+
+  const auto& fileSchema = sharedState_->schema.file;
+  ReadOptions readOpts;
+  const bool use_batch_read = readOpts.batch_read.get(fileSchema.options);
+
+  auto read_config = read_config_for_supplier(config);
+  if (sharedState_->skipRows) {
+    // Need all columns to evaluate row-filter expression;
+    // full_read will project afterwards.
+    read_config.include_columns = config.column_names;
```
shirly121, line 519: batch_read=true 情况下，直接用文件 columns 作为 include_columns，会导致 load from 输出列和 schema 不一致，目前是有 segment fault 报错，查询例子：
```python
    def test_copy_from_load_from_where_with_batch_read(self):
        """Test COPY FROM LOAD FROM subquery applies WHERE when batch_read is enabled."""
        csv_path = os.path.join(self.tinysnb_path, "vPerson.csv")
        if not os.path.exists(csv_path):
            pytest.skip(f"CSV file not found: {csv_path}")

        create_schema = """
        CREATE NODE TABLE person_filtered_batch (
            ID INT64,
            age INT64,
            fName STRING,
            gender INT64,
            isStudent BOOLEAN,
            PRIMARY KEY (ID)
        )
        """
        self.conn.execute(create_schema)

        copy_query = f"""
        COPY person_filtered_batch FROM (
            LOAD FROM "{csv_path}" (header=true, delimiter=",", batch_read=true)
            WHERE age > 20
            RETURN ID, age, fName, gender, isStudent
        )
        """
        self.conn.execute(copy_query)

        query = "MATCH (p:person_filtered_batch) RETURN p.ID, p.age ORDER BY p.ID"
        result = self.conn.execute(query)
        records = list(result)

        assert records == [
            [0, 35],
            [2, 30],
            [3, 45],
            [8, 25],
            [9, 40],
            [10, 83],
        ]
```

src/utils/reader/csv_reader.cc
```diff
@@ -0,0 +1,639 @@
...
+  output.append_chunk(std::move(projected));
+}
+
+void CsvReader::batch_read(
+    const std::vector<std::shared_ptr<IDataChunkSupplier>>& suppliers,
+    execution::Context& output) {
+  output.clear();
+  for (const auto& supplier : suppliers) {
+    while (auto chunk = supplier->GetNextChunk()) {
+      output.append_chunk(std::move(*chunk));
```
shirly121, line 581: batch_read=true 跳过了 filtering 和 project，有可能导致结果不正确：

```cypher
      COPY person_filtered_batch FROM (
          LOAD FROM "{csv_path}" (header=true, delimiter=",", batch_read=true)
          WHERE age > 20
          RETURN ID, age, fName, gender, isStudent
      )
```

src/utils/reader/json_reader.cc
```diff
@@ -0,0 +1,555 @@
...
+    return execution::Value::STRING(rapidjson_stringify(value));
+  }
+}
+
+class JsonChunkSupplier : public IDataChunkSupplier {
+ public:
+  JsonChunkSupplier(const std::string& file_path, JsonReadConfig config)
+      : file_path_(file_path), config_(std::move(config)) {
+    if (config_.json_array_input) {
+      lines_ = convert_json_array_to_lines(file_path_);
```
shirly121: JsonArray 会对每个object 执行两次 string parse: 先将每个 object 序列化回 string，保存为lines，然后在 GetNextChunk 里又重新 parse 每个 string。是不是可以直接存储 DOM node，避免两次string parse?

src/utils/reader/csv_reader.cc
luoxiaojian: 移除了对interval类型的推导

src/storages/loader/loader_utils.cc
luoxiaojian: 当escaping和qouting都enable的场景下加了warning

src/utils/reader/csv_reader.cc
luoxiaojian: fixed

src/utils/reader/csv_reader.cc
luoxiaojian: fixed

src/utils/reader/json_reader.cc
luoxiaojian: fixed

longbinlai (approved):
LGTM

PR #590, opened 2026-06-22
feat: add adaptive prefetch for CSR edge expansion

liulx20:
Introduce degree-distribution-based prefetch policy and pipeline prefetching in edge/path expand and triangle counting hot paths.

Fixes


lnfjpt (approved):
LGTM

PR #595, opened 2026-06-23
feat: support temporary graph computation via COPY TEMP

### LINKED ISSUE #526: [Feature] Support Temporary Graph Computation in AP via COPY TEMP

The AP pipeline in NeuG currently lacks session-level temporary data support. During graph analysis, users frequently need to temporarily load external data (CSV/Parquet/JSON) into the graph to perform associative queries with persistent graph data. Once the analysis is complete, the temporary data does not need to be persisted. Currently, users must use `CREATE TABLE` + `COPY FROM` to permanently write data into graph storage and then manually `DROP TABLE` to clean up — a cumbersome workflow that incurs unnecessary disk I/O and checkpoint overhead.

Add temporary graph capabilities on the AP read-write Connection via the `COPY TEMP` syntax, consisting of three functional modules:

1. **COPY TEMP Syntax**:
   ```sql
   -- Node table (simple)
   COPY TEMP TempUser FROM "users.csv" (primary_key='user_id', header=true);

   -- Node table (with filter/projection via LOAD FROM subquery)
   COPY TEMP TempAdults FROM (
       LOAD FROM "users.csv" (primary_key='user_id', header=true)
       WHERE age >= 18
       RETURN user_id, name
   );

   -- Relationship table (simple, first two columns are src/dst keys)
   COPY TEMP TempPurchased FROM "purchases.csv" (
       header=true,
       from='TempUser',
       to='Product'
   );

   -- Relationship table (with column reordering via subquery)
   COPY TEMP TempPurchased FROM (
       LOAD FROM "purchases.csv" (header=true)
       RETURN buyer_id, product_id, amount
   ) (from='TempUser', to='Product');
   ```
2. **Unified Cypher Query Interface**: Temporary and persistent types are queried through the same Cypher interface, supporting mixed MATCH (referencing both temporary and persistent labels simultaneously) with zero changes to the query path.
3. **Session-Level Lifecycle Management**: Temporary data is automatically released when the Connection closes and is never written to disk (`Dump`/`Checkpoint` skip temporary labels). `DROP TABLE <label>` is supported for explicit release of individual temporary types.

**Key Design Decisions**:

- `from`/`to` are graph-semantic options at the `COPY TEMP` statement level — they are NOT placed inside `LOAD FROM`. `LOAD FROM` only reads flat tables without graph semantics.
- `from_col`/`to_col` are not supported in `COPY TEMP`. Column position is the only mechanism — the first two columns are always src/dst keys. Use `RETURN` in a subquery to reorder columns when needed.
- `COPY TEMP` always infers schema automatically — no predefined table or `auto_detect` option is needed.

**Technical approach**: Shared storage with schema marking. Temporary data coexists with persistent data in the same PropertyGraph instance, distinguished by a `bool temporary` field on `VertexSchema`/`EdgeSchema`. No multi-graph merge layer is introduced; query operators (scan/filter/join/dml) require no modifications. WAL writes skip temporary types (no ACID guarantee for temporary data). Checkpoint (`Dump`/`Serialize`) also skips temporary types, ensuring temporary data is never persisted to disk.

**Tasks**

- [x] Schema Layer Temporary Marking + Serialization Separation
- [x] PropertyGraph Dump/Open/Compact
- [x] Param + Proto + Execution Layer Temporary Flag Propagation
- [x] Connection::Close() Temporary Data Cleanup
- [x] Support COPY TEMP Syntax in Parser/Binder/Planner (refactored from LOAD AS)
- [x] Support filter, project via LOAD FROM subquery in COPY TEMP
- [x] e2e tests (C++ test_copy_temp.cc + Python test_copy_temp.py)
- [x] Tests on different data sources: parquet/csv/json, local filesystem / remote filesystem
- [x] Error handling (read-only rejection, label conflict, nonexistent vertex label, dangling reference)
- [x] Documentation (import_data.md updated with COPY TEMP sections)

- **Separate PropertyGraph Instance**: Create an independent PropertyGraph for temporary data and merge results from both graphs at query time. This provides better isolation but requires implementing a multi-graph merge layer, introduces significant changes to the query path, and makes mixed queries (temporary edges connecting to persistent vertices) difficult to implement efficiently.

BingqingLyu:
## What does this PR do?

Implement temporary graph computation via `COPY TEMP` syntax. Loads external files (CSV, JSON, JSONL, Parquet) into temporary node/edge tables that participate in standard Cypher queries alongside persistent graph data. Temporary tables exist only for the lifetime of the Connection and are automatically cleaned up when the Connection is closed.

Closes #526

## Syntax

```sql
-- Node table
COPY TEMP TempUser FROM "users.csv" (primary_key='user_id', header=true);

-- Node table with filter/projection
COPY TEMP TempAdults FROM (
    LOAD FROM "users.csv" (primary_key='user_id', header=true)
    WHERE age >= 18
    RETURN user_id, name
);

-- Relationship table (from/to are graph-semantic options at COPY TEMP level)
COPY TEMP TempKnows FROM "edges.csv" (header=true, from='TempUser', to='TempUser');

-- Relationship table with column reordering via subquery
COPY TEMP TempKnows FROM (
    LOAD FROM "edges.csv" (header=true)
    RETURN src, dst, weight
) (from='TempUser', to='TempUser');
```

### Parser & Binder

- ANTLR4 Cypher grammar: `COPY TEMP <label> FROM ...` syntax rules (refactored from former `LOAD AS`)
- Binder: routes `COPY TEMP` through `bindCopyTempNodeFrom` / `bindCopyTempRelFrom` depending on presence of `from`/`to` options
- Subquery path: `COPY TEMP ... FROM (LOAD FROM ... WHERE ... RETURN ...)` for filter/projection pushdown

### Query Compilation & Optimization

- Column position is the only key mechanism — first column is primary key for nodes; first two columns are src/dst for relationships
- `RETURN` in subquery enables column reordering when key columns are not at expected positions
- `ProjectIntoDataSourceOptimizer`: preserves projection nodes under `COPY_FROM` to ensure WHERE-only columns are correctly trimmed

### Storage Layer

- Schema temporary label support: `VertexSchema`/`EdgeSchema` with `temporary` flag
- `Schema::StripTemporary()`: creates a clean Schema copy without temporary labels for checkpoint persistence; uses `vector<bool>` + sequential iteration for deterministic label ID ordering
- `DumpToYaml` always dumps all labels (including temporary) for runtime compiler/planner use; persistence paths (`ToJson`, `PropertyGraph::Dump`) call `StripTemporary()` first
- `PropertyGraph::Dump` uses `StripTemporary()` directly instead of YAML round-trip

### Lifecycle Management

- Automatic cleanup of all temporary tables when Connection closes
- `DROP TABLE` support for manual early release of temporary tables
- `COPY TEMP` rejected in read-only mode with explicit error message

### Documentation & Tests

- `doc/source/data_io/import_data.md`: integrated `COPY TEMP` content into existing COPY FROM sections
- `specs/004-temp-graph/`: feature spec, design proposal, task breakdown
- C++ unit tests: `test_temporary_graph.cc` + `test_copy_temp.cc` (25 test cases)
- Python E2E tests: `test_copy_temp.py` (48 test cases)
- Coverage: basic load, WHERE/RETURN pushdown, key column reorder, mixed queries, lifecycle management, read-only rejection, remote HTTPFS, multi-format support (CSV, JSON, JSONL, Parquet)


doc/source/data_io/load_as.md
```diff
@@ -0,0 +1,417 @@
+# Load As (Temporary Graph)
+
+`LOAD NODE TABLE AS` / `LOAD REL TABLE AS` reads data from external files (CSV, JSON, JSONL, Parquet, etc.) and creates **temporary node/relationship tables** that participate in standard Cypher queries alongside persistent graph data. Temporary tables exist only for the lifetime of the Connection and are automatically cleaned up when the Connection is closed.
```
longbinlai: The doc of `load as` just be integrated into `load from`, it is just one extra `AS` clause. Explain when you need the as, and how to use is fine. Does not need one extra document page.

doc/source/data_io/import_data.md
```diff
@@ -70,6 +79,27 @@ RETURN u1.name, u2.name, f.since_year
...
 
+### Temporary Graph
+
+To import the same data as a **temporary** graph (no DDL, no persistence):
+
+-- Temporary node table (auto-inferred schema, first column = primary key)
+COPY TEMP TempUser FROM "users.csv" (header=true, delimiter=",");
+
+-- Temporary relationship table (from/to specify endpoint labels)
+COPY TEMP TempFriends FROM "friendships.csv" (
```
longbinlai: edge name should be `TEMP_FRIENDS`.  Fix all edge names as convention

doc/source/data_io/import_data.md
```diff
@@ -70,6 +79,27 @@ RETURN u1.name, u2.name, f.since_year
 LIMIT 5;
 ```
 
+### Temporary Graph
+
+To import the same data as a **temporary** graph (no DDL, no persistence):
+
+-- Temporary node table (auto-inferred schema, first column = primary key)
+COPY TEMP TempUser FROM "users.csv" (header=true, delimiter=",");
```
longbinlai, line 88: I think for keyword `COPY TEMPORARY ...` is better than  `COPY TEMP`, TEMP has the meaning of temporal.

doc/source/data_io/import_data.md
```diff
@@ -106,13 +149,32 @@ COPY KNOWS FROM "person_knows_person.csv" (from="Person", to="Person", header=tr
...
+    from='Person',
+    to='Person'
+);
+
+-- With column reordering (when keys are NOT at positions [0/1]):
+COPY TEMP TempKnows FROM (
+    LOAD FROM "edges_shuffled.csv" (header=true)
+    RETURN src_id, dst_id, weight
+) (from='Person', to='Person');
+```
```
zhanglei1949, line 172: Maybe we should mention how we name the properties if the csv is header-less?

include/neug/main/query_processor.h
```diff
@@ -66,6 +66,8 @@ class QueryProcessor {
                               const rapidjson::Value& parameters_json,
                               int32_t num_threads = 0);
 
+  void clear_cache();
```
zhanglei1949, line 69: `clear_cache()` overlaps with `update_compiler_meta_if_needed`, consider reusing `update_compiler_meta_if_needed`

src/storages/graph/schema.cc
```diff
@@ -1630,10 +1670,14 @@ static Status parse_schema_config_file(const std::string& path,
 bool dump_vertices_schema(const Schema& schema, YAML::Node& node) {
   auto v_labels = schema.get_vertex_label_ids();
   for (auto& v_label : v_labels) {
+    bool is_temp = schema.is_vertex_label_temporary(v_label);
     YAML::Node cur_node(YAML::NodeType::Map);
     cur_node["type_name"] = schema.get_vertex_label_name(v_label);
     cur_node["description"] = schema.get_vertex_description(v_label);
     cur_node["type_id"] = std::to_string(v_label);
+    if (is_temp) {
```
zhanglei1949: We already have `StripTemporary`, so do we still need ` cur_node["temporary"] = true;`?

tools/python_bind/tests/test_copy_temp.py
```diff
@@ -0,0 +1,829 @@
...
+
+
+def _write_csv(directory, filename, content):
+    path = os.path.join(directory, filename)
+    with open(path, "w") as f:
+        f.write(content)
+    return path
+
+
+class TestCopyTemp:
```
zhanglei1949, line 35: Need a test which submit temp queries to Session, and expected to be rejected.

doc/source/data_io/import_data.md
BingqingLyu: `TEMPORARY` is indeed more explicit. However, I think `TEMP` is a common abbreviation in SQL ecosystems (e.g., Databricks uses `CREATE TEMP TABLE` syntax), and it keeps the syntax concise?

doc/source/data_io/import_data.md
BingqingLyu: Done

doc/source/data_io/import_data.md
BingqingLyu: Done.

include/neug/main/query_processor.h
BingqingLyu: Done.

tools/python_bind/tests/test_copy_temp.py
BingqingLyu: Done.

doc/source/data_io/import_data.md
longbinlai: okay

doc/source/data_io/import_data.md
```diff
@@ -166,7 +166,9 @@ COPY TEMP TEMP_KNOWS FROM (
 ) (from='Person', to='Person');
 ```
 
-> **Note:** `from`/`to` are graph-semantic options at the `COPY TEMP` level — they are not placed inside `LOAD FROM`. 
+> **Note:** `from`/`to` are graph-semantic options at the `COPY TEMP` level — they are not placed inside `LOAD FROM`.
+
+> **Tip:** If the CSV has no header row, property names default to the auto-generated column names from the file reader (e.g., `f0`, `f1`, `f2`, ...).
```
longbinlai: Should put this somewhere. It works for every load/copy/copy temp, not just copy temp

doc/source/data_io/import_data.md
```diff
@@ -166,7 +166,9 @@ COPY TEMP TEMP_KNOWS FROM (
 ) (from='Person', to='Person');
 ```
 
-> **Note:** `from`/`to` are graph-semantic options at the `COPY TEMP` level — they are not placed inside `LOAD FROM`. 
+> **Note:** `from`/`to` are graph-semantic options at the `COPY TEMP` level — they are not placed inside `LOAD FROM`.
```
longbinlai: I have no idea what this note mean

tools/python_bind/tests/test_copy_temp.py
```diff
@@ -827,3 +828,34 @@ def test_copy_temp_rel_rejected(self):
...
+        f.write("id|name|age\n1|Alice|30\n2|Bob|25\n")
+
+    db = Database(db_path=db_dir, mode="w")
+    uri = db.serve(10086, "localhost", False)
+    time.sleep(1)
+
+    try:
+        session = Session(uri, timeout="10s")
+        with pytest.raises(Exception):
+            session.execute(
```
longbinlai, line 835: Should verify the exception, to see if it matches the design

src/storages/graph/schema.cc
BingqingLyu: Agreed. Removed the temporary field from YAML serialization. The temporary flag is maintained in the in-memory VertexSchema/EdgeSchema and is never parsed back from YAML. Persistence safety is guaranteed by StripTemporary() which excludes temporary labels entirely before Dump.

doc/source/data_io/import_data.md
BingqingLyu: I removed this note.

tools/python_bind/tests/test_copy_temp.py
BingqingLyu: Done.

doc/source/data_io/import_data.md
BingqingLyu: Done.

doc/source/data_io/import_data.md
```diff
@@ -2,12 +2,16 @@
...
 
+A variant — **`COPY TEMP`** — imports external data as a **temporary graph** whose lifetime is bound to the current connection. Temporary tables are automatically removed when the connection closes, making `COPY TEMP` ideal for ad-hoc analytics without polluting the persistent schema.
+
 ## Schema Requirement
 
 You can create a **predefined schema** — that is, define node/relationship tables before importing data — where the columns in the external file must match the table properties.
 
 Since v0.1.2, NeuG supports schema-flexible persistent import — allowing `COPY FROM` to leverage the capability of type inference of `LOAD FROM`, without requiring a predefined schema. This will make it much easier to quickly onboard new datasets. See [Import without a predefined schema](#import-without-a-predefined-schema) for more usages.
 
+> **COPY TEMP** always infers the schema automatically — no predefined table or `auto_detect` option is needed. The first column becomes the primary key for nodes; for relationships, the first two columns are source/destination keys.
```
longbinlai: no predefined table or `auto_detect` option is needed，这句话有歧义，实际是 他不能提前定义

src/compiler/binder/bind/copy/bind_copy_from.cpp
```diff
@@ -193,6 +195,18 @@ static bool autoDetectEnabled(
 std::unique_ptr<BoundStatement> Binder::bindCopyFromClause(
     const Statement& statement) {
   auto& copyStatement = neug_dynamic_cast<const CopyFrom&>(statement);
+
+  // --- COPY TEMP: always infer schema, create temporary table ---
+  if (copyStatement.isTemporary()) {
```
longbinlai, line 200: **[High] Security: COPY TEMP file path has no validation**

`COPY TEMP` allows loading from arbitrary file paths with no path validation or sandboxing. In Service mode, a malicious user could read arbitrary files via:
```sql
COPY TEMP T FROM "/etc/passwd" (header=true);
```

**Suggestion:**
- In service mode, restrict file paths to an allowed data directory (similar to the existing `FLEX_DATA_DIR` mechanism)
- Or provide a configurable `allowed_temp_file_dirs` whitelist
- At minimum, reject paths containing `..` or absolute paths outside the data dir when running as a service

src/compiler/binder/bind/copy/bind_copy_from.cpp
```diff
@@ -467,6 +481,96 @@ std::unique_ptr<BoundStatement> Binder::bindCopyRelFromNoSchema(
   return std::make_unique<BoundCopyFrom>(std::move(boundCopyFromInfo));
 }
 
+std::unique_ptr<BoundStatement> Binder::bindCopyTempNodeFrom(
```
longbinlai: **[Medium] Significant code duplication with bindCopyNodeFromNoSchema / bindCopyRelFromNoSchema**

`bindCopyTempNodeFrom` (lines 484-509) is nearly identical to `bindCopyNodeFromNoSchema` (lines 391-416), and `bindCopyTempRelFrom` (lines 511-572) is nearly identical to `bindCopyRelFromNoSchema` (lines 418-482). The only difference is `temporary=true` passed to `DDLVertexInfo` / `DDLEdgeInfo`.

~90 lines of duplicated code that must be kept in sync.

**Suggestion:** Add a `bool temporary = false` parameter to the existing `*NoSchema` functions and call them directly:
```cpp
std::unique_ptr<BoundStatement> Binder::bindCopyNodeFromNoSchema(
    const Statement& statement,
    const case_insensitive_map_t<Value>& boundCopyOptions,
    bool temporary = false);
```
Then in the `isTemporary()` branch:
```cpp
return bindCopyNodeFromNoSchema(statement, boundOpts, /*temporary=*/true);
```

src/compiler/binder/bind/copy/bind_copy_from.cpp
longbinlai: **Correction:** My original comment about service mode is incorrect — `index.md` clearly documents that `LOAD FROM`, `COPY FROM`, and `COPY TO` are embedded-mode only, and `COPY TEMP` follows the same restriction. So the service-mode attack vector does not apply.

The path traversal concern is downgraded to **low** — in embedded mode, the application is typically trusted. However, if an embedded application accepts untrusted Cypher input (e.g., a notebook-style UI), unrestricted file paths could still be a concern. This is an application-level responsibility, not a blocker for this PR.

Apologies for the noise.

doc/source/data_io/import_data.md
```diff
@@ -70,6 +79,27 @@ RETURN u1.name, u2.name, f.since_year
 LIMIT 5;
 ```
 
+### Temporary Graph
+
+To import the same data as a **temporary** graph (no DDL, no persistence):
+
+-- Temporary node table (auto-inferred schema, first column = primary key)
```
longbinlai: **[Doc] Temporary Graph example code is not in a code block**

Lines 86-101 are plain text, not wrapped in ` ```cypher ... ``` `. This breaks markdown rendering — the code will appear as regular paragraphs. Compare with Step 2/3 above which use proper code blocks.

Also, the SQL-style comments (`--`) are inconsistent with the rest of the doc which uses Cypher comments (`//`).

doc/source/data_io/import_data.md
```diff
@@ -92,6 +122,19 @@ COPY Person FROM "person*.csv" (header=true);
 
 > **Note:** The number and order of columns in the CSV file must match the properties defined in the node table exactly.
 
+**Temporary node table** — no DDL needed, schema is auto-inferred:
+
+```cypher
+COPY TEMP TempPerson FROM "person.csv" (primary_key='id', header=true);
```
longbinlai: **[Doc] `primary_key='id'` option in COPY TEMP is confusing**

The PR design states: "Column position is the only key mechanism — first column is primary key for nodes." The doc also says (line 13): "The first column becomes the primary key for nodes."

So `primary_key='id'` at the COPY TEMP level is either:
- **Redundant** — if `id` is already the first column
- **Misleading** — if `id` is NOT the first column, the option may be ignored by the binder (which always uses `columns[0]`)

**Suggestion:** Either remove `primary_key='id'` from the example (since it's unnecessary for COPY TEMP), or add a note explaining that COPY TEMP always uses the first column as PK regardless of the `primary_key` option. Users who need a different PK column should use the `LOAD FROM` subquery with `RETURN` to reorder columns.

doc/source/data_io/import_data.md
```diff
@@ -70,6 +79,27 @@ RETURN u1.name, u2.name, f.since_year
 LIMIT 5;
 ```
 
+### Temporary Graph
```
longbinlai, line 82: **[Doc] Missing `DROP TABLE` documentation for temporary tables**

The PR body mentions "DROP TABLE support for manual early release of temporary tables," but the docs don't cover this. Users have no way to know they can manually drop temporary tables before the connection closes.

**Suggestion:** Add a brief note under "Temporary Graph" explaining:
```cypher
// Manually drop a temporary table before connection close
DROP TABLE TempUser;
```
And note that all remaining temporary tables are automatically cleaned up when the connection closes.

doc/source/data_io/import_data.md
```diff
@@ -106,13 +149,30 @@ COPY KNOWS FROM "person_knows_person.csv" (from="Person", to="Person", header=tr
 
 > **Note:** NeuG assumes the first two columns are the primary keys of the `FROM` and `TO` nodes. The remaining columns correspond to relationship properties. The `from` and `to` parameters must be specified to identify the endpoint node tables.
 
+**Temporary relationship table:**
+
+```cypher
+-- Simple: first two columns are src/dst keys
+COPY TEMP TEMP_KNOWS FROM "edges.csv" (
+    header=true,
+    from='Person',
```
longbinlai, line 163: **[Doc] `from`/`to` in COPY TEMP should clarify they must reference existing vertex labels**

The regular COPY FROM section (line 217) states: "Both tables must already exist (create or `COPY` them first)." But the COPY TEMP relationship example doesn't mention this. A user might try `from='TempUser'` before creating `TempUser`, or use a label name that doesn't exist.

**Suggestion:** Add a note that `from`/`to` must reference existing vertex labels (either persistent or previously created temporary labels).

doc/source/data_io/index.md
```diff
@@ -47,5 +47,5 @@ Bulk file I/O operations (`LOAD FROM`, `COPY FROM`, `COPY TO`) involve reading o
 ## What's Next
 
 - **[LOAD FROM](load_data)** — Read external files into temporary tables with relational operations
-- **[COPY FROM](import_data)** — Persist external data into graph storage
+- **[COPY FROM / COPY TEMP](import_data)** — Persist or temporarily import external data into graph storage
```
longbinlai: **[Doc] Two documentation issues in index.md**

1. **"Embedded Mode Only" section omits COPY TEMP**: The section above lists `LOAD FROM`, `COPY FROM`, and `COPY TO` but not `COPY TEMP`. Since `COPY TEMP` is built on `LOAD FROM`, it follows the same restriction. Users might assume `COPY TEMP` works in service mode. Suggest updating to: "`LOAD FROM`, `COPY FROM`, `COPY TEMP`, and `COPY TO` are supported **only in embedded mode**."

2. **Architecture diagram is outdated**: The diagram (line 19-20) still says `COPY FROM (requires predefined schema)`, but `auto_detect=true` has been the default since v0.1.2. Consider updating to reflect auto-detection and COPY TEMP.

src/compiler/binder/bind/copy/bind_copy_from.cpp
BingqingLyu: Currently, `COPY TEMP` is rejected in Service (TP) mode via validate_flags() in neug_db_session.cc.

longbinlai:
## Test Report: Comprehensive Testing of COPY TEMP (PR #595)

I performed a thorough review and testing of this PR. Here are the results:

### Test Execution Summary

| Suite | Result |
|-------|--------|
| C++ `test_copy_temp` (25 cases) | ✅ All passed |
| C++ `test_temporary_graph` (24 cases) | ✅ All passed |
| Python `test_copy_temp.py` (42 cases) | ✅ All passed |
| Python `test_gds_temp_graph.py` (10 cases) | ✅ All passed |
| Additional edge-case tests (25 cases) | ✅ All passed |
| Additional security tests (19 cases) | ✅ All passed |
| Parquet/HTTPFS tests (6 cases) | ⚠️ Skipped (extensions not built in test env) |

### Coverage of Additional Tests
- Non-existent file paths, malformed CSV, empty files
- Label name collisions (temp vs persistent, duplicate temp)
- Self-loop edges, dangling edge references
- Multiple temp tables + cache invalidation
- DROP + recreate same label name
- Temp data not persisted after reopen (verified on disk)
- Mixed persistent + temp JOIN queries
- Unicode values, special characters, large column counts
- WHERE with OR/arithmetic, RETURN-only-PK
- Float/mixed type inference, negative/large integers
- File paths with spaces, relative paths
- Concurrent connection rejection
- Label ID remapping with interleaved temp/persistent labels
- Aggregation (SUM/MIN/MAX/COUNT) on temp tables

---

### 🐛 Bug Found: `primary_key` option is silently ignored

**Severity: Medium (silent data loss)**

The `primary_key` option in COPY TEMP is accepted by the parser but has no effect. The first CSV column is always used as the primary key, regardless of the `primary_key` value specified.

**Root cause:** In `bind_copy_from.cpp`, `bindCopyNodeFromNoSchema()` explicitly voids the bound options:

```cpp
std::unique_ptr<BoundStatement> Binder::bindCopyNodeFromNoSchema(
    const Statement& statement,
    const case_insensitive_map_t<Value>& boundCopyOptions, bool temporary) {
  (void) boundCopyOptions;  // <-- primary_key is ignored here
  ...
  const auto& primaryKey = columns[0]->rawName();  // Always first column
```

**Reproduction:**

```sql
-- CSV: name|id|age (id is the SECOND column)
COPY TEMP TempUser FROM "users.csv" (primary_key='id', header=true);

-- Edge CSV: src|dst with id values (1, 2, 3)
COPY TEMP TempKnows FROM "edges.csv" (header=true, from='TempUser', to='TempUser');

-- Result: 0 edges created! Edge keys (1, 2, 3) are looked up against
-- the 'name' column (first column) instead of 'id' (as specified).
-- No error is reported.
```

**Impact:**
- Edges referencing vertices by the specified primary key column are **silently dropped**
- The PR documentation shows `primary_key='user_id'` as a supported option, but it doesn't actually work
- All existing tests use `primary_key='id'` where 'id' IS the first column, so this bug is not exposed

**Suggested fix options:**
1. **Honor the `primary_key` option**: Use it to identify which column is the primary key (reorder columns or use the specified column for index lookup)
2. **Reject with error**: If `primary_key` is specified for COPY TEMP, return an error: "primary_key option is not supported with COPY TEMP; the first column is always used as the primary key"
3. **At minimum, add a warning** when `primary_key` doesn't match the first column name

---

### ⚠️ UX Issue: `from`/`to`-only silently creates a node table

When only `from` or `to` is specified (but not both), COPY TEMP silently creates a **node table** and ignores the `from`/`to` option. No error or warning is given.

```sql
-- User intends to create an edge table but forgot 'to'
COPY TEMP TempEdge FROM "edges.csv" (header=true, from='TempUser');
-- Result: A node table is created. The 'from' option is silently ignored.
```

**Root cause:** The binder checks `if (fromIt != end && toIt != end)` — if either is missing, it falls through to node table creation.

**Suggested fix:** When `from` or `to` is specified without the other, return an error: "Both 'from' and 'to' options are required to create a relationship table."

---

### ✅ What Works Well

- **Persistence protection**: Temporary labels are correctly stripped from checkpoints via `StripTemporary()`, binary serialization, and `Dump()` filtering. Verified that temp data does not survive database reopen.
- **Label ID remapping**: When temp labels are interspersed with persistent labels, the `StripTemporary()` function correctly remaps label IDs and persistent data survives close/reopen.
- **Lifecycle management**: `Connection::Close()` properly cleans up temp edges before temp vertices. Double-close is idempotent.
- **Read-only rejection**: COPY TEMP is correctly rejected in read-only mode with a clear error message.
- **Service mode rejection**: COPY TEMP is correctly rejected in TP service mode.
- **Concurrent access**: Second read-write connection is properly rejected with `TX_STATE_CONFLICT`.
- **WHERE/RETURN pushdown**: Filter and projection work correctly across CSV format, including filter-only columns not in RETURN.
- **Error handling on failure**: If COPY TEMP fails (e.g., non-existent file), no temp label remains in the schema.

longbinlai (dismissed):
LGTM

longbinlai (approved):
LGTM

PR #605, opened 2026-06-23
fix: pin manylinux tag

### LINKED ISSUE #601: [BUG] Pin manylinux base image version for neug-manylinux
Since the latest manylinux image dropped Python 3.8 support, use a pinned manylinux version to build the neug-manylinux image and resolve the compatibility issue.

lnfjpt:
Fixes https://github.com/alibaba/neug/issues/601


zhanglei1949 (approved):
LGTM

PR #606, opened 2026-06-23
fix: Fix multi-way intersect edge aliases

### LINKED ISSUE #564: [BUG] INTERSECT operator rejects N≥3 sub-plans that need to keep edge aliases, blocking K_4-like patterns (e.g. LSQB-glogs/p6 on LDBC)
**Describe the bug**

When the optimizer chooses an N-way (N ≥ 3) `INTERSECT` plan **and** any sub-plan needs to keep its edge alias (because a later operator references the edge variable), `IntersectOprBuilder::Build` rejects the plan with:

```
Not supported: Keeping edge aliases is only supported for binary intersect.
  at src/execution/execute/ops/retrieve/intersect.cc:260
  func: Build, Error Code: 1011
```

This is a structural execution-engine gap: the binary INTERSECT path (`IntersectWithEdgeOpr`) supports keeping edge aliases, but the multi-way path (`IntersectOprMultip`) does not, and the builder hard-throws when both conditions hold.

The optimizer reliably picks such plans for any K_4-like 6-edge pattern that has edges converging on more than two vertices. On LDBC SF1, this blocks **all 10 variants of LSQB-glogs/p6** under three independent cardinality-input modes (baseline neug estimator, identity-override, and external estimator override) — confirming it is unrelated to cardinality estimation quality.

**To Reproduce**
```
echo "MATCH (_n0:FORUM)-[e1:HASMEMBER]->(_n1:PERSON),
            (_n0)-[e2:HASMEMBER]->(v3:PERSON),
            (_n0)-[e3:CONTAINEROF]->(v4:POST),
            (_n1)-[e4:KNOWS]->(v3),
            (_n1)-[e5:LIKES]->(v4),
            (v3)-[e6:LIKES]->(v4)
      WHERE v3.birthday >= 315532800000
        AND v4.explicitlydeleted = false
      RETURN count(*);" | ldbc_query ~/ldbc_sf1_db -
```

**Expected behavior**

**Screenshots**

 - NeuG version: [e.g., v0.1, v0.2.2, master]
 - OS: [e.g. macOS, Linux]
 - Version [e.g. 10.15]
 - Python Version [e.g. 3.12]

**Additional context**

shirly121:
## Related issue number

Fixes #564


src/execution/common/operators/retrieve/intersect.cc
```diff
@@ -423,6 +423,153 @@ neug::result<ContextChunk> Intersect::Multiple_Intersect(
...
+            EdgeValues values(eeps.size());
+            values[eep_idx] = edge_value;
+            target[nbr].emplace_back(std::move(values));
+            return;
+          }
+          auto iter = previous->find(nbr);
+          if (iter == previous->end()) {
+            return;
+          }
+          for (auto values : iter->second) {
```
longbinlai, line 484: If a neighbor `nbr` has N matching edges in expand 0 and M in expand 1, this produces N×M `EdgeValues` combinations. For N-way intersect on high-degree vertices this becomes N×M×K... — potentially exponential. Consider adding a limit or documenting this as a known limitation for dense multi-edge graphs.

tools/python_bind/tests/test_db_query.py
```diff
@@ -3202,3 +3202,23 @@ def test_delete_all_rows_then_reinsert_visible(tmp_path):
 
     conn.close()
     db.close()
+
+
+def test_lsqb_6_test(tmp_path):
+    db_dir = tmp_path / "/tmp/lsqb"
```
longbinlai: `tmp_path / "/tmp/lsqb"` doesn't produce the intended path — should be `"/tmp/lsqb"` directly, or better, use an environment variable.

Also, the test requires pre-loaded LSQB data but has no skip mechanism — it will fail in CI when data is missing. Suggest adding:
```python
@pytest.mark.skipif(not os.path.isdir("/tmp/lsqb"), reason="LSQB data not found")
```

src/execution/execute/ops/retrieve/intersect.cc
```diff
@@ -105,34 +110,33 @@ class IntersectWithEdgeOpr : public IOperator {
...
-    EdgeAndNbrPredicate left_pred(std::move(left_v_pred),
-                                  std::move(left_e_pred));
-    EdgeAndNbrPredicate right_pred(std::move(right_v_pred),
-                                   std::move(right_e_pred));
     return ctx.apply_chunks(
         [&](ContextChunk&& chunk) -> neug::result<ContextChunk> {
-          return Intersect::Binary_Intersect_With_Edge(
-              graph, params, std::move(chunk), std::move(left_pred),
-              std::move(right_pred), eep0_, eep1_, v_alias_, edge_alias_);
+          std::vector<EdgeAndNbrPredicate> preds;
```
longbinlai: Predicates are now re-bound inside the `apply_chunks` lambda for every chunk, instead of once before it. This is consistent with `IntersectOprMultip` above, but if there are many chunks the repeated `bind()` calls add overhead. Worth confirming `bind()` is lightweight enough for this to be negligible.

longbinlai (approved):
LGTM

Issue #592, opened 2026-06-23
bvar sampler warns 'busy at sampling for 2 seconds' under long-running service


longbinlai:
## Problem

When running `neug` in service mode overnight, the following warnings appear repeatedly in the log:

```
W20260622 20:06:16.355587 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
W20260622 20:08:41.038022 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
W20260622 20:09:01.197245 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
W20260622 20:09:06.730468 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
W20260622 20:10:33.863665 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
W20260622 20:13:22.903072 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
W20260622 20:19:45.149266 27121391 sampler.cpp:192] bvar is busy at sampling for 2 seconds!
```

These warnings come from brpc's bvar `sampler.cpp` (line 192), indicating the background sampling thread is unable to complete its sampling cycle within the expected 2-second window.

## Environment

- NeuG service mode (HTTP server), running overnight (~hours)
- Platform: macOS (darwin, arm64)

## Impact

- The warnings indicate potential performance degradation or resource contention in long-running service mode
- bvar metrics may be inaccurate or delayed during these periods
- Could be a symptom of thread starvation, excessive bvar registration, or I/O contention

/cc @zhanglei1949 — could you take a look? This appears to be related to the brpc bvar sampling mechanism.

zhanglei1949:
Update from local reproduction:

I was able to reproduce the warning more deterministically by running the NeuG HTTP service stress test for about 100 seconds while compiling the project in another shell on the same machine. During the overlapping build workload, brpc emitted repeated warnings like:

```text
sampler.cpp:192] bvar is busy at sampling for 2 seconds!
```

This makes the issue credible, but the trigger looks like CPU/scheduler pressure rather than a NeuG query correctness bug. brpc prints this warning when the bvar sampler cannot keep its normal 1Hz sampling cadence for consecutive rounds. Under heavy co-located workload, such as service pressure plus a concurrent compile, this behavior is expected.

Recommended handling for NeuG:

[...]

PR #611, opened 2026-06-24
refactor: Refine default parallelism configuration for service

zhanglei1949:
Related issue: #592


tools/python_bind/src/py_database.h
```diff
@@ -68,8 +68,8 @@ class PyDatabase : public std::enable_shared_from_this<PyDatabase> {
    * @brief Start the database server.
    * @param port The port to listen on, default is 10000.
    * @param host The host to bind to, default is "localhost".
-   * @param num_thread The number of threads to use, default is 0, which means
-   * use all hardware threads.
+   * @param num_thread The number of brpc worker threads to use. 0 means
```
liulx20, line 71: There is no need to emphasize `brpc`, we can support compatibility with different protocols in the future.

src/server/brpc_service_mgr.cc
```diff
@@ -327,10 +332,21 @@ void BrpcServiceManager::Stop() {
   LOG(INFO) << "Brpc server stopped";
 }
 
+uint32_t BrpcServiceManager::resolve_num_threads() const {
+  if (service_config_.shard_num != 0) {
```
liulx20, line 336: What is the relationship between these variables and why is there an inconsistency?

tools/python_bind/src/py_database.h
liulx20, line 38: A `thread_num` is passed in the constructor, and another one is passed when starting the server. What is the relationship between these two?

tools/python_bind/README.md
```diff
@@ -64,6 +64,29 @@ fallback if absent). The wheel bundles both `neug_py_bind*.so` and
 `libneug.{dylib,so}` into the `neug` package; they find each other at runtime
 via `@loader_path` (macOS) / `$ORIGIN` (Linux) RPATH.
 
+## Python service API
+
+`Database.serve()` starts the database in service mode and can configure the
+server-side brpc worker thread count:
```
liulx20, line 70: Similarly, there is no need to emphasize brpc here

longbinlai (approved):
LGTM

src/server/brpc_service_mgr.cc
zhanglei1949: `shared_num` is the legacy naming, we shall all migrate to `thread_num`. 
Replace the usage of `SessionNum` with `db_config_.max_thread_num`, actually they are equal.

tools/python_bind/README.md
zhanglei1949: Removed.

tools/python_bind/src/py_database.h
zhanglei1949: `Database` is constructed with a maximum thread num.
`db.serve()` could be provided with a specific `thread_num` argument to control the number of threads used for serving. If not specified, the default `max_thread_num` will be used.

tools/python_bind/src/py_database.h
zhanglei1949: Removed

PR #614, opened 2026-06-24
fix: force vendored rapidjson for Arrow JSON build

### LINKED ISSUE #613: [BUG] Arrow build picks up system /usr/include rapidjson instead of third_party, breaking compilation when rapidjson-dev is installed
## Summary

When the system package `rapidjson-dev` is installed, building Arrow (with `ARROW_JSON` enabled) fails to compile because Arrow ends up including the system header `/usr/include/rapidjson/document.h` instead of the vendored `third_party/rapidjson`, despite `RapidJSON_ROOT` being set to the latter.

## Error
```bash
usr/include/rapidjson/document.h: In member function 'arrow::rapidjson::GenericStringRef& ...::operator=(...)': /usr/include/rapidjson/document.h:319:82: error: assignment of read-only member 'arrow::rapidjson::GenericStringRef::length' 319 | GenericStringRef& operator=(const GenericStringRef& rhs) { s = rhs.s; length = rhs.length; }
```
The system rapidjson (the 1.1.0 release tarball, e.g. Ubuntu `rapidjson-dev`) ships a broken `GenericStringRef::operator=` that assigns to `const` members and has no return statement. It fails under our GCC/strict standard. The vendored `third_party/rapidjson` has the fixed version (`operator=` declared `= delete`).

liulx20:
Arrow's FindRapidJSONAlt runs config-mode find_package(RapidJSON) before honoring RapidJSON_ROOT. When the system rapidjson-dev package is installed, its /usr/lib/cmake/RapidJSON config wins and points includes at /usr/include (the broken 1.1.0 release whose GenericStringRef::operator= fails to compile), silently ignoring the third_party/rapidjson hint.

Set CMAKE_DISABLE_FIND_PACKAGE_RapidJSON=ON so config-mode discovery is skipped and Arrow falls back to RapidJSON_ROOT (third_party/rapidjson).

Fix #613

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -593,6 +593,12 @@ set(ARROW_JSON ON CACHE BOOL "" FORCE)
 set(RapidJSON_SOURCE "SYSTEM" CACHE STRING "" FORCE)
 # Point Arrow to use the project's RapidJSON
 set(RapidJSON_ROOT "${CMAKE_SOURCE_DIR}/third_party/rapidjson" CACHE PATH "" FORCE)
+# Disable config-mode find_package(RapidJSON) so Arrow's FindRapidJSONAlt falls
+# back to RapidJSON_ROOT above. Otherwise it picks up the apt-installed
+# /usr/lib/cmake/RapidJSON config, whose include dir is /usr/include (rapidjson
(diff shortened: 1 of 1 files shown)


zhanglei1949 (approved):
LGTM

PR #616, opened 2026-06-24
fix: avoid GCC dangling-pointer false positive in expand_degree 

### LINKED ISSUE #615: [BUG] EdgeExpand::expand_degree fails to build on GCC 13+ with -Werror=dangling-pointer
### Summary
When building with GCC 13+ (libstdc++ 14 headers) and -Werror enabled, compilation of edge_expand.cc fails in expand_degree with a -Wdangling-pointer error. This appears to be a false positive: a lambda-local variable is passed by const T& into vector::push_back via mimalloc’s STL allocator, and the compiler incorrectly reports a dangling pointer even though the reference is valid for the full duration of the call.

### Environment
OS: Linux (Ubuntu 20.04)
Compiler: GCC 13+ with libstdc++ 14 headers (/usr/include/c++/14/)
Build flags: -Werror (see root CMakeLists.txt)
Allocator: mimalloc (mi_stl_allocator)
Affected file: src/execution/common/operators/retrieve/edge_expand.cc

Error
```
In member function ‘void mi_stl_allocator<T>::construct(U*, Args&& ...) ...’
    inlined from ‘std::vector<_Tp, _Alloc>::push_back(const value_type&) ...’
    inlined from ‘ValueColumnBuilder<T>::push_back_opt(const T&) ...’
    inlined from ‘EdgeExpand::expand_degree(...)::<lambda(...)>’
    at edge_expand.cc:86:26:
mimalloc.h:437:76: error: dangling pointer to ‘degree’ may be used [-Werror=dangling-pointer=]
```
Relevant code (before fix):

```cpp
foreach_vertex(*vertex_col, [&](size_t index, label_t label, vid_t v) {
  int64_t degree = 0;
  // ...
  builder.push_back_opt(degree);
  shuffle_offset.push_back(index);
});
```

liulx20:
Fix #615

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -593,6 +593,12 @@ set(ARROW_JSON ON CACHE BOOL "" FORCE)
 set(RapidJSON_SOURCE "SYSTEM" CACHE STRING "" FORCE)
 # Point Arrow to use the project's RapidJSON
 set(RapidJSON_ROOT "${CMAKE_SOURCE_DIR}/third_party/rapidjson" CACHE PATH "" FORCE)
+# Disable config-mode find_package(RapidJSON) so Arrow's FindRapidJSONAlt falls
+# back to RapidJSON_ROOT above. Otherwise it picks up the apt-installed
+# /usr/lib/cmake/RapidJSON config, whose include dir is /usr/include (rapidjson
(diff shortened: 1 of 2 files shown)


zhanglei1949 (approved):
LGTM

PR #621, opened 2026-06-25
refactor: Introduce gc for checkpoint

### LINKED ISSUE #531: enhancement: Add checkpoint garbage collection to CheckpointManager
## Summary

`CheckpointManager` has no mechanism to delete old checkpoints. Every `CreateCheckpoint()` call adds a new `checkpoint-NNNNN` directory that persists forever. Over time this causes unbounded disk usage growth.

## Current Behavior

- `CreateCheckpoint()` always creates a new directory with an incremented ID
- No method exists to remove or prune old checkpoints
- `Open()` loads ALL checkpoint directories found on disk
- Old checkpoint data (snapshot files) is never reclaimed

## Expected Behavior

The system should:
1. Keep only the N most recent **valid** checkpoints (configurable, e.g., default 2)
2. After a new checkpoint is successfully created and verified, delete older checkpoints
3. Ensure the deletion only happens after the new checkpoint is confirmed durable

## Suggested Implementation

- Add a `max_checkpoints` config parameter (default: 2)
- After successful `UpdateMeta()`, invoke a GC pass that removes checkpoints older than the N-th most recent
- Deletion should happen outside the critical path (async or deferred)
- Before deleting, verify the checkpoint is not the one currently in use

### LINKED ISSUE #620: Cypher CHECKPOINT currently updates checkpoint in-place instead of creating a new checkpoint
## Background

While implementing checkpoint retention / GC, we found that the current Cypher `CHECKPOINT;` statement does not create a new `checkpoint-N` directory.

The current path is roughly:

- `CHECKPOINT;` -> `CheckpointOpr` -> `StorageUpdateInterface::CreateCheckpoint()`
- AP path: `StorageAPUpdateInterface::CreateCheckpoint()` calls `graph_.Dump()`
- TP path: `StorageTPUpdateInterface::CreateCheckpoint()` calls `cow_graph_->Dump()`

The no-argument `PropertyGraph::Dump()` dumps into the current `PropertyGraph::ckp_`, so it updates the current checkpoint in place instead of creating a new checkpoint through `CheckpointManager::CreateCheckpoint()`.

## Problem

This makes the user-visible Cypher `CHECKPOINT;` semantics inconsistent with the DB close / recovery checkpoint path:

- DB close / recovery creates a new checkpoint directory, so it can participate in checkpoint retention and GC.
- Cypher `CHECKPOINT;` only overwrites the current checkpoint, does not create checkpoint history, and therefore does not trigger directory-level checkpoint GC.

If users expect `CHECKPOINT;` to publish a durable checkpoint, it should create a new checkpoint directory and then run retention / GC after the new checkpoint is successfully published.

## Suggested Direction

Route user-visible `CHECKPOINT;` through the DB-level checkpoint flow instead of calling `PropertyGraph::Dump()` directly:

- Keep `PropertyGraph::Dump(current_ckp)` as a low-level internal flush primitive.
- Change Cypher `CHECKPOINT;` to create a new checkpoint: `CheckpointManager::CreateCheckpoint()` -> `PropertyGraph::Dump(new_ckp)` -> prune old checkpoints after success.
- On failure, roll back the newly created checkpoint and leave older checkpoints untouched.
- Unify AP and TP semantics. If the current execution layer only has access to `IStorageInterface`, the checkpoint operator and DB / query processor interface boundary may need to be adjusted.

## Acceptance Criteria

- Running `CHECKPOINT;` multiple times creates increasing `checkpoint-N` directories.
- With the default `max_checkpoints = 2`, repeated `CHECKPOINT;` calls retain only the latest two valid checkpoints.
- If `CHECKPOINT;` fails, old checkpoints are not garbage-collected.
- Old checkpoints referenced by active snapshots / readers are protected from deletion.

### LINKED ISSUE #651: [BUG] checkpoint corrupted after manual checkpoint
After `CHECKPOINT`, the copied table lost primary key.

```python
#!/usr/bin/env python3
"""Minimal NeuG CHECKPOINT/COPY primary-key lookup repro.

This is the already-shrunk version of the CodeGraph BELONGS_TO failure. All
non-primary-key properties have been removed, and the repro still holds:

  COPY File
  CHECKPOINT
  COPY Function
  CHECKPOINT
  MATCH (Function) count == 1, but MATCH by Function.id == 0
"""

import csv
import logging
import os
import tempfile
from pathlib import Path

os.environ.setdefault("GLOG_minloglevel", "3")
logging.basicConfig(level=logging.ERROR)
logging.getLogger("neug").setLevel(logging.ERROR)
logging.getLogger("neug.database").setLevel(logging.ERROR)

FILE_ID = "b31986a61becd3c2c030ccadeca6f1f4a2419ffb"
FUNCTION_ID = "78ff0eb259fb0b9e7d24c5342ef277196ce8e4ad"

MINIMAL_PROPS = ["id"]


def write_csv(path: Path, columns: list[str], row: dict[str, object]) -> None:
    with path.open("w", newline="", encoding="utf-8") as f:
        writer = csv.DictWriter(f, fieldnames=columns, quoting=csv.QUOTE_ALL)
        writer.writeheader()
        writer.writerow(row)


def scalar(conn, query: str):
    rows = list(conn.execute(query))
    return rows[0][0] if rows else None


def create_node_table(conn, table: str, props: list[str]) -> None:
    cols = ", ".join(f"{p} STRING" for p in props)
    conn.execute(f"CREATE NODE TABLE {table}({cols}, PRIMARY KEY(id));")


def row_for(table: str, props: list[str]) -> dict[str, object]:
    if table == "File":
        return {"id": FILE_ID}
    if table == "Function":
        return {"id": FUNCTION_ID}
    return {"id": f"{table.lower()}-id"}


def run_once(
    file_props: list[str],
    function_props: list[str],
    class_props: list[str],
    module_props: list[str],
    *,
    checkpoint_between_copies: bool,
) -> tuple[int, int]:
    """Run one isolated database and return (Function row count, id lookup count)."""
    with tempfile.TemporaryDirectory(prefix="neug-pk-shrink-") as tmp:
        root = Path(tmp)
        file_csv = root / "File.csv"
        function_csv = root / "Function.csv"
        write_csv(file_csv, file_props, row_for("File", file_props))
        write_csv(function_csv, function_props, row_for("Function", function_props))

        db = Database(db_path=str(root / "db"), mode="w")
        conn = db.connect()
        try:
            create_node_table(conn, "File", file_props)
            create_node_table(conn, "Function", function_props)
            create_node_table(conn, "Class", class_props)
            create_node_table(conn, "Module", module_props)

            # These rel tables are part of the reduced CodeGraph-like schema.
            # The repro does not need relation rows; only the schema shape.
            conn.execute("CREATE REL TABLE DEFINES_FUNC(FROM File TO Function);")
            conn.execute("CREATE REL TABLE DEFINES_CLASS(FROM File TO Class);")
            conn.execute("CREATE REL TABLE BELONGS_TO(FROM File TO Module);")

            conn.execute(f'COPY File FROM "{file_csv}" (header=true, delim=",", escaping=false);')
            if checkpoint_between_copies:
                # This checkpoint is the trigger: the later Function COPY still
                # creates a row, but Function.id lookup fails after checkpoint.
                conn.execute("CHECKPOINT;")
            conn.execute(
                f'COPY Function FROM "{function_csv}" '
                '(header=true, delim=",", escaping=false);'
            )
            conn.execute("CHECKPOINT;")

            count = scalar(conn, "MATCH (n:Function) RETURN count(n);")
            lookup = scalar(
                conn,
                f"MATCH (n:Function {{id: '{FUNCTION_ID}'}}) RETURN count(n);",
            )
            return int(count), int(lookup)
        finally:
            conn.close()
            db.close()


def main() -> None:
    file_props = MINIMAL_PROPS
    function_props = MINIMAL_PROPS
    class_props = MINIMAL_PROPS
    module_props = MINIMAL_PROPS

    for with_checkpoint in [False, True]:
        count, lookup = run_once(
            file_props,
            function_props,
            class_props,
            module_props,
            checkpoint_between_copies=with_checkpoint,
        )
        print(
            f"checkpoint_between_copies={with_checkpoint}: "
            f"Function rows={count}, Function PK lookup={lookup}"
        )


if __name__ == "__main__":
    main()
```

zhanglei1949:
Fix #531 
Fix #651 
FIx #620


include/neug/storages/checkpoint_manager.h
```diff
@@ -98,6 +99,15 @@ class CheckpointManager {
    */
   std::shared_ptr<Checkpoint> GetCheckpoint(int32_t id) const;
 
+  /**
+   * @brief Prune old valid checkpoints after a successful durable checkpoint.
+   *
+   * Keeps the newest @p max_keep valid checkpoints. Checkpoints listed in
+   * @p protected_ids are never removed. A max_keep of 0 disables pruning.
+   */
+  void PruneOldCheckpoints(size_t max_keep,
```
liulx20: Can it handle the case of multiple processes opening it？Or, is support for multiple processes opening it still being considered for the future?

zhanglei1949: Currently we don't support opening one database with multiple processes. Maybe supported in the future.

CMakeLists.txt
```diff
@@ -48,6 +48,12 @@ option(NEUG_DEFAULT_REL_STORAGE_DIRECTION "Only store fwd direction in rel table
 if(NOT NEUG_DEFAULT_REL_STORAGE_DIRECTION)
     set(NEUG_DEFAULT_REL_STORAGE_DIRECTION BOTH)
 endif()
+set(NEUG_DEFAULT_MAX_CHECKPOINTS "2" CACHE STRING
```
longbinlai: Can be just 1 default checkpoint.

longbinlai (commented):
Can we add doc somewhere to show how to configure this?

zhanglei1949:
> Can we add doc somewhere to show how to configure this?

I would advise against exposing this configuration to end users. It should be restricted to developers only.

CMakeLists.txt
zhanglei1949: Okay

longbinlai (dismissed):
LGTM

zhanglei1949:
NeuG should only support a single checkpoint. We should remove options like max_checkpoints to avoid confusion about multi-checkpoint support.

zhanglei1949:
Closed, the changes will be commit to main via other PRs.

PR #631, opened 2026-06-26
fix: fix SSSP compiles failed on apple clang 15

### LINKED ISSUE #629: [BUG] gds::SSSP extension compiles failed on macos
as titled

```
/Users/shaoyu/neug/extension/gds/src/impl/sssp_pred_impl.cc:118:21: error: reference to local binding 'dist' declared in enclosing function 'neug::gds::SSSPPred::compute'
      double cand = dist + weight;
                    ^
/Users/shaoyu/neug/extension/gds/src/impl/sssp_pred_impl.cc:127:12: note: in instantiation of function template specialization 'neug::gds::SSSPPred::compute()::(anonymous class)::operator()<neug::NbrIterator>' requested here
      relax(it, *it, u, *it);
           ^
/Users/shaoyu/neug/extension/gds/src/impl/sssp_pred_impl.cc:103:11: note: 'dist' declared here
    auto [dist, u] = pq.top();
```

```cpp
auto [dist, u] = pq.top();
auto relax = [&](auto& it, vid_t w, vid_t edge_src, vid_t edge_dst) {
    ......
    double cand = dist + weight;
    ......
};
```

This is a C++20 feature that capturing a structured binding (`dist`) in a lambda format.
However, Apple Clang 15 still rejects it even with -std=c++20.

Similar issue: https://github.com/ml-explore/mlx/issues/3750
A smallest fix method is to copy the variable into a plain local and captures that instead: https://github.com/ml-explore/mlx/pull/3751

```cpp
auto [dist, u] = pq.top();
auto relax = [&, dist = dist](auto& it, vid_t w, vid_t edge_src, vid_t edge_dst) {
    ......
    double cand = dist + weight;
    ......
};
```

Spockkk0225:
## Summary
- Fix Apple Clang builds of SSSP predicate implementation by init-capturing the structured binding distance value in the relaxation lambda.

## Root cause
Apple Clang 15 rejects directly capturing a structured binding from the enclosing function scope in this lambda, even under C++20.

## Validation
- cmake --build build --target neug_gds_extension -j1 was started after the scoped change; the previous same-target build completed successfully before retrying with explicit -j1, with only the existing unused private field warning.

Fixes #629

### DIFF
--- a/extension/gds/src/impl/sssp_pred_impl.cc
+++ b/extension/gds/src/impl/sssp_pred_impl.cc
@@ -106,7 +106,8 @@ void SSSPPred::compute() {
       continue;
     }
 
-    auto relax = [&](auto& it, vid_t w, vid_t edge_src, vid_t edge_dst) {
+    auto relax = [&, dist = dist](auto& it, vid_t w, vid_t edge_src,
+                                  vid_t edge_dst) {
       if (!in_subgraph[w]) {
         return;
       }


longbinlai (approved):
LGTM

PR #644, opened 2026-06-29
fix: support GDS YIELD subset and reordered output columns

### LINKED ISSUE #645: [Bug] GDS YIELD with a column subset or reordering fails at runtime
**Describe the bug**

GDS algorithm functions support an optional `YIELD` clause, but it only works when `YIELD` lists **all** output columns in their **natural order** (e.g. `YIELD node, comp`). As soon as `YIELD` selects a **subset** of columns or **reorders** them, the query fails at runtime with an internal error:

- `YIELD <subset>` → `Runtime exception: column already set: 0`
- `YIELD <reordered columns>` → `Not supported: unsupported record variable tag type: <N>`

Root cause: `bindGDSFunction` prunes the bound output columns down to the `YIELD` subset (`src/compiler/function/gds/gds_algo_function.cpp:117-154`), but the executor still materializes the full, fixed output column set positionally, and the per-column type metadata is paired by **static index** (`src/execution/execute/ops/retrieve/gds_algo.cc:75-82`: `ret_meta.set(meta_data(i).alias(), outputColumns[i].second)`). The two sides only stay aligned when the `YIELD` list is the full set in natural order — which is exactly what all current tests and docs use, so the bug is masked.

**To Reproduce**

1. Build the `gds` extension:
   ```bash
   cmake -S . -B build -DBUILD_EXTENSIONS="gds" -DBUILD_TEST=ON
   cmake --build build --target neug_gds_extension -j$(sysctl -n hw.ncpu)   # Linux: -j$(nproc)
   ```
2. Run the reproduction script below: `python3 gds_yield_repro.py`
3. Equivalent Cypher (on a homogeneous projected graph; `wcc` outputs `node`, `comp`):
   ```cypher
   LOAD gds;
   CALL project_graph('pk', ['person'], {'[person, knows, person]': ''});
   
   -- OK: full list, natural order
   CALL wcc('pk', {concurrency: 1}) YIELD node, comp
   RETURN node.id AS id, comp ORDER BY id LIMIT 3;
   
   -- FAILS: subset (scalar only) -> "column already set: 0"
   CALL wcc('pk', {concurrency: 1}) YIELD comp
   RETURN comp ORDER BY comp LIMIT 3;
   
   -- FAILS: subset (node only) -> "column already set: 0"
   CALL wcc('pk', {concurrency: 1}) YIELD node
   RETURN node.id AS id ORDER BY id LIMIT 3;
   
   -- FAILS: reordered -> "unsupported record variable tag type"
   CALL wcc('pk', {concurrency: 1}) YIELD comp, node
   RETURN node.id AS id, comp ORDER BY id LIMIT 3;
   ```
4. See error on the subset / reordered cases.

Observed results:

| `YIELD` form | Result |
|---|---|
| `node, comp` (full, natural order) | OK |
| `node AS x, comp AS y` (full + rename) | OK |
| `comp` (subset, scalar only) | FAIL: `Runtime exception: column already set: 0` |
| `node` (subset, node only) | FAIL: `Runtime exception: column already set: 0` |
| `comp, node` (reordered) | FAIL: `unsupported record variable tag type: 14` |

**Expected behavior**

`YIELD` on a GDS algorithm should support selecting a subset of the output columns and reordering them (and renaming them with `AS`), returning the correct columns. At minimum, an unsupported `YIELD` form should produce a clear binder-level error rather than an internal `column already set` / `unsupported record variable tag type` runtime failure.

**Screenshots**

N/A (query-level bug, no UI).

 - NeuG version: master
 - OS: macOS (also reproducible on Linux; the bug is platform-independent)
 - Version: 15 (Darwin 25.5.0)
 - Python Version: 3.13

**Additional context**

Reproduction script (`gds_yield_repro.py`), self-contained, prints OK/ERR per case:

```python
#!/usr/bin/env python3
"""Reproduce the GDS YIELD subset/reorder column-misalignment bug.

Prereq: build the gds extension first
    cmake -S . -B build -DBUILD_EXTENSIONS="gds" -DBUILD_TEST=ON
    cmake --build build --target neug_gds_extension -j$(sysctl -n hw.ncpu)

Run from the repo root:
    python3 gds_yield_repro.py
"""

import os
import shutil
import sys
import tempfile

# Make the in-repo `neug` package importable without installing the wheel.
sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                "tools", "python_bind"))

import neug  # noqa: E402  (import for side effects: .so discovery + env setup)
from neug.database import Database  # noqa: E402


def run_case(conn, title, query, expect_ok):
    try:
        result = conn.execute(query)
        rows = list(result)
        preview = [[str(c)[:24] for c in row] for row in rows[:3]]
        tag = "OK " if expect_ok else "OK (unexpected!)"
        print(f"[{tag}] {title}\n      cols={result.column_names()} rows={preview}")
    except Exception as exc:  # noqa: BLE001
        msg = str(exc).split("Error Message:")[-1].strip()
        tag = "ERR (as expected)" if not expect_ok else "ERR (unexpected!)"
        print(f"[{tag}] {title}\n      {msg[:120]}")


def main():
    db_dir = tempfile.mkdtemp(prefix="gds_yield_repro_")
    db = Database(db_path=os.path.join(db_dir, "db"), mode="w")
    try:
        db.load_builtin_dataset("tinysnb")
        conn = db.connect()
        conn.execute("CALL project_graph('pk', ['person'], "
                     "{'[person, knows, person]': ''});")
        conn.execute("LOAD gds;")

        # wcc output columns (static, fixed order): node (NODE), comp (INT64).
        run_case(conn, "A. YIELD node, comp  (full, natural order)",
                 "CALL wcc('pk', {concurrency: 1}) YIELD node, comp "
                 "RETURN node.id AS id, comp ORDER BY id LIMIT 3;", True)
        run_case(conn, "B. YIELD node AS x, comp AS y  (full + rename)",
                 "CALL wcc('pk', {concurrency: 1}) YIELD node AS x, comp AS y "
                 "RETURN x.id AS id, y ORDER BY id LIMIT 3;", True)
        run_case(conn, "C. YIELD comp  (subset: scalar only)  -- BUG",
                 "CALL wcc('pk', {concurrency: 1}) YIELD comp "
                 "RETURN comp ORDER BY comp LIMIT 3;", False)
        run_case(conn, "D. YIELD node  (subset: node only)  -- BUG",
                 "CALL wcc('pk', {concurrency: 1}) YIELD node "
                 "RETURN node.id AS id ORDER BY id LIMIT 3;", False)
        run_case(conn, "E. YIELD comp, node  (reordered)  -- BUG",
                 "CALL wcc('pk', {concurrency: 1}) YIELD comp, node "
                 "RETURN node.id AS id, comp ORDER BY id LIMIT 3;", False)
        conn.close()
    finally:
        db.close()
        shutil.rmtree(db_dir, ignore_errors=True)


if __name__ == "__main__":
    main()
```

liulx20:
Resolve alias/type misalignment by deriving GDS bind aliases and ContextMeta from physical plan meta_data types instead of fixed column indices, and sink results via ContextChunk. Add yield regression tests to extension CI and make dataset archive extraction compatible with Python versions before 3.12.

#645


longbinlai (approved):
LGTM

PR #671, opened 2026-07-01
fix: reject partial from/to in COPY TEMP and auto_detect COPY FROM

### LINKED ISSUE #632: [BUG] COPY FROM / COPY TEMP silently creates node table when only one of from/to is specified
## Problem

When a user specifies only `from` or only `to` (but not both) in a `COPY TEMP` or `COPY FROM` (auto_detect) statement, the binder silently falls through to node table creation, ignoring the partial option. No error or warning is given.

```cypher
// User intends to create an edge table but forgot 'to'
COPY TEMP TempEdge FROM "edges.csv" (header=true, from='TempUser');
// Result: A node table is created. The 'from' option is silently ignored.
```

## Root Cause

In `bind_copy_from.cpp`, the dispatching logic checks `if (fromIt != end && toIt != end)` — if either is missing, it falls through to `bindCopyNodeFromNoSchema()` without any validation.

This affects both:
- COPY TEMP path (L199-208)
- COPY FROM auto_detect path (L266-271)

## Suggested Fix

When `from` or `to` is specified without the other, return an error:

```
Both 'from' and 'to' options are required to create a relationship table.
```

## Scope

This is not introduced by COPY TEMP — the same logic exists in the auto_detect COPY FROM path.

BingqingLyu:
## Related Issues
fix #632

## What does this PR do?
When a user specifies only `from` or only `to` (but not both) in a `COPY TEMP` or auto-detect `COPY FROM` statement, the binder previously fell through silently to node table creation, ignoring the incomplete intent. This PR adds explicit validation that throws a clear error message instead.

## What changes in this PR?
- **`src/compiler/binder/bind/copy/bind_copy_from.cpp`**: Added `THROW_BINDER_EXCEPTION` in two paths:
  - COPY TEMP path (when `copyStatement.isTemporary()` is true)
  - Auto-detect COPY FROM path (when target table does not exist and auto_detect is enabled)
  
  Both throw: `"Both 'from' and 'to' options are required to create a relationship table."`

- **`tools/python_bind/tests/test_copy_temp.py`**: Added `TestFromToValidation` class with 2 tests covering the COPY TEMP path (`test_only_from_raises`, `test_only_to_raises`).

- **`tools/python_bind/tests/test_load.py`**: Added 2 tests in `TestCopyFrom` class covering the auto-detect COPY FROM path (`test_auto_detect_only_from_raises`, `test_auto_detect_only_to_raises`).

### DIFF
--- a/src/compiler/binder/bind/copy/bind_copy_from.cpp
+++ b/src/compiler/binder/bind/copy/bind_copy_from.cpp
@@ -202,6 +202,11 @@ std::unique_ptr<BoundStatement> Binder::bindCopyFromClause(
     if (fromIt != boundOpts.end() && toIt != boundOpts.end()) {
       return bindCopyRelFromNoSchema(statement, boundOpts, /*temporary=*/true);
     }
+    if (fromIt != boundOpts.end() || toIt != boundOpts.end()) {
+      THROW_BINDER_EXCEPTION(
+          "Both 'from' and 'to' options are required to create a "
+          "relationship table.");
+    }
     return bindCopyNodeFromNoSchema(statement, boundOpts, /*temporary=*/true);
   }
 
@@ -265,6 +270,11 @@ std::unique_ptr<BoundStatement> Binder::bindCopyFromClause(
   auto toIt = boundOpts.find(CopyConstants::TO_OPTION_NAME);
   if (fromIt != boundOpts.end() && toIt != boundOpts.end()) {
     return bindCopyRelFromNoSchema(statement, boundOpts);
+  }
+  if (fromIt != boundOpts.end() || toIt != boundOpts.end()) {
+    THROW_BINDER_EXCEPTION(
+        "Both 'from' and 'to' options are required to create a "
+        "relationship table.");
   }
   return bindCopyNodeFromNoSchema(statement, boundOpts);
 }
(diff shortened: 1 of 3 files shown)


src/compiler/binder/bind/copy/bind_copy_from.cpp
```diff
@@ -266,6 +271,11 @@ std::unique_ptr<BoundStatement> Binder::bindCopyFromClause(
   if (fromIt != boundOpts.end() && toIt != boundOpts.end()) {
     return bindCopyRelFromNoSchema(statement, boundOpts);
   }
+  if (fromIt != boundOpts.end() || toIt != boundOpts.end()) {
+    THROW_BINDER_EXCEPTION(
+        "Both 'from' and 'to' options are required to create a "
```
longbinlai, line 276: 感觉这个提示信息有点和代码不符合～ 你说 Both 'from' and 'to' options are required，但是实际是 他们同时可以为空的（上面那个判定条件）。另外一方面，如果同时为空，是不是需要给个warning，然后说默认会自动推导 输入数据的前两列为 from/to,

BingqingLyu: 同时为空的话，那会作为点表load进去；同时都给了的话，就是边表；只给了from或者to的话，我们现在给了报错

longbinlai (approved):
LGTM

PR #672, opened 2026-07-01
fix: gtest EXPECT_THROW macro parsing issue in ArrayValue debug test on clang/macOS

### LINKED ISSUE #673: [BUG] gtest EXPECT_THROW macro parsing issue in ArrayValue debug test on clang/macOS
When using [EXPECT_THROW] with an inline statement block that contains an initializer list, compilation may fail on clang/macOS with.

Example pattern (fails in our environment):
EXPECT_THROW( { std::vector<execution::Value> values = {execution::Value::INT32(1), execution::Value::INT64(2)}; auto value = execution::Value::ARRAY(array_type, std::move(values)); (void)value; }, exception::InvalidArgumentException);

This is test-only and does not affect production runtime behavior.

johnmzz:
Fixes a gtest macro parsing issue in ArrayValueTest on clang/macOS by replacing inline statement blocks with lambda callables.

The previous form used initializer lists inside macro arguments, which could be misparsed by the preprocessor and fail compilation (too many arguments provided to function-like macro invocation). This change keeps test behavior the same while making the test compile reliably across environments.

No production code is changed; test-only refactor.

### DIFF
--- a/tests/unittest/test_column.cc
+++ b/tests/unittest/test_column.cc
@@ -85,22 +85,25 @@ void expect_signature_eq(const ColumnCowSignature& lhs,
 TEST(ArrayValueTest, ConstructorValidatesPayloadShapeInDebug) {
   auto array_type = DataType::Array(DataType::INT32, 2);
 
-  EXPECT_THROW(
-      {
-        std::vector<execution::Value> values = {execution::Value::INT32(1)};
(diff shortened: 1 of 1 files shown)


zhanglei1949 (approved):
LGTM

PR #678, opened 2026-07-02
refactor: checkpoint storage GC foundation

### LINKED ISSUE #531: enhancement: Add checkpoint garbage collection to CheckpointManager

zhanglei1949:
## Summary

Extract the checkpoint storage core / GC foundation from PR #621 into a focused first PR.

- Replace multi-checkpoint discovery with a single durable `CHECKPOINT` pointer plus `checkpoint-N.next` staging publish/discard flow.
- Add storage-level `CheckpointSession` for DB close/recovery checkpoints, including safe publish, reopen, rollback, and old checkpoint cleanup.
- Add runtime-file materialization in `CheckpointFileManager` and update CSR/property graph dump paths to use the staging checkpoint APIs.
- Change `IDataContainer::Fork` / `MMapContainer::Fork` to shared ownership so checkpoint GC can safely remove old checkpoint directories while readers still hold containers.
- Add/update checkpoint GC, file-manager, and safety coverage.

Fix #531


src/main/neug_db.cc
```diff
@@ -319,25 +325,17 @@ void NeugDB::initPlannerAndQueryProcessor() {
...
-  }
-  if (reopen) {
-    // Dump with reopen=true rebuilds the PropertyGraph in-place (Clear + Open),
-    // invalidating raw pointers held by the GraphView. Rebuild the view so it
-    // points to the freshly loaded internal structures.
-    guard.get().mutable_view().Rebuild(*guard.get().mutable_graph());
+  auto checkpoint_graph = guard.get().mutable_graph()->Clone();
+  auto checkpoint_session = CheckpointSession::Begin(
+      checkpoint_mgr_, *snapshot_store_, reopen, *checkpoint_graph);
+  auto status = checkpoint_session.Dump(*checkpoint_graph);
```
zhanglei1949: checkpoint_graph passed twice？

src/storages/graph/checkpoint.cc
```diff
@@ -129,6 +130,11 @@ void Checkpoint::UpdateMeta(CheckpointManifest&& meta) {
       }
       meta_with_relative_paths.set_module(key, std::move(relative_desc));
     }
+    if (!file_mgr_->SyncSnapshotDirectory()) {
+      THROW_IO_EXCEPTION(
+          "Checkpoint::UpdateMeta: failed to fsync snapshot_dir " +
+          snapshot_dir());
+    }
```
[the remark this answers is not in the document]
zhanglei1949: resolved.

src/utils/io/file/file_utils.cc
```diff
@@ -590,6 +590,16 @@ void create_file(const std::string& path, size_t size) {
...
 
+bool fsync_directory(const std::string& dir_path) {
+  int dir_fd = ::open(dir_path.c_str(), O_RDONLY);
+  if (dir_fd < 0) {
+    return false;
+  }
+  bool ok = (::fsync(dir_fd) == 0);
+  ::close(dir_fd);
+  return ok;
+}
```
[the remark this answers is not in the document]
zhanglei1949: resolved

src/main/neug_db.cc
zhanglei1949: resolved.

include/neug/main/neug_db.h
```diff
@@ -305,6 +307,8 @@ class NeugDB {
   void openGraphAndIngestWals();
   void ingestWals(IWalParser& parser, PropertyGraph& graph);
   void initPlannerAndQueryProcessor();
+  std::shared_ptr<Checkpoint> consumeLiveGraphAndCommitCheckpoint(
```
liulx20, line 310: Why does this interface require the `MemoryLevel` parameter?

include/neug/main/neug_db.h
```diff
@@ -314,7 +318,8 @@ class NeugDB {
    * compacts storage timestamps before dumping, and a successful checkpoint
    * resets last_ts_ to 0.
    */
-  void createCheckpoint(bool reopen = true);
+  void createCheckpointAfterRecovery();
+  void createCheckpointOnClose();
```
liulx20, line 337: Will the behavior of these two interfaces differ?

zhanglei1949: Added method-level comments to make the behavior difference explicit: createCheckpointAfterRecovery publishes the recovered graph, reopens it from the new checkpoint, and rolls back to the previous checkpoint if reopening fails; createCheckpointOnClose only publishes the checkpoint and tears down live resources because the DB is closing.

include/neug/main/neug_db.h
zhanglei1949: Removed.

longbinlai (approved):
LGTM

Issue #638, opened 2026-06-29
Add cross-platform tests for the neug npm package


lnfjpt:
As titled.

PR #684, opened 2026-07-02
feat: add npm nightly test

### LINKED ISSUE #681: [BUG] nodejs test entry should be `require('neug')` rather than `require('./lib')`
The nodejs tests expose the internal path:
```
const { Database } = require('../lib');
```

Instead, we should use a package-style entry:
```
const { Database } = require('neug');
```

Spockkk0225:
## Summary
- add a nightly workflow for npm package testing

## Validation
- not run; workflow-only change

fix #681 #638


lnfjpt (approved):
LGTM

PR #694, opened 2026-07-07
refactor: Avoid unnecessary value cast cost when setting properties from context column

### LINKED ISSUE #683: Restore bulk loader fast path for fixed-width property columns
## Background

Commit a281c0dd1d467cc2ca9a584dd90fad9ecfd34242 replaced the Arrow CSV/JSON IO path with the native reader framework. In `src/storages/loader/loader_utils.cc`, the old `set_properties_column` path dispatched by storage type, cast the Arrow `ChunkedArray` to a concrete typed array, and wrote directly into `TypedColumn<T>` for primitive/date/timestamp/string columns.

The new `set_properties_from_context_column` path iterates row by row, calls `ctx_col->get_elem(k)`, materializes an `execution::Value`, switches on `col_type`, unwraps the value, and then writes it into the storage column. This adds per-row virtual calls and `Value` construction/unwrapping on the hot graph-loading path. It is especially expensive for fixed-width columns, and the same pattern is problematic for fixed-length `ArrayColumn`/list-derived data because `get_elem()` recursively materializes child `Value`s.

Commit link: https://github.com/alibaba/neug/commit/a281c0dd1d467cc2ca9a584dd90fad9ecfd34242

## Impact

Graph loading performance regresses after the refactor, because the loader lost the typed-column fast path that avoided generic row-wise `Value` materialization.

## Suggested fix

Restore specialized copy paths in the loader for typed execution columns into storage columns, preserving the `vids` mapping and null handling. In particular:

- Copy primitive/date/timestamp fixed-width columns by casting to the concrete execution column type and writing directly to `TypedColumn<T>`.
- Keep varchar/string resize handling separate, as before.
- Add an optimized path for fixed-length `ArrayColumn` so it can copy the flattened child column without constructing one `Value::ARRAY` per row.
- Add a graph-load benchmark or regression test that covers fixed-width properties and fixed-length arrays.

## Notes

The current generic implementation in `set_properties_from_context_column` is a useful fallback, but it should not be the default hot path for common fixed-width property imports.

### LINKED ISSUE #695: [BUG] Unnecessary value cast overhead when setting properties from context column
**Describe the bug**

In `set_properties_from_context_column` (`src/storages/loader/loader_utils.cc`), when setting property values from a context column to a storage column, the original implementation uses `ctx_col->get_elem(k)` to retrieve each row as a generic `Value` object, then calls `val.GetValue<ctype>()` to convert it back to the typed value. This per-row Value round-trip (creation + type checking + conversion) introduces unnecessary overhead, especially since the context column is already a typed `ValueColumn<T>` with direct access to the underlying typed data array.

**Associated Commits**

- [0c7e4b0bb76607531a23c93d6ebe34981ee82181](https://github.com/alibaba/neug/commit/0c7e4b0bb76607531a23c93d6ebe34981ee82181) — opt1: set prop column with typed value
- [5bc5de3573c173c69f14813ec9071697d62ade97](https://github.com/alibaba/neug/commit/5bc5de3573c173c69f14813ec9071697d62ade97) — avoid unnecessary value cast cost when setting properties from context column

zhanglei1949:
Fixes #695
Fix #683 

Avoid unnecessary value cast cost when setting properties from context column by directly accessing the typed `ValueColumn<T>` data instead of going through the generic `Value` interface (`get_elem()` → `GetValue<ctype>()`).

### DIFF
--- a/src/storages/loader/loader_utils.cc
+++ b/src/storages/loader/loader_utils.cc
@@ -23,12 +23,17 @@
 #include <cctype>
 #include <limits>
 #include <memory>
+#include <mutex>
 #include <ostream>
+#include <shared_mutex>
 #include <sstream>
+#include <string_view>
+#include <type_traits>
 #include <unordered_set>
 
 #include "csv.hpp"
 #include "neug/common/columns/columns_utils.h"
+#include "neug/common/columns/value_columns.h"
 #include "neug/common/types/value.h"
 #include "neug/utils/datetime_parsers.h"
 #include "neug/utils/exception/exception.h"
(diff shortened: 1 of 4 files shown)


src/storages/loader/loader_utils.cc
```diff
@@ -950,63 +955,94 @@ void fillEdgeReaderMeta(label_t src_label_id, label_t dst_label_id,
...
+      << "Storage column type does not match expected type.";
+
+  auto value_col = std::dynamic_pointer_cast<ValueColumn<SRC_T>>(ctx_col);
+
+  if constexpr (std::is_same_v<COL_T, std::string_view>) {
+    // String writes require mutex-protected capacity management.
+    CHECK(mutex != nullptr) << "String column write requires a shared mutex.";
+    auto write = [&](vid_t vid, const auto& s) {
+      std::shared_lock<std::shared_mutex> lock(*mutex);
       if (typed->available_space() <= s.size()) {
```
luoxiaojian, line 972: 这里会有竞争，如果两个线程都要写6bytes，available_space为8bytes。两个线程都能通过这个判断，进入后续的set_value

zhanglei1949: 我看了一下，目前insert_vertices链路并没有涉及到多线程并发，怀疑shared_mutex可能是历史遗留代码。我建议在这个PR里面删掉：https://github.com/alibaba/neug/blob/1a56a03152ee1464acb9f210da931ae1cadd491e/include/neug/storages/graph/vertex_table.h#L291

zhanglei1949: 删掉了shared_mutex相关代码：https://github.com/alibaba/neug/pull/694/changes/70f16e242beb4b911d3f6913835f6a234dba4ab1

liulx20 (approved):

PR #734, opened 2026-07-14
refactor: bulk build vetices/edges from CSV COPY inputs

### LINKED ISSUE #735: [Feature] Optimize bulk load performance for CSV COPY inputs
Currently, bulk loading vertices/edges from CSV via `COPY` reads all batches into memory before insertion, and the bulk load process runs single-threaded. This leads to high memory usage and limits load throughput for large CSV datasets.

- Fuse `DataSource` and `BatchInsert` so batches are consumed as they are produced, avoiding buffering the entire input in memory.
- Parallelize the bulk loading process across multiple threads to improve load throughput.

Keeping the current single-threaded, fully-buffered load path, which is simpler but does not scale for large datasets.

**Additional context**
Related to #734, which implements this optimization by refactoring `csv_read_function`, `batch_insert_edge`/`batch_insert_vertex`, `loader_utils`, and related storage/graph interfaces.

zhanglei1949:
Optimize bulk load from CSV COPY inputs, by 
- Avoid reading all batches into memory by fusing DataSource & BatchInsert.
- Parallelizing the bulk loading process with multiple threads.

Fix #735


src/storages/loader/bundled_edge_csr_loader.cc
```diff
@@ -0,0 +1,775 @@
...
+  EDATA_T Get(size_t row) const {
+    if (values_ != nullptr) {
+      return (*values_)[row];
+    }
+    return column_->get_elem(row).template GetValue<EDATA_T>();
+  }
+
+ private:
+  const IContextColumn* column_;
+  const vector_t<EDATA_T>* values_ = nullptr;
```
liulx20: I think there's no need to expose `vector_t` to the storage layer

zhanglei1949: resolved.

include/neug/storages/loader/loader_utils.h
```diff
@@ -68,32 +72,139 @@ CsvReadConfig build_csv_read_config(
...
+
+  /// Whether GetNextChunk() may be called concurrently by several consumers.
+  virtual bool SupportsConcurrentGetNext() const { return false; }
+
+  /// Stops any background producers and wakes blocked GetNextChunk() calls.
+  virtual void Cancel() {}
+
+  /// Returns the repeatable source backing this supplier, when available.
+  /// Storage uses this hint to select staged bulk build internally.
+  virtual std::shared_ptr<IDataChunkSource> RepeatableSource() const {
```
liulx20: Perhaps we could implement this by spilling intermediates. This would relax the "repeated" constraint and also avoid performing the oid→vid resolution twice.

include/neug/storages/loader/chunk_pipeline_utils.h
```diff
@@ -0,0 +1,234 @@
...
+#include <vector>
+
+#include "neug/storages/loader/loader_utils.h"
+
+namespace neug {
+
+namespace chunk_pipeline_detail {
+
+inline int32_t hardware_worker_count() {
+  auto workers = static_cast<int32_t>(std::thread::hardware_concurrency());
```
liulx20: should this be constrained by parameters such as max_thread_num?

PR #738, opened 2026-07-14
refactor: Avoid unnecessary compact for csr before dump

### LINKED ISSUE #701: Checkpoint performs standalone full compaction before dump, duplicating scan work on large graphs
## Problem

Checkpoint currently runs a full graph compaction before dumping:

```cpp
graph.Compact(MAX_TIMESTAMP);
graph.Dump(ckp, reopen);
```

On LDBC SNB SF1000 this standalone compaction took about 52 minutes before dump even started.

This duplicates work because dump immediately scans the same large vertex/edge structures again to write checkpoint files. For SF1000, the expensive parts include:

- out/in CSR compaction for every edge table
- timestamp reset for mutable CSR entries
- optional neighbor sorting for `sort_key_for_nbr` edge tables
- very large edges such as HASCREATOR, HASMEMBER, LIKES, HASTAG, REPLYOF

## Current behavior

Checkpoint has two full passes:

1. `PropertyGraph::Compact()`
2. module/CSR/column dump

This is costly for tens of billions of edges.

## Expected direction

Fold checkpoint compaction into dump.

During checkpoint dump:

- filter deleted edges/vertices
- reset persisted timestamps
- apply required neighbor sorting
- write compact checkpoint data directly

Keep non-checkpoint compaction paths unchanged for background/manual compaction.

## Acceptance criteria

- `NeugDB::createCheckpoint()` no longer calls standalone `graph.Compact(MAX_TIMESTAMP)`
- loader checkpoint path also avoids separate `graph_.Compact(...)`
- checkpoint output remains equivalent after reopen
- existing manual/background compaction behavior remains intact
- tests cover deleted edge filtering, timestamp reset, and sorted-neighbor checkpoint persistence

### LINKED ISSUE #736: Avoid unneccessary compaction for csr
There are at least two scenerios where we don't actually need a compaction before dump.
- Newly created vertex/edge Table, bulk loaded, no per-vertex/per-edge insertion performed.
- Open Database in AP mode and start TP service, even no changes are made to the database.

zhanglei1949:
Fix #736 
Fix #701


include/neug/storages/csr/mutable_csr.h
```diff
@@ -478,6 +500,9 @@ class EmptyCsr : public TypedCsrBase<EDATA_T> {
   static std::string type_name() {
     return "empty_csr<" + type_name_string<EDATA_T>() + ">";
   }
+
+ private:
+  void mark_edge_data_dirty() override {}
```
liulx20: The semantics of "dirty" here seem a bit odd; is there a better way to handle this?

zhanglei1949: What about `mark_compaction_required()`?

liulx20 (approved):

PR #756, opened 2026-07-17
refactor: skip unchanged tables during checkpoint 

liulx20:
Fixes

### DIFF
--- a/src/storages/graph/property_graph.cc
+++ b/src/storages/graph/property_graph.cc
@@ -1130,10 +1130,7 @@ bool PropertyGraph::AnyTableChanged() const {
 
 const Schema& PropertyGraph::schema() const { return schema_; }
 
-Schema& PropertyGraph::mutable_schema() {
-  schema_changes_.MarkModified();
-  return schema_;
-}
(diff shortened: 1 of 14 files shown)


src/storages/graph/property_graph.cc
```diff
@@ -912,10 +944,22 @@ void PropertyGraph::DumpAndClear(std::shared_ptr<Checkpoint> ckp) {
...
   for (size_t i = 0; i < vertex_label_total_count_; ++i) {
-    if (schema_.is_vertex_label_valid(i) &&
-        !schema_.is_vertex_label_temporary(i)) {
-      vertex_tables_[i].DisassembleTo(store, meta, *ckp);
+    if (!schema_.is_vertex_label_valid(i) ||
+        schema_.is_vertex_label_temporary(i)) {
+      continue;
+    }
+    auto& vtable = vertex_tables_[i];
+    if (vtable.HasChanges()) {
```
zhanglei1949, line 952: So in this implementation, we could only reuse a VertexTable/EdgeTable if all its columns are not modified?
My point is, there are cases that only one column of VertexTable is modified, in this occasion, we still could reuse other columns.

PR #762, opened 2026-07-17
refactor: Correct the behavior of manual checkpoint for AP and TP

### LINKED ISSUE #651: [BUG] checkpoint corrupted after manual checkpoint

zhanglei1949:
Fix #651


src/server/neug_db_session.cc
```diff
@@ -218,13 +265,48 @@ neug::result<std::string> NeugDBSession::Eval(const std::string& req) {
...
+          RETURN_ERROR(checkpoint_status);
+        }
+      }
+    } else {
+      neug::StorageTPUpdateInterface gui(update_txn);
+      GS_AUTO(ctx, ExecutePreparedPipelineInTransaction(
+                       cache_value, mode, db_config_, param_json_obj,
+                       result_schema, update_txn, gui, response));
+      response->mutable_schema()->CopyFrom(result_schema);
+      neug::execution::Sink::sink_results(ctx, gui, response);
```
zhanglei1949: We should first sink, then commit the transaction, not vice verse. @liulx20

include/neug/compiler/gopt/g_physical_analyzer.h
```diff
@@ -47,7 +47,7 @@ struct ExecutionFlag {
   bool schema = false;
   bool batch = false;
   bool create_temp_table = false;
-  bool transaction = false;
+  bool checkpoint = false;
```
zhanglei1949, line 50: Later when we introduce explicit transaction, we will add bool transaction back. @shirly121 @liulx20

include/neug/server/session_pool.h
```diff
@@ -14,46 +14,50 @@
...
 #include "bthread/bthread.h"
 
 namespace neug {
+class CheckpointCoordinator;
 class NeugDBService;
 
-struct SessionLocalContext {
+inline constexpr size_t kSessionContextAlignment = 4096;
+
+struct alignas(kSessionContextAlignment) SessionLocalContext {
```
zhanglei1949: @liulx20 use alignas(kSessionContextAlignment) to ensure the alignment.

include/neug/storages/graph_snapshot_store.h
```diff
@@ -80,6 +82,36 @@ class GraphSnapshotStore {
...
+  /**
+   * Scoped capability for destructive in-place maintenance.
+   *
+   * GraphSnapshotStore owns the precondition checks and grants this handle only
+   * after verifying that the current slot is safe for maintenance. The handle
+   * keeps the low-level mutable/reopen operations out of the public
+   * GraphSnapshotStore API while avoiding dependencies on higher-level
+   * coordinators.
+   */
+  class CheckpointMaintenanceHandle {
```
zhanglei1949: Return a mutable handle for CheckpointCoordinator, to do in-place update for Snapshot.

src/server/neug_db_session.cc
```diff
@@ -218,13 +265,48 @@ neug::result<std::string> NeugDBSession::Eval(const std::string& req) {
...
     CHECK(planner_ != nullptr);
     auto update_txn = GetUpdateTransaction();
-    neug::StorageTPUpdateInterface gui(update_txn);
-    GS_AUTO(ctx, ExecutePipelineInTransaction(
-                     pipeline_cache_, update_txn.statistic(), query, mode,
-                     db_config_, param_json_obj, result_schema, update_txn, gui,
-                     response));
-    response->mutable_schema()->CopyFrom(result_schema);
-    neug::execution::Sink::sink_results(ctx, gui, response);
+    GS_AUTO(cache_value, pipeline_cache_.Get(update_txn.statistic(), query));
```
zhanglei1949: Deal with `checkpoint` inside `update/schema` branch.

src/server/neug_db_session.cc
```diff
@@ -110,13 +138,11 @@ Status validate_flags(AccessMode mode, const physical::ExecutionFlag& flags,
...
 
 template <typename Transaction>
-inline neug::result<execution::Context> ExecutePipelineInTransaction(
-    execution::LocalQueryCache& pipeline_cache, const GraphStats& stats,
-    const std::string& query, AccessMode mode, const NeugDBConfig& db_config,
-    const rapidjson::Document& param_json_obj, neug::MetaDatas& result_schema,
-    Transaction& txn, IStorageInterface& storage_interface,
-    neug::QueryResponse* response) {
-  GS_AUTO(cache_value, pipeline_cache.Get(stats, query));
+inline neug::result<execution::Context> ExecutePreparedPipelineInTransaction(
```
zhanglei1949: Changed to commit transaction after sink. Will this cause performance degradation? @liulx20

include/neug/main/neug_db.h
```diff
@@ -347,15 +352,13 @@ class NeugDB {
...
+   * NeugDBService is running.
    */
   void createCheckpointOnClose();
 
-  friend class NeugDBSession;
   friend class neug::NeugDBService;
 
-  timestamp_t last_compaction_ts_;
-  timestamp_t last_ts_;
   // Configuration and settings
```
zhanglei1949, line 393: last_ts is maintained in CheckpointCoordinator.

src/server/neug_db_session.cc
```diff
@@ -50,38 +52,64 @@
...
-                                 *version_manager_, db_.graph_snapshot_store(),
+  auto ts = version_manager_->acquire_update_timestamp();
+  auto cow_graph = snapshot_store_.CurrentSnapshot().Clone();
+  auto& wal_writer = wal_writer_manager_.Writer(thread_id_);
+  return neug::UpdateTransaction(std::move(cow_graph), alloc_, wal_writer,
+                                 *version_manager_, snapshot_store_,
                                  pipeline_cache_, ts);
 }
 
+result<std::string> NeugDBSession::GetSchema() const {
```
liulx20: Can this avoid being a member function of the Session class?

src/server/neug_db_session.cc
```diff
@@ -160,18 +182,43 @@ inline neug::result<execution::Context> ExecutePipelineInTransaction(
...
     *response->mutable_profile_result() =
         execution::OprTimer::ToProfileResult(timer_ptr.get());
   }
+  // The caller owns commit ordering so graph-backed results can be materialized
+  // while this transaction's snapshot/view is still valid.
   return ctx_res;
 }
 
+template <typename Transaction>
+inline neug::result<execution::Context> ExecutePipelineInTransaction(
```
liulx20: It seems it has never been called.

include/neug/main/admin_statement_executor.h
```diff
@@ -0,0 +1,52 @@
...
+
+  static Status ValidateCheckpointStatement(
+      const execution::CacheValue& cache_value);
+  Status ExecuteAdminAction(const AdminStatementAction& action) const;
+  void BeginProfileIfNeeded(const execution::CacheValue& cache_value,
+                            ProfileContext& profile) const;
+  void BuildCheckpointResponse(const execution::CacheValue& cache_value,
+                               ProfileContext& profile);
+
+  QueryResponse& response_;
```
liulx20: Why is it necessary to hold this response object here?

src/execution/execute/ops/admin/checkpoint.cc
```diff
@@ -33,9 +33,13 @@ class CheckpointOpr : public IOperator {
 neug::result<Context> CheckpointOpr::Eval(IStorageInterface& graph_interface,
```
liulx20, line 33: Should this operator be removed?

include/neug/main/admin_statement_executor.h
zhanglei1949: The class is totally removed now.

src/execution/execute/ops/admin/checkpoint.cc
zhanglei1949: It turns out that we should not remove the CheckpointerOpr and CheckpointOprBuilder, otherwise the pipeline building will fail.

src/server/neug_db_session.cc
zhanglei1949: Removed.

src/server/neug_db_session.cc
zhanglei1949: removed. All transactions now use ExecutePreparedPipeline.

include/neug/transaction/README.md
```diff
@@ -46,13 +46,13 @@ The `VersionManager` state machine has three effective states for read, insert,
...
+When an `UpdateTransaction` is created, `gate_state_` transitions from `kOpen` to `kUpdateExecuting`. It waits for all in-flight insert transactions to finish, but does not block or wait for read transactions. New insert transactions and new update transactions are blocked during this phase; existing and new reads continue.
 
-When `VersionManager::begin_update_commit` is called, the update enters the commit phase (`update_state_`: `1 -> 2`). New reads and new inserts are blocked until the `UpdateTransaction` is committed or aborted. Already-acquired reads continue unaffected on their pinned snapshot.
+When `VersionManager::begin_update_commit` is called, `gate_state_` transitions from `kUpdateExecuting` to `kClosed`. New reads and new inserts are blocked until the `UpdateTransaction` is committed or aborted. Already-acquired reads continue unaffected on their pinned snapshot.
```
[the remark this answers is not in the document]
zhanglei1949: resolved.

src/server/neug_db_session.cc
```diff
@@ -197,34 +212,82 @@ neug::result<std::string> NeugDBSession::Eval(const std::string& req) {
...
     CHECK(planner_ != nullptr);
-    auto update_txn = GetUpdateTransaction();
-    neug::StorageTPUpdateInterface gui(update_txn);
-    GS_AUTO(ctx, ExecutePipelineInTransaction(
-                     pipeline_cache_, update_txn.statistic(), query, mode,
-                     db_config_, param_json_obj, result_schema, update_txn, gui,
-                     response));
-    response->mutable_schema()->CopyFrom(result_schema);
-    neug::execution::Sink::sink_results(ctx, gui, response);
+    auto timestamp_guard = getUpdateTimestampGuard();
```
liulx20: 如果仅是Explain，可以考虑不做这个getUpdateTimestamp吗

zhanglei1949: 修改为先获取read txn来获取 plan，如果为explain，则直接执行explain，不获取update timestamp；如果不是explain，则commit read txn + 获取update timestamp。

src/main/query_processor.cc
```diff
@@ -17,17 +17,40 @@
...
-#include "neug/utils/pb_utils.h"
 
 namespace neug {
 
-result<std::pair<AccessMode, std::shared_ptr<execution::CacheValue>>>
+namespace {
+
+// Validate JSON parameters against the declared parameter types and
+// build a typed ParamsMap; unknown keys are rejected.
+result<execution::ParamsMap> build_params_map(
```
liulx20: This function seems suitable for placement in params_map.h

src/main/query_processor.cc
```diff
@@ -17,17 +17,40 @@
...
+    const rapidjson::Value& parameters,
+    const execution::ParamsMetaMap& params_type) {
+  execution::ParamsMap params_map;
+  if (parameters.IsObject()) {
+    for (const auto& member : parameters.GetObject()) {
+      std::string key = member.name.GetString();
+      auto iter = params_type.find(key);
+      if (iter == params_type.end()) {
```
liulx20: I think it's better to simply ignore any unnecessary keys.

src/main/query_processor.cc
zhanglei1949: resolved.

src/main/query_processor.cc
zhanglei1949: resolved.

include/neug/execution/common/params_map.h
```diff
@@ -25,5 +27,24 @@ struct DataType;
 namespace execution {
 using ParamsMap = std::map<std::string, Value>;
 using ParamsMetaMap = std::map<std::string, DataType>;
+
+inline ParamsMap build_params_map(const rapidjson::Value& parameters,
```
luoxiaojian: 实现放到.cc里

zhanglei1949: 实现挪到了`params_map.cc`里了

longbinlai (commented):
Doc review for `transaction/checkpoint.md` and `transaction/transaction.md`.

First, the good part: the API descriptions are accurate. I cross-checked every user-facing call against the implementation and they all match — `Database(checkpoint_on_close=...)`, `Connection.execute(query, access_mode=...)`, `Session(endpoint)` / `Session.execute(query, access_mode=...)`, the valid `access_mode` values, the CHECKPOINT mode restriction (update/u only, read-only DB rejected), the inferred-update behavior, and the EXPLAIN/PROFILE CHECKPOINT semantics. No doc/API mismatch found.

The inline comments below are about example runnability and a couple of clarity gaps.

One pre-existing nit I could not anchor inline because the line is not part of this diff: `tools/python_bind/neug/database.py:97` says the `mode` parameter 'Default is r for read-only', but the actual default (line 82) is `"read-write"`. Since this PR already edits docstrings in that file, worth fixing while you're here.

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+conn.close()
+db.close()
+```
+
+### Service mode example
+
+```python
+from neug import Session
+
+session = Session("http://localhost:10000/")
```
longbinlai, line 80: This doc assumes a NeuG service is already listening on :10000, but this page never shows how to start one (`db.serve()`). A one-line pointer to the serve/TP setup doc would help readers who land on this page directly.

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+Checkpoints do not make a temporary in-memory database survive after it is
+closed.
+
+## Failure and recovery
+
+On startup, NeuG opens the newest completely published checkpoint; an
+incomplete generation is never selected. In Service mode, NeuG then replays
+WAL records created after that checkpoint.
+
+A manual `CHECKPOINT` fails in one of two ways:
```
longbinlai: This section only covers manual `CHECKPOINT` failures. Two gaps: (1) it doesn't say what happens when a *recovery* checkpoint fails — in the code, the fail-stop (`LOG(FATAL)`) only applies to the manual path (`reason == Reason::kManual` in `checkpoint_coordinator.cc`); a recovery-path failure instead returns an error and fails the database open. Worth a sentence so readers don't assume recovery also terminates the process. (2) 'deliberately terminates the database process' is a surprising behavior — consider making explicit that this is intentional and that a restart recovers cleanly, so users don't read it as a bug.

doc/source/transaction/transaction.md
```diff
@@ -171,24 +147,22 @@ When executing queries, you can specify an `access_mode` to control transaction
...
-read < insert < update
-```
-
-`schema` uses the same serialized update path as `update`.
-
-- **Upward compatibility**: Using a higher access mode than required always works (e.g., `update` mode for a read-only query), but may reduce throughput due to stronger locking
-- **Downward restriction**: Using a lower access mode than required causes an error (e.g., `read` mode for an insert operation)
+**Access Mode Capabilities:** Access modes are operation-specific rather than a strict hierarchy. In Service mode, `read` and `insert` are narrow modes, while `update` and `schema` use the general serialized update path. Using a mode that does not support the planned operation returns an error.
 
 | Specified Mode | Actual Operation | Result |
```
longbinlai, line 154: The narrowing shown here (e.g. `insert` + read/update → Error) is only enforced in Service mode: `ExecutionSlot::validatePlan` gates the insert-only check behind `execution_strategy_ == QueryExecutionStrategy::kTransactional`. In Embedded mode an `insert` access mode on a read query silently takes the exclusive lock and runs instead of erroring. The intro sentence and the note below mention this, but the table itself reads as universal. Consider labeling the table 'Service mode' (or adding a Mode column) so readers don't apply it to Embedded mode.

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+For transaction boundaries and concurrency outside checkpoint operations, see
+[Transaction Management](transaction.md).
+
+## Run a checkpoint
+
+```cypher
+CHECKPOINT;
+```
+
+`CHECKPOINT` takes no arguments and runs under the `update` access mode.
```
longbinlai: 这里的描述比较奇怪，一开始我以为 set it explicitly 是可以设置，但是读到后来好像只能设置成 update. 建议修改

`CHECKPOINT` takes no arguments and must run under the `update` access mode.

If `access_mode` is omitted, NeuG infers `update`. If it is specified
explicitly, it must be `"update"` or `"u"`. Any other access mode
(`"read"`/`"r"`, `"insert"`/`"i"`, or `"schema"`/`"s"`) is rejected, and
a database opened read-only cannot create a checkpoint. 

Usage examples: 

```python
# NeuG infers `update` access mode
conn.execute("CHECKPOINT") 
```

```python
 # or "u", and all other access modes will be rejected
conn.execute("CHECKPOINT", access_mode="update") 
```

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+
+```python
+conn.execute("CHECKPOINT")
+conn.execute("CHECKPOINT", access_mode="update")  # or "u"
+```
+
+Any other explicit mode (`read`/`r`, `insert`/`i`, or `schema`/`s`) is
+rejected, and a database opened read-only cannot create a checkpoint.
+
+- `EXPLAIN CHECKPOINT` returns the execution plan without creating a
```
longbinlai, line 47: 这两个出现的比较突然，建议前面补充一个。`CHECKPOINT` can be used with [`EXPLAN/PROFILE` clause](../cypher_manual/explain_profile.md)

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+```python
+import neug
+
+# This application checkpoints explicitly and checks the result of each
+# call, so it disables the automatic checkpoint on close.
+db = neug.Database("/path/to/database", checkpoint_on_close=False)
+conn = db.connect()
+
+conn.execute("COPY Person FROM 'people.csv'")
+conn.execute("CHECKPOINT")  # The loaded data is now durable.
```
longbinlai:
```
conn.execute("COPY Person FROM 'people.csv'") # The loaded data is in memory
conn.execute("CHECKPOINT")  # The loaded data has now been persisted to disk.
```

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+
+conn.close()
+db.close()
```
longbinlai, line 68: 加个 comment吧。

```
# If the database is opened with `checkpoint_on_close=True`, e.g.
#   db = neug.Database("/path/to/database", checkpoint_on_close=True)
# or, as `checkpoint_on_close` defaults to True,
#   db = neug.Database("/path/to/database")
# this call triggers an implicit checkpoint.
db.close()
```

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+    access_mode="insert",
+)
+
+# Optional maintenance: fold committed WAL records into a fresh snapshot.
+session.execute("CHECKPOINT")
+session.close()
+```
+
+Closing a client `Session` only disconnects that client; it neither closes
+nor checkpoints the server database.
```
longbinlai: 这里 我有个疑问，如果用户在这里做了一些写操作，然后 server 那边在关闭 db （或者那边关闭 service mode，转成embedded mode 执行） 的时候做了checkpoint，这种情况下 WAL 也会被consume 掉吗？

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+   recovery point; the previous generation is no longer needed for recovery.
+4. Reopen the live database on the new generation and resume normal
+   operation. In Service mode this also starts a fresh, empty WAL.
+5. Remove the retired generation (best effort).
+
+Because step 2 writes a full copy while the current generation still exists,
+peak disk usage during a checkpoint is roughly twice the database size, and
+checkpoint time grows with database size. Disk usage drops back once the
+retired generation is removed in step 5. Run checkpoints as often as your
+recovery-point and WAL-size requirements justify, rather than on a fixed
```
longbinlai: `Run checkpoints as often as your recovery-point and WAL-size requirements justify` 这句话我不太懂什么意思？

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+
+Use an explicit `CHECKPOINT` when the application must know whether
+persistence succeeded. If you set `checkpoint_on_close=False`:
+
+- Embedded mode discards changes made after the last successful checkpoint
+  when the database closes.
+- Service mode can recover committed changes from the WAL, even if they were
+  made after the last checkpoint.
+
+Checkpoints do not make a temporary in-memory database survive after it is
```
longbinlai: 感觉这句话可以删掉，前面已经说得比较清楚了。

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,176 @@
...
+  begin maintenance — the statement returns an error and the database keeps
+  running normally.
+- **After the snapshot build has started**, a failure can leave the
+  in-memory state undefined, so NeuG deliberately terminates the database
+  process rather than continue. Restart the database to recover from the
+  latest published checkpoint and, in Service mode, the WAL.
+
+### Recovering from a failed Embedded write
+
+Large Embedded-mode writes are not fully atomic. To make a failed batch
```
longbinlai: 我们统一说 write in the embedded mode吧 （避免 embedded-mode write / embedded write 这种说法）。同时，这里可能需要补个: we currently do not support bulk load (`COPY/LOAD FROM`) data in the service mode.

longbinlai: 或者这里直接说 bulk load ，而不是 large write （这个说法也有点歧义） 吧。

doc/source/transaction/checkpoint.md
zhanglei1949: resolved

doc/source/transaction/checkpoint.md
zhanglei1949: resolved.

doc/source/transaction/transaction.md
zhanglei1949: resolved.

doc/source/transaction/checkpoint.md
zhanglei1949: resolved.

doc/source/transaction/checkpoint.md
zhanglei1949: resolved.

doc/source/transaction/checkpoint.md
zhanglei1949: added.

doc/source/transaction/checkpoint.md
zhanglei1949: My version:
```
db = neug.Database("/path/to/database", checkpoint_on_close=False)

# With the default checkpoint_on_close=True, db.close() would trigger an
# implicit checkpoint. This example disables it and checkpoints explicitly.
db.close()
```

doc/source/transaction/checkpoint.md
zhanglei1949: wal会被直接清理，因为wal是放在checkpoint目录里面的。
例如，neug刚开始在ckp-N上提供服务，从client过来一些写入被成功执行，则在ckp-N/wal/wal.xxx里会有写入。
然后我们执行checkpoint，这个时候内存里面的图本身就是最新的，所以直接dump到新的ckp-(N+1)，完成之后会清理掉ckp-N（包括目录底下的wal）

doc/source/transaction/checkpoint.md
zhanglei1949: 改成
```
Schedule checkpoints based on how much work you can afford to redo after a crash (your recovery point) and how large you want the WAL to grow, rather than on a fixed tight interval.
```
其实之后我们应该也加一下wal size增长到一定程度之后，自动checkpoint。

doc/source/transaction/checkpoint.md
zhanglei1949: removed.

doc/source/transaction/checkpoint.md
zhanglei1949: changed to `Recovering from a failed bulk load in Embedded mode`

tools/python_bind/neug/database.py
```diff
@@ -94,7 +94,7 @@ def __init__(
             Note that in memory mode, the database will not be persisted to disk, and all data will be
             lost when the program exits. In this case, the db_path should not contain any illegal characters.
         mode : str
-            Mode to open the database, could be 'r', 'read', 'readwrite', 'w', 'rw', 'write'. Default is 'r' for read-only.
+            Mode to open the database, could be 'r', 'read', 'readwrite', 'w', 'rw', 'write'. Default is 'read-write'.
```
longbinlai: so is it read-write, or readwrite?

doc/source/transaction/checkpoint.md
```diff
@@ -0,0 +1,201 @@
...
+conn.execute("CHECKPOINT", access_mode="update")
+```
+
+`CHECKPOINT` can also be used with an
+[`EXPLAIN`/`PROFILE` clause](../cypher_manual/explain_profile.md):
+
+- `EXPLAIN CHECKPOINT` returns the execution plan without creating a
+  checkpoint.
+- `PROFILE CHECKPOINT` creates the checkpoint and reports its execution time
+  as a single `Checkpoint` operator.
```
longbinlai: This should be `CHECKPOINT` (all cypher keyword capitalized by default)

tools/python_bind/neug/database.py
zhanglei1949: changed to `readwrite`.

doc/source/transaction/checkpoint.md
zhanglei1949: ok.

longbinlai (dismissed):
LGTM

src/main/execution_slot.cc
```diff
@@ -248,47 +295,67 @@ Status ExecutionSlot::executeCore(const std::string& query,
...
   const auto access_mode = requested_mode == AccessMode::kUnKnown
-                               ? planner_->analyzeMode(query)
+                               ? analysis.access_mode
                                : requested_mode;
   std::shared_ptr<execution::CacheValue> prepared_query;
 
-  auto execute_on_storage =
-      [this, &query, access_mode, &parameters, num_threads, &response,
-       &prepared_query](const GraphStats& stats, auto& storage) -> Status {
+  if (NEUG_UNLIKELY(analysis.checkpoint())) {
```
liulx20: EXPLAIN CHECKPOINT 仍走 validateCheckpointRequest，文档写 EXPLAIN 不落盘，但只读库 / access_mode=read 会先被拒。

zhanglei1949: resolved.

liulx20 (dismissed):

liulx20 (approved):

PR #726, opened 2026-07-10
feat: support fixed-size arrays in data readers

shirly121:
## Summary

- support parsing explicitly typed fixed-size arrays from CSV `LOAD FROM`, including nested arrays, quoted strings, temporal values, and length validation
- support fixed-size array values in the JSON reader and when materializing `COPY FROM` results into graph properties
- support Arrow `FIXED_SIZE_LIST` columns in the Parquet reader, including nested arrays and scalar child types
- add end-to-end CSV, JSON, and Parquet array coverage and document the explicit `CAST(..., T[N])` workflow

Variable-length `LIST` types such as `FLOAT[]` remain unsupported by these readers. Users must explicitly specify a fixed-size array type such as `FLOAT[3]`; implicit LIST-to-ARRAY conversion is not performed.

## Testing

- `python3 -m pytest -q tools/python_bind/tests/test_load_array.py` — 15 passed

Fixes #441
Fixes #442
Fixes #443

PR #768, opened 2026-07-22
fix: fix Parquet array CI failures

shirly121:
## Summary

This PR fixes CI regressions introduced by feat: support fixed-size arrays in data readers (#726), which caused the extension CI job to fail.

## Fixes

- Keeps the null fixed-size ARRAY test but marks it as skipped because PyArrow cannot write null fixed-size lists to Parquet and raises `ArrowNotImplementedError: Lists with non-zero length null components are not supported` before NeuG can read the fixture.
- Stores interval values in the Parquet fixture as fixed-size string arrays (for example, `2 days` and `3 hours`) and explicitly casts them to `INTERVAL[2]`. This verifies that NeuG returns intervals in the expected human-readable format instead of millisecond strings.
- Removes the obsolete `timedelta` import.

These changes keep the extension test suite stable while preserving coverage for fixed-size Parquet ARRAY ingestion.


lnfjpt (approved):
LGTM

PR #782, opened 2026-07-27
refactor: Refactor SessionPool,Session, remove query processor

### LINKED ISSUE #783: Unify AP and TP query execution ownership and version lifecycle
## Problem

NeuG has separate AP and TP query execution stacks with overlapping responsibilities. Embedded queries depend on AP-specific session/query-processor plumbing, while service queries use a separate TP session pool and transactional entry. This creates several structural problems:

- query preparation, pipeline execution, storage dispatch, and result handling can drift between AP and TP;
- `NeugDB` is coupled to execution-slot lease/release mechanics instead of containing only database-wide resources;
- `Connection` does not own its execution context, making ownership and shutdown behavior harder to reason about;
- AP/TP mode transitions and checkpoint recovery lack one clearly defined version-manager timeline;
- TP-only resources such as WAL writers are not naturally scoped to the TP execution pool;
- duplicated lifecycle and concurrency code makes deterministic AP/TP behavior difficult to test.

## Desired outcome

- Use one core `ExecutionSlot` abstraction for shared query preparation and pipeline execution.
- Let each embedded `Connection` exclusively own its execution slot.
- Let `TpExecutionSlotPool` own and lease TP execution slots together with their WAL writers.
- Keep `NeugDB` focused on database-wide resources such as snapshots, version management, planning, cache, and allocators.
- Share one AP/TP version timeline and preserve checkpoint visibility semantics.
- Make TP capability restrictions and WAL ownership explicit, with regression coverage for storage/WAL side effects and deterministic concurrency.

## Scope

This issue covers the P2 execution-context and ownership consolidation. Further policy unification and intra-query runtime work remain follow-up P3/P4 work.

zhanglei1949:
## Summary

- Consolidate the duplicated AP and TP execution plumbing into a shared `ExecutionSlot` for query preparation and pipeline execution.
- Make embedded `Connection` instances exclusively own their execution slots, while `TpExecutionSlotPool` owns TP slots together with their WAL writers.
- Remove execution-slot lease queues and session-management details from `NeugDB`, leaving it responsible for database-wide snapshots, version management, planning, cache, and allocators.
- Keep explicit adapters for structured embedded queries (`ExecuteQuery`) and serialized transactional requests (`ExecuteTransactionalRequest`) while sharing the core execution path.
- Preserve a shared AP/TP version timeline and AP checkpoint visibility semantics, and add regression coverage for capability restrictions, WAL/storage side effects, checkpoint round trips, and deterministic concurrency.

This PR implements the execution-context and ownership consolidation. Execution-policy unification and intra-query runtime work remain follow-up work.

Fixes #783


include/neug/main/connection.h
```diff
@@ -139,7 +133,7 @@ class Connection {
    * @since v0.1.0
    */
   result<QueryResult> Query(const std::string& query_string,
-                            const std::string& access_mode = "update",
+                            const std::string& access_mode = "",
```
zhanglei1949, line 141: Let planner analyze, rather than by default `update`.

include/neug/main/execution_slot.h
```diff
@@ -0,0 +1,275 @@
+/** Copyright 2020 Alibaba Group Holding Limited.
```
zhanglei1949, line 1: Renamed from `NeugDBSession`.

include/neug/server/tp_execution_slot_pool.h
```diff
@@ -0,0 +1,185 @@
+/** Copyright 2020 Alibaba Group Holding Limited.
```
zhanglei1949, line 1: Renamed from `SessionPool`.

src/storages/graph/graph_interface.cc
```diff
@@ -78,7 +78,7 @@ Status StorageAPUpdateInterface::AddEdgeImpl(
   int32_t oe_offset = 0;
   auto status =
       graph_.AddEdge(src_label, src, dst_label, dst, edge_label, properties,
-                     neug::timestamp_t(0), alloc_, oe_offset, prop, true);
+                     kMutationTimestamp, alloc_, oe_offset, prop, true);
```
liulx20: What is the difference between `read_ts` and `kMutableTimestamp`?

src/server/tp_execution_slot_pool.cc
```diff
@@ -0,0 +1,47 @@
...
+                                               size_t slot_id) noexcept {
+  auto* pool = static_cast<TpExecutionSlotPool*>(owner);
+  bthread_mutex_lock(&pool->mutex_);
+  CHECK_LT(slot_id, pool->slot_num_);
+  CHECK_LT(pool->available_slot_ids_.size(), pool->slot_num_);
+  CHECK_GE(pool->available_slot_ids_.capacity(), pool->slot_num_);
+  pool->available_slot_ids_.push_back(slot_id);
+  VLOG(10) << "Released slot_id=" << slot_id;
+  bthread_cond_signal(&pool->cond_);
+  bthread_mutex_unlock(&pool->mutex_);
```
liulx20, line 44: Can we use std functions or mechanisms like spinlocks, given that brpc might not be supported on Windows in the future?

src/transaction/version_manager.cc
```diff
@@ -201,6 +201,16 @@ void VersionManager::begin_update_commit(uint32_t ts) {
   }
 }
 
+void VersionManager::drain_readers() {
+  if (update_state_.load(std::memory_order_acquire) != 2) {
+    THROW_INTERNAL_EXCEPTION(
+        "drain_readers called outside update commit state");
+  }
+  while (active_readers_.load(std::memory_order_acquire) > 0) {
+    // P4a will replace this wait loop with a runtime-aware wait strategy.
```
liulx20, line 210: Does this need to be implemented?

src/main/execution_slot.cc
```diff
@@ -0,0 +1,519 @@
...
+                             const std::string& requested_mode) {
+  auto mode = requested_mode.empty() ? AccessMode::kUnKnown
+                                     : ParseAccessMode(requested_mode);
+  if (mode == AccessMode::kUnKnown) {
+    mode = planner->analyzeMode(query);
+  }
+  return mode;
+}
+
+result<execution::ParamsMap> parseJsonParameters(
```
liulx20: This could perhaps be moved to params_map.h.

src/main/execution_slot.cc
```diff
@@ -0,0 +1,519 @@
...
+}  // namespace
+
+ReadTransaction ExecutionSlot::GetReadTransaction() const {
+  uint32_t ts = version_manager_.acquire_read_timestamp();
+  SnapshotGuard guard(snapshot_store_);
+  return ReadTransaction(std::move(guard), version_manager_, ts);
+}
+
+InsertTransaction ExecutionSlot::GetInsertTransaction() {
+  if (wal_writer_ == nullptr) {
```
liulx20: Do we have to handle this here? Perhaps it could be checked when switching from AP to TP.

src/server/tp_execution_slot_pool.cc
zhanglei1949: This indeed should be considered in the future. But I will not include it in this PR. https://github.com/alibaba/neug/issues/785

src/storages/graph/graph_interface.cc
zhanglei1949: 值 | 含义 | 主要用途
-- | -- | --
read_ts_ | 当前操作持有的可见性 / VM lease timestamp | 判断已有顶点或边在当前视图中是否有效
kMutationTimestamp (0) | AP 写入数据时使用的固定持久化基线时间戳 | 新增顶点、边，以及更新边属性

read_ts_ 用于“读已有数据”：例如更新顶点属性或删除顶点/边前，需要确认该记录在当前操作的可见版本中存在。
kMutationTimestamp 用于“写入持久数据”。AP 模式的 VersionManager 高水位不会随 checkpoint/WAL 持久化；数据库重启后，可见性时间戳可能从 0 重新开始。因此，AP 新写入的数据必须使用 0 作为存储版本，否则写入时若使用较大的 read_ts_，重启后该记录可能在 read_ts = 0 时不可见。
一句话：
read_ts_ 用于当前操作的可见性校验；kMutationTimestamp 用于保证 AP 写入数据在重启后仍从基线可见。

zhanglei1949: 这里的改动应该主要是抽了一个static constexpr的成员变量，并且增加了一些comments，描述这里为什么使用0（即便我们已经使用VersionManager来控制并发，获取ts了）。

src/transaction/version_manager.cc
zhanglei1949: Currently here is a tight loop. In later PR, we will provide different implementation for `pthread` and `bthread`.

src/main/execution_slot.cc
zhanglei1949: Removed. Those `GetXXXTransaction` are only called in TP, where the `WalWriter` must be provided.

src/main/execution_slot.cc
zhanglei1949: Moved to parseJsonParameters.

src/main/execution_slot.cc
```diff
@@ -0,0 +1,492 @@
...
+        std::chrono::duration_cast<std::chrono::microseconds>(end - start)
+            .count());
+    ++query_num_;
+  }
+  return result;
+}
+
+result<std::string> ExecutionSlot::ExecuteTransactionalRequest(
+    const std::string& request) {
+  if (wal_writer_ == nullptr) {
```
liulx20: Same issue as GetTransaction.

src/main/execution_slot.cc
```diff
@@ -0,0 +1,492 @@
...
+  rapidjson::Document parameters_json;
+  auto parse_result =
+      RequestParser::ParseFromString(request, query, mode, parameters_json);
+  if (!parse_result.ok()) {
+    RETURN_ERROR(parse_result);
+  }
+  if (mode == AccessMode::kUnKnown) {
+    mode = planner_->analyzeMode(query);
+  }
+  constexpr ExecutionCapabilities capabilities{
```
liulx20, line 365: how to configure this?

include/neug/main/execution_slot.h
```diff
@@ -0,0 +1,277 @@
...
+ * TpExecutionSlotPool release their slots before NeugDB destroys those shared
+ * dependencies.
+ *
+ * @see NeugDBService for HTTP service wrapper
+ * @see TpExecutionSlotPool for execution-slot management
+ * @since v0.1.0
+ */
+class ExecutionSlot {
+ public:
+  static constexpr int32_t MAX_RETRY = 3;
```
liulx20: This constexpr has not been used.

include/neug/execution/common/params_map.h
```diff
@@ -18,12 +18,30 @@
...
+#include "neug/utils/result.h"
 
 namespace neug {
 
-struct DataType;
 namespace execution {
 using ParamsMap = std::map<std::string, Value>;
 using ParamsMetaMap = std::map<std::string, DataType>;
+
+inline result<ParamsMap> parseJsonParameters(
```
liulx20: 这个函数以 inline 形式进了被广泛 include 的 params_map.h,把 rapidjson 依赖带了进去,可以考虑只留声明。

longbinlai (commented):
## Review summary

Architecture refactor is clean — AP/TP single-track, slot count derived from `max_thread_num`, RAII lease, 4KB-aligned entries. Product-facing impact:

- **Python/HTTP/Node.js users**: no behavior change. Python binding was already passing `access_mode=""` and C++ was already inferring.
- **C++ direct users**: three breaking renames (`AcquireSession` → `AcquireExecutionSlot`, `Session::Eval` → `ExecutionSlot::ExecuteTransactionalRequest`, `RegisterService` → private `registerService`).
- **One pre-existing doc bug now visible**: `tools/python_bind/src/py_connection.cc:51-52` says "If the access mode is not specified, it will be set to `update` by default" — but Python binding was always passing `""` and C++ was always inferring via `planner->analyzeMode`. Fixing the docstring is a follow-up but worth flagging.
- **`registerService` new `closeAllConnections()` throw**: documented as "none may be in use" but `Connection::Close()` doesn't actually throw for idle connections. Defensive try/catch is fine, but Python `PyDatabase::serve()` should give users a clear error when they have an outstanding `PyConnection`.

Detailed inline comments below.

longbinlai (commented):
## Re-review with corrected line positions

The previous review's inline comments landed on slightly-off lines (the `position` field GitHub reports doesn't match `line` 1:1 for multi-hunk files). Re-posting with positions verified against the actual files at PR HEAD commit `d2c091d`.

Key findings remain the same:
- `access_mode=""` default change: **no impact on Python users** (Python binding always passed empty string and C++ was always inferring). But the Python docstring at `py_connection.cc:51-52` claiming "set to update by default" is wrong and should be fixed in a follow-up.
- `Connection` lifetime contract is now stricter and not documented.
- `registerService` + `closeAllConnections` is defensive but doesn't actually reject in-use connections — Python `PyDatabase::serve` should validate upfront.
- `wal_writer_==nullptr` invariant and slot-count-equals-max_thread_num are not enforced / documented.

Inline comments are now anchored on the actual lines.

longbinlai (commented):
## Review v3 — comments anchored on actual diff positions

Verified against the PR HEAD diff. Each comment is now on the exact code line.

Quick recap of the four issues:
1. **`access_mode = ""` default** — no impact on Python users (Python binding was always passing `""` and C++ was always inferring via `planner->analyzeMode`). The docstring at `tools/python_bind/src/py_connection.cc:51-52` is pre-existing wrong, worth a follow-up fix.
2. **Connection lifetime** — Connection's `unique_ptr<ExecutionSlot>` borrows references to NeugDB-owned resources. The new lifetime contract needs a `@warning` in the class docstring.
3. **`registerService` + `closeAllConnections`** — defensive try/catch is fine, but `Connection::Close()` doesn't actually throw for in-use connections. Python `PyDatabase::serve` should validate outstanding `PyConnection`s upfront.
4. **`wal_writer_` invariant + slot count = max_thread_num** — both should be enforced / documented explicitly.

src/main/execution_slot.cc
```diff
@@ -0,0 +1,492 @@
...
+      .batch = true,
+      .temporary_table = true,
+  };
+
+  auto execute = [&](const GraphStats& stats,
+                     StorageReadInterface& storage) -> result<QueryResult> {
+    GS_AUTO(cache_value,
+            prepareQuery(stats, query_string, mode, capabilities, num_threads));
+    execution::ParamsMap parsed_parameters;
+    if (parameters != nullptr) {
```
liulx20: 这里的参数处理为啥会有两个分支，有办法统一吗

longbinlai (commented):
## Consolidated review — please ignore my three earlier attempts

I posted three noisy reviews earlier (IDs 4786452202 / 4786478140 / 4786486965) trying to chase diff positions. The author pushed `resolve review comments` between attempts, which orphaned most of those comments. This single review supersedes them. **Please dismiss the three earlier reviews** — they are duplicates and only the `Behavior change for C++ direct users` thread at `connection.h:84` is still anchored to a live diff position.

### 4 product-impact concerns, anchored to the current head `d2c091d3`

1. `access_mode = ""` default change (C++ direct users only)
2. `Connection` now exclusively owns an `ExecutionSlot` — lifetime contract is implicit
3. `registerService()` swallows `closeAllConnections()` errors, but the doc still says "none may be in use"
4. `wal_writer_` is the *sole* TP-vs-AP distinguisher, with no runtime enforcement

Details in the inline comments. None of these block the refactor; they are correctness/clarity follow-ups.

src/main/execution_slot.cc
```diff
@@ -0,0 +1,492 @@
...
+  const auto& flags = cache_value->flags;
+
+  if ((!capabilities.batch && flags.batch()) ||
+      (!capabilities.temporary_table && flags.create_temp_table())) {
+    RETURN_ERROR(Status(
+        StatusCode::ERR_NOT_SUPPORTED,
+        "Temporary table creation and batch operations are not supported "
+        "for TP service."));
+  }
+  if (db_config_.mode == DBMode::READ_ONLY && !IsReadOnlyExecutionFlag(flags)) {
```
liulx20: 这里还需要check mode吗

longbinlai (commented):
## Consolidated review — supersedes my three earlier attempts

I posted three noisy reviews earlier (IDs 4786452202 / 4786478140 / 4786486965) chasing diff positions; the author pushed `resolve review comments` between attempts and most of those comments are now orphaned. **Please dismiss the three earlier reviews** — they are duplicates and only the `Behavior change for C++ direct users` thread at `connection.h:84` is still anchored to a live diff position.

The four concerns below are anchored to the **current head `d2c091d3`** with verified `position` values.

include/neug/main/connection.h
```diff
@@ -139,7 +133,7 @@ class Connection {
...
                             const execution::ParamsMap& parameters = {});
```
longbinlai, line 142: **Behavior change for C++ direct users, no change for Python users.**

C++ header default was `access_mode = "update"`, now `""`. C++ callers relying on implicit `update` (e.g. `conn->Query("CREATE ...")` with no second arg) will now get planner-inferred mode. The C++ docstring was updated to say *"empty string: Infer access mode from query text"*.

For Python, the binding at `tools/python_bind/src/py_connection.cc:90` has always passed `access_mode=""` explicitly, and the old `QueryProcessor::execute` already did `if (requested_mode.empty()) infer via planner->analyzeMode(query)` — Python users have been getting inference all along. **No Python behavior change.**

The Python docstring at `tools/python_bind/src/py_connection.cc:47-52` still says *"...will be set to `update` by default."* — pre-existing inconsistency (binding never sent `"update"`), not introduced by this PR, but worth a follow-up doc fix to avoid confusion for users reading both sides.

longbinlai (commented):
## Consolidated review — supersedes my three earlier attempts

I posted three noisy reviews earlier (IDs 4786452202 / 4786478140 / 4786486965) chasing diff positions; the author pushed `resolve review comments` between attempts and most of those comments are now orphaned. **Please dismiss the three earlier reviews** — they are duplicates and only the `Behavior change for C++ direct users` thread at `connection.h:84` is still anchored to a live diff position.

The four concerns below are anchored to the **current head `d2c091d3`** with verified `position` values.

include/neug/main/connection.h
longbinlai, line 142: **Behavior change for C++ direct users, no change for Python users.**

C++ header default was `access_mode = "update"`, now `""`. C++ callers relying on implicit `update` will now get planner-inferred mode. C++ docstring updated to say *"empty string: Infer access mode from query text"*.

For Python, the binding at `tools/python_bind/src/py_connection.cc:90` has always passed `access_mode=""` explicitly, and the old `QueryProcessor::execute` already did `if (requested_mode.empty()) infer via planner->analyzeMode(query)` — Python users have been getting inference all along. **No Python behavior change.**

The Python docstring at `tools/python_bind/src/py_connection.cc:47-52` still says *"...will be set to `update` by default."* — pre-existing inconsistency, worth a follow-up doc fix.

include/neug/main/connection.h
```diff
@@ -83,19 +80,15 @@ class NeugDB;
...
 class Connection {
  public:
-  Connection(GraphSnapshotStore& snapshot_store,
-             std::shared_ptr<QueryProcessor> query_processor)
-      : snapshot_store_(snapshot_store),
-        query_processor_(query_processor),
-        is_closed_(false) {}
-  ~Connection() { Close(); }
+  explicit Connection(std::unique_ptr<ExecutionSlot> execution_slot);
+  ~Connection();
```
longbinlai, line 89: **New lifetime contract is implicit and should be documented on the class.**

Before, `Connection` held `GraphSnapshotStore&` + `shared_ptr<QueryProcessor>` — borrowed refs into NeugDB. Now it owns a `unique_ptr<ExecutionSlot>`. The diff added a sentence (*"Each connection exclusively owns one execution slot"*) but the *destruction order* is the real new constraint:

`ExecutionSlot` holds raw pointers to `NeugDB`-owned resources (snapshot store, version manager, planner, allocator, WAL writer) that are destroyed in `NeugDB::Close()` / `~NeugDB()`. So **a `Connection` MUST be destroyed before the `NeugDB` that created it.** This is a real constraint (UB otherwise), and `Connection::~Connection()` does not enforce it. Worth either:
- Adding a `NeugDB*` back-pointer to `Connection` and asserting in `~Connection()`, or
- Documenting the constraint explicitly on the class.

src/main/neug_db.cc
```diff
@@ -264,12 +253,19 @@ void NeugDB::RegisterService(NeugDBService* svc) {
         "one service instance is allowed per database.");
   }
   active_service_ = svc;
+
+  try {
+    closeAllConnections();
+  } catch (...) {
```
longbinlai, line 269: **The defensive try/catch is good, but the `"none may be in use"` claim in `service.md:64` is misleading.**

`registerService()` now wraps `closeAllConnections()` in a try/catch that resets `active_service_` and re-throws. This is a *correctness* improvement, but `closeAllConnections() -> ConnectionManager::Close() -> Connection::~Connection()` does not actually reject in-use connections — it just calls `Close()` and lets the user find out via the next `Query()` returning an error. The doc still says *"Construction closes existing embedded connections; none may be in use"* (`service.md:64`), which was always aspirational.

Two options:
1. Soften the doc to say *"Connections are closed when the service starts; in-flight queries will fail."*
2. Make `registerService` actually check (e.g. refuse if `connection_count() > 0`) so the doc claim becomes true.

Also, `py_database.cc:serve` calls `database->PrepareForServing()` which also calls `closeAllConnections()` — same doc/behavior mismatch applies to Python users who already have open connections when they call `db.serve(...)`.

include/neug/main/execution_slot.h
```diff
@@ -0,0 +1,277 @@
...
+
+  GraphSnapshotStore& snapshot_store_;
+  std::shared_ptr<IGraphPlanner> planner_;
+  execution::LocalQueryCache pipeline_cache_;
+  IVersionManager& version_manager_;
+  Allocator& alloc_;
+  // Null for AP slots. TpExecutionSlotPool binds this before exposing a TP
+  // slot to callers.
+  IWalWriter* wal_writer_{nullptr};
+  const NeugDBConfig& db_config_;
```
longbinlai, line 272: **`wal_writer_` is the *sole* TP-vs-AP distinguisher, with no runtime enforcement.**

The difference between a TP slot and an AP slot is one nullable raw pointer. Concerns:

1. **No runtime enforcement.** Nothing in `ExecutionSlot`'s read/write paths asserts `wal_writer_ != nullptr` for TP-only methods. If `TpExecutionSlotPool` forgets to bind, the call will silently skip WAL or crash deep inside the writer. Consider `CHECK(wal_writer_ != nullptr)` at the top of TP-only methods.

2. **Slot count policy is not documented.** `NeugDB` creates `max_thread_num` slots. The consequence (*"a request on an N-thread service will block if all slots are checked out"*) should be called out in the class doc or in `service.md`.

3. **Tagged type instead of member state.** A small `enum class SlotKind { AP, TP }` or two `static` factory methods (`createApSlot`, `createTpSlot`) would make the distinction a compile-time invariant. Optional.

longbinlai (commented):
## Consolidated review — supersedes my earlier attempts

I posted several noisy reviews earlier (IDs 4786452202 / 4786478140 / 4786486965 / 4786501402 / 4786554675 / 4786565544 / 4786586032 / 4786584983) chasing diff positions. **Please dismiss those earlier reviews** — they are duplicates and only the `Behavior change for C++ direct users` thread at `connection.h:84` is still anchored.

The four concerns below are anchored to the **current head `d2c091d3`** with verified `position` values.

src/main/execution_slot.cc
zhanglei1949: Removed.

src/main/execution_slot.cc
zhanglei1949: User don't need to configure it, it will be modified by `prepareQuery`, by parsing the executionFlags.

include/neug/main/execution_slot.h
zhanglei1949: Removed.

src/main/execution_slot.cc
zhanglei1949: 还是需要的，这里整合了一下：
```
const bool database_read_only = db_config_.mode == DBMode::READ_ONLY;
const bool plan_read_only = IsReadOnlyExecutionFlag(flags);

if ((database_read_only && mode != AccessMode::kRead) ||
    ((database_read_only || mode == AccessMode::kRead) && !plan_read_only)) {
  RETURN_ERROR(Status(
      StatusCode::ERR_INVALID_ARGUMENT,
      database_read_only
          ? "Database is in read-only mode; write operations are not allowed."
          : "Write queries are not supported in read-only mode"));
}
```

include/neug/execution/common/params_map.h
zhanglei1949: 定义放在`params_map.h`,实现放在`params_map.cc`里

src/main/execution_slot.cc
zhanglei1949:
```c++
result<QueryResult> ExecutionSlot::ExecuteQuery(
    const std::string& query_string, const std::string& access_mode,
    const execution::ParamsMap& parameters, int32_t num_threads) {
  return executeQueryInternal(
      query_string, access_mode,
      [&parameters](const execution::ParamsMetaMap&)
          -> result<execution::ParamsMap> { return parameters; },
      num_threads);
}

result<QueryResult> ExecutionSlot::ExecuteQuery(
    const std::string& query_string, const std::string& access_mode,
    const rapidjson::Value& parameters_json, int32_t num_threads) {
  return executeQueryInternal(
      query_string, access_mode,
      [&parameters_json](const execution::ParamsMetaMap& parameter_types)
          -> result<execution::ParamsMap> {
        return execution::parseJsonParameters(parameter_types, parameters_json);
      },
      num_threads);
}
```

include/neug/execution/common/params_map.h
```diff
@@ -18,12 +18,30 @@
...
+    const ParamsMetaMap& parameter_types, const rapidjson::Value& parameters) {
```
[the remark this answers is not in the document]
zhanglei1949: resovled.

include/neug/main/connection.h
zhanglei1949: resolved.

tools/python_bind/tests/test_tp_service.py
```diff
@@ -511,11 +511,11 @@ def test_parameterized_query(tmp_path):
 
     session.execute(
         "CREATE (p:param_values {"
-        "id: 2, bool_prop: true, date_prop: date('2024-01-01'), "
-        "timestamp_prop: Timestamp('2024-01-02 03:04:05'), "
-        "int32_prop: 42, int64_prop: 1234567890123, "
-        "uint32_prop: 123, uint64_prop: 456, float_prop: 3.14, "
-        "double_prop: 7.28, string_prop: 'parameterized'"
+        "id: 2, bool_prop: false, date_prop: date('2024-02-01'), "
```
liulx20, line 514: 这个case要改吗

longbinlai (dismissed):
LGTM

tools/python_bind/tests/test_tp_service.py
zhanglei1949: 这里其实是之前的测试case有问题。

1. 测试先创建顶点 `id=1`，并定义参数查询，例如 `bool_prop = true`、`int32_prop = 42`，均断言只返回一行。
2. TP 阶段又创建 `id=2`，但除 `double_prop` 外，几乎所有属性都和 `id=1` 相同。
3. 服务停止后，测试重新以 AP `Connection` 执行同一组参数查询，仍断言 `len(res) == 1`。
4. 正确的 AP/TP 语义下，TP 已提交写入必须可见，因此这些查询应命中两行；但是之前的测试代码assert只有一行。

之所以是这样，是因为之前AP查询路径把可见行时间硬编码为0，但是实际上应该使用version_manager获取read_ts。
https://github.com/alibaba/neug/blob/576f99a7b973322aacefde062f33cce45c6a0e79/src/main/query_processor.cc#L123

liulx20 (approved):

PR #789, opened 2026-07-28
fix: use aggregate initialization for search results

shirly121:
## Summary

Fix a macOS libc++ compilation error introduced by #742. `SearchResult` is an aggregate without a matching two-argument constructor, so constructing it through `vector::emplace_back(vid, score)` fails on the affected toolchain. Use explicit aggregate initialization before insertion instead.

Fixes the issue introduced by https://github.com/alibaba/neug/pull/742.

## Testing

- `ap_index_test`
- `tp_index_test`


zhanglei1949 (approved):
LGTM

PR #801, opened 2026-07-29
refactor: unify AP and TP query execution paths

### LINKED ISSUE #800: Unify AP and TP query execution paths in ExecutionSlot
## Background

The P2 ownership work in #783 moved AP and TP execution onto a common `ExecutionSlot` foundation, but the two entry paths can still diverge in query preparation, plan validation, parameter binding, pipeline execution, result handling, and metrics.

This follow-up should merge those paths while preserving the concurrency semantics that are intentionally different between AP and TP.

## Goals

- Use one shared execution core for AP embedded queries and TP service requests.
- Keep query preparation, plan validation, parameter conversion, physical pipeline execution, result sinking, and execution metrics in the shared path.
- Keep AP/TP differences limited to storage and concurrency lifecycle boundaries.
- Prevent future behavior drift without adding parallel AP/TP abstractions or unnecessary wrapper types.

## Design constraints

- `ExecutionSlot` has an immutable embedded or transactional mode.
- AP acquires timestamps directly through the version manager and does not use a WAL writer.
- TP executes through `ReadTransaction`, `InsertTransaction`, or `UpdateTransaction`, and commits through the transaction lifecycle.
- TP-only resources, including WAL writers, remain owned by the TP execution-slot pool.
- Unsupported TP operations such as batch execution or temporary-table creation are rejected from compiled plan metadata.
- Public query parameters use one JSON object representation; the typed `execution::ParamsMap` remains internal to execution.
- Avoid introducing checkpoint-related changes in this work.

## Acceptance criteria

- AP and TP both pass through the same prepare/validate/bind/execute/result pipeline.
- AP read and write queries retain direct version-manager timestamp leases.
- TP read, insert, update, and schema queries retain explicit transaction and commit semantics.
- TP insert mode rejects plans that are not insert-only.
- AP-supported but TP-unsupported plan capabilities return clear errors.
- JSON parameters are validated as objects and converted once into the typed internal representation.
- Parameter references in vertex and edge `SET` expressions are included in compiled parameter metadata.
- Existing AP and TP query behavior is covered by regression tests.
- No checkpoint-specific code is introduced by this change.

## Related work

- Predecessor ownership consolidation: #783
- Broader transaction-model tracking: #584

zhanglei1949:
## Summary

- introduce an immutable embedded/transactional mode on `ExecutionSlot` and make mode-specific resources explicit at construction time
- route AP embedded queries and TP service requests through one shared prepare, validate, parameter-bind, pipeline-execute, and result path
- preserve the intended concurrency boundary: AP uses version-manager timestamp leases without WAL, while TP uses `ReadTransaction`, `InsertTransaction`, or `UpdateTransaction` and commits through the transaction lifecycle
- keep per-slot WAL ownership in `TpExecutionSlotPool`, rely on ordinary `unique_ptr` ownership, and guarantee that each `ExecutionSlot` is destroyed before its WAL writer
- make concrete WAL writer destructors noexcept RAII boundaries while retaining explicit `close()` error reporting
- consolidate the public C++ query interface on RapidJSON object parameters while retaining typed `execution::ParamsMap` as the internal execution representation
- reject non-object and unexpected JSON parameters, and collect parameter metadata from vertex and edge `SET` expressions

## Motivation

P2 consolidated execution ownership, but AP and TP still had separate orchestration for query preparation and execution. Those paths duplicated capability checks, parameter conversion, pipeline invocation, result handling, metrics, and error propagation, which allowed their behavior to drift.

This change keeps one execution core and limits AP/TP branching to the storage and concurrency lifecycle that is genuinely different between the two modes.

## Transactional behavior

- TP read queries acquire and commit a `ReadTransaction`
- TP insert queries acquire and commit an `InsertTransaction` and reject plans that are not insert-only
- TP update and schema queries acquire and commit an `UpdateTransaction`
- TP continues to reject batch and temporary-table plans
- TP writes do not clear the embedded global query cache path
- no checkpoint-specific behavior is changed

## API impact

`Connection::Query` now has one parameter representation:

```cpp
result<QueryResult> Query(
    const std::string& query,
    const std::string& access_mode = "",
    const rapidjson::Value& parameters =
        rapidjson::Value{rapidjson::kObjectType});
```

Python and Node bindings continue to construct JSON objects at the boundary.

## Validation

- `cmake --build build --target neug test_connection test_db_svc -j8`
- `cmake --build build --target neug_py_bind neug_node_bind -j8`
- `cmake --build build --target test_request test_node_create -j8`
- `cmake --build build --target neug test_db_svc tp_index_test -j8`
- `ctest --test-dir build -R '^(test_connection|test_db_svc|test_request|QueryRequestSerializerTypedTest/.*\.RoundTripParameters)$' --output-on-failure`
- `ctest --test-dir build -R '^(tp_index_test|test_db_svc)$' --output-on-failure`
- `git diff --check upstream/main...HEAD`

All selected builds and tests pass.

Closes #800

### DIFF
--- a/include/neug/server/tp_execution_slot_pool.h
+++ b/include/neug/server/tp_execution_slot_pool.h
@@ -69,7 +69,7 @@ class TpExecutionSlotPool {
           logger(std::move(in_logger)),
           slot(snapshot_store, std::move(planner),
                std::move(global_query_cache), version_manager, *allocator,
-               ExecutionSlotMode::kTransactional, logger.get(), config,
+               QueryExecutionStrategy::kTransactional, logger.get(), config,
                slot_id) {
       CHECK(logger != nullptr);
       logger->open();
--- a/include/neug/server/tp_execution_slot_pool.h
+++ b/include/neug/server/tp_execution_slot_pool.h
@@ -17,7 +17,6 @@
 #include <glog/logging.h>
 
 #include <cstdlib>
-#include <exception>
 #include <memory>
 #include <new>
 #include <string>
@@ -57,22 +56,6 @@ class NeugDBService;
  * @since v0.1.0
  */
 class TpExecutionSlotPool {
-  struct WalWriterDeleter {
-    void operator()(IWalWriter* writer) const noexcept {
-      if (writer == nullptr) {
-        return;
-      }
-      try {
-        writer->close();
-      } catch (const std::exception& e) {
-        LOG(WARNING) << "Failed to close slot WAL writer: " << e.what();
-      } catch (...) { LOG(WARNING) << "Failed to close slot WAL writer"; }
-      delete writer;
-    }
-  };
-
-  using WalWriterPtr = std::unique_ptr<IWalWriter, WalWriterDeleter>;
-
   // TP-only per-slot record. Implementation detail of the pool; external code
   // only ever sees ExecutionSlotLease.
   struct alignas(4096) Entry {
@@ -83,7 +66,7 @@ class TpExecutionSlotPool {
           int slot_id, std::unique_ptr<IWalWriter> in_logger,
           const NeugDBConfig& config)
         : allocator(std::move(alloc)),
-          logger(in_logger.release()),
+          logger(std::move(in_logger)),
           slot(snapshot_store, std::move(planner),
                std::move(global_query_cache), version_manager, *allocator,
                ExecutionSlotMode::kTransactional, logger.get(), config,
@@ -94,8 +77,10 @@ class TpExecutionSlotPool {
 
     std::shared_ptr<Allocator> allocator;
     char _padding0[128 - sizeof(std::shared_ptr<Allocator>)];
-    WalWriterPtr logger;
-    char _padding1[4096 - sizeof(WalWriterPtr) -
+    // Declaration order is intentional: slot is destroyed before the WAL
+    // writer it references.
+    std::unique_ptr<IWalWriter> logger;
+    char _padding1[4096 - sizeof(std::unique_ptr<IWalWriter>) -
                    sizeof(std::shared_ptr<Allocator>) - sizeof(_padding0)];
     ExecutionSlot slot;
     char _padding2[(4096 - sizeof(ExecutionSlot) % 4096) % 4096];
--- a/include/neug/server/tp_execution_slot_pool.h
+++ b/include/neug/server/tp_execution_slot_pool.h
@@ -57,6 +57,22 @@ class NeugDBService;
  * @since v0.1.0
  */
 class TpExecutionSlotPool {
+  struct WalWriterDeleter {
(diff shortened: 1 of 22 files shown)


include/neug/server/tp_execution_slot_pool.h
```diff
@@ -57,6 +57,22 @@ class NeugDBService;
  * @since v0.1.0
  */
 class TpExecutionSlotPool {
+  struct WalWriterDeleter {
```
liulx20: Can this be implemented as the `WalWriter` destructor?

src/main/execution_slot.cc
```diff
@@ -134,26 +133,17 @@ class TimestampLease {
...
   return flags.schema() || flags.create_temp_table() || flags.batch() ||
          flags.insert() || flags.update();
 }
 
-template <typename Storage>
 Status executePreparedQuery(execution::CacheValue& prepared_query,
                             const execution::ParamsMap& parameters,
-                            Storage& storage, neug::QueryResponse& response) {
+                            IStorageInterface& storage,
+                            StorageReadInterface* result_storage,
```
liulx20: do we need this additional parameter?

include/neug/server/tp_execution_slot_pool.h
zhanglei1949: Moved to `LocalWalWriter`'s desctructor.

src/main/execution_slot.cc
zhanglei1949: Not needed, removed. Switch back to template function.

liulx20: Can we pass IStorageInterface directly here? It seems unnecessary to use a template.

src/execution/common/params_map.cc
```diff
@@ -24,7 +24,8 @@ result<ParamsMap> parseJsonParameters(const ParamsMetaMap& parameter_types,
                                       const rapidjson::Value& parameters) {
```
liulx20, line 24: ParamsParser::ParseFromJsonObj 现在基本是死代码，只剩 tests/main/test_query_request.cc 还在调；跟活跃执行路径无关。

liulx20: 感觉可以统一这两个函数，可以静默忽略多余参数

src/main/execution_slot.cc
zhanglei1949: Changed to `IStorageInterface` and use `storage.readable()` to determine whether or not to sink.

src/execution/common/params_map.cc
zhanglei1949: Remove ParamsParser::ParseFromJsonObj

liulx20 (dismissed):

liulx20 (approved):

Issue #794, opened 2026-07-28
VersionManager quiescence protocol is not linearizable across independent atomics


zhanglei1949:
## Problem

`VersionManager` represents transaction admission with independent atomics: a phase/state atomic plus separate active-reader and active-inserter counters.

Reader and inserter admission follow a pre-check / counter increment / post-check sequence. Writers and compaction first change the phase and then inspect or drain the corresponding counter. The operations currently do not participate in one linearization order across all of these atomics.

The C++ memory model guarantees a modification order for each atomic object, but acquire/release operations on different atomic objects do not establish a single total order. Making only the phase transition sequentially consistent does not include non-`seq_cst` counter operations in that total order.

A permitted reader interleaving is:

1. R reads the phase as unblocked.
2. W changes the phase to readers-blocked.
3. W reads the independent reader counter as zero and proceeds.
4. R increments the reader counter.
5. R post-checks the phase but is allowed to observe the earlier unblocked value and proceeds.

The analogous inserter interleaving can let an update observe zero active inserters while a delayed inserter subsequently completes admission. This can violate the assumptions used when cloning or compacting storage.

This is not a C++ data race because the fields are atomic; it is an algorithmic correctness issue caused by the lack of a cross-atomic linearization point. It may be difficult to reproduce on strongly ordered machines, but the implementation must also be correct for weaker architectures and under legal compiler transformations.

## Impact

The protocol cannot formally guarantee that:

- blocking new readers and observing zero readers establishes reader quiescence;
- entering update execution and observing zero inserters excludes a delayed inserter;
- compaction starts only after every admitted reader and inserter has drained;
- lifecycle operations that require a quiescent manager observe a stable quiescent state.

Stress tests and ThreadSanitizer are not sufficient to prove these properties because all accesses are atomic and the problematic execution is timing- and memory-model-dependent.

## Recommended solution

Replace the independently coordinated phase and counters with a single atomic control word, for example a `std::atomic<uint64_t>` encoding:

- transaction phase;
- active reader count;
- active inserter count;
- any generation bits required by the admission protocol.

All admission and phase transitions should use CAS on that control word:

- A reader increments the reader field only when the phase permits reads.
- An inserter increments the inserter field only while the phase is idle.
- An update or compaction changes the phase in the same modification order as all admissions.
- Once the phase transition succeeds, forbidden transaction types cannot increment their counters.
- Reader and inserter release decrement their fields in the same control word.
- Drain loops reload that control word until the required fields reach zero.

[...]

Issue #775, opened 2026-07-23
Query cache can be repopulated with stale plans after schema changes


zhanglei1949:
## Problem

GlobalQueryCache is cleared when a schema-changing update transaction commits. However, read transactions that pinned the previous graph snapshot are allowed to continue while the new snapshot is committed.

A reader can therefore race with cache invalidation as follows:

1. A read transaction pins the current snapshot and begins resolving a query against its schema.
2. A concurrent schema-changing update clears the global query cache and publishes a new snapshot.
3. The existing reader compiles the query against its pinned old snapshot and inserts that plan into the now-empty global cache.
4. A reader on the new snapshot looks up the same query and reuses the plan compiled against the old schema.

```mermaid
sequenceDiagram
    participant R as Reader Thread
    participant W as Schema Update Thread
    participant C as GlobalQueryCache
    participant S as Snapshot Store

    R->>S: Pin current snapshot S0
    R->>S: Check S0 is current
    S-->>R: true
    Note over R,W: Reader is paused before accessing the cache

    W->>C: Clear all cached plans
    W->>S: Publish schema snapshot S1
    W->>W: Commit completes

    R->>C: Get query Q
    C-->>R: Cache miss
    R->>R: Compile plan P0 against pinned S0
    R->>C: Insert Q to P0

    Note over C,S: Cache now contains P0 while the current snapshot is S1
```

The reverse mismatch is also possible: an old-snapshot reader may reuse a plan that was compiled and cached for the newly published schema.

Because cache entries are keyed only by query text and do not carry a schema or snapshot version, clearing the cache alone does not prevent plans from crossing schema generations.

## Impact

After a concurrent schema change, a cached physical plan may be executed against a different schema from the one used during compilation. Depending on the schema change, this can produce incorrect property or label resolution, execution-time errors, or allow a query that should fail during planning to bypass schema validation.

PR #803, opened 2026-07-29
fix: publish coherent read views

### LINKED ISSUE #793: Read transactions can pair an old timestamp with a newly published graph snapshot
## Problem

`Session::GetReadTransaction()` acquires a read timestamp before it pins the current `GraphSnapshotStore` slot. An update commit blocks new readers before publishing a replacement snapshot, but it intentionally does not drain readers that have already acquired a timestamp.

This leaves a valid interleaving:

1. Reader R registers as an active reader and obtains timestamp `T_old`.
2. Before R pins a snapshot, writer W blocks new readers.
3. W publishes snapshot `S_new`.
4. R pins the now-current snapshot and receives `S_new`.

R therefore runs with the pair `(T_old, S_new)` instead of a timestamp and snapshot from one coherent publication generation.

This is a protocol gap rather than a weak-memory-only behavior: it remains possible with sequentially consistent atomics because timestamp acquisition and snapshot pinning are separate operations.

## Impact

Timestamp-filtered records may hide some changes newer than `T_old`, but snapshot-level metadata is not uniformly timestamp-versioned. Schema changes are the clearest unsafe case: R can observe schema from `S_new` while executing with `T_old`. More generally, the read transaction no longer satisfies the invariant that its timestamp and pinned graph snapshot describe one coherent view.

## Recommended solution

Introduce a coherent read-acquisition lease keyed by a monotonically increasing snapshot epoch.

Reader acquisition should:

1. Register the reader and capture `{timestamp, epoch}`.
2. Pin the current snapshot.
3. Revalidate, after pinning, that the reader gate is open and the epoch is unchanged.
4. If validation fails, unpin the snapshot, release the reader registration, wait for the blocked phase to finish, and retry.
5. Only construct `ReadTransaction` after validation succeeds.

Update publication should:

1. Atomically switch the reader gate to blocked and advance the epoch before publishing.
2. Publish the replacement snapshot.
3. Advance the visible read timestamp.
4. Reopen the reader gate.

The final validation gives a clear linearization rule:

- If publication starts before validation, the changed epoch forces the reader to retry.
- If publication starts after validation, the reader has already pinned the old snapshot and can safely continue as an existing reader.
- If a complete blocked-to-open cycle occurs during acquisition, the changed epoch prevents an ABA false success.

The epoch transition must share a well-defined linearization point with blocking new readers. A read-lease API is preferable to exposing the current timestamp-only acquisition sequence to callers.

## Simpler alternative

Block new readers and drain every existing reader before `PublishSnapshot()`. This is easier to prove correct but makes every update publication wait for all read transactions and defeats the intended benefit of retaining multiple snapshot slots. It should be treated as a correctness-first fallback, not the preferred design.

## Acceptance criteria

- Add a deterministic test that pauses a reader after timestamp acquisition and before snapshot pinning, publishes a schema-changing snapshot, and verifies that the reader retries rather than returning `(T_old, S_new)`.
- Verify a reader that completes validation before publication continues on its pinned old snapshot.
- Verify a reader spanning an entire blocked-to-open cycle detects the epoch change.
- Preserve snapshot/timestamp release ordering on success, retry, exception, commit, and abort paths.
- Document the coherent timestamp/snapshot invariant and the acquisition linearization point.

zhanglei1949:
## Summary

- Publish the visibility timestamp and snapshot generation as one coherent read-view value.
- Acquire the pair through a move-only `ReadSnapshotLease` that retries if the pinned snapshot slot does not match the published generation.
- Tag copy-on-write snapshots with their reserved generation and preserve snapshot-before-admission release order.
- Add deterministic regressions for generation mismatch, retry, pinned old snapshots, and release ordering.

## Root cause

A reader could previously acquire an old visibility timestamp and then pin a newly published graph snapshot. Each operation was individually valid, but together they did not form one linearizable read view.

## Scope

This PR is intentionally limited to coherent read-view publication. Admission-gate hardening (#794), snapshot/WAL lifecycle hardening, and schema-aware query-cache invalidation (#775) are split into follow-up PRs.

Fixes #793

### DIFF
--- a/include/neug/transaction/read_snapshot_lease.h
+++ b/include/neug/transaction/read_snapshot_lease.h
@@ -15,7 +15,6 @@
 #pragma once
 
 #include <stdint.h>
-#include <utility>
 
 #include "neug/storages/graph_snapshot_store.h"
 #include "neug/transaction/version_manager.h"
@@ -36,41 +35,15 @@ class ReadSnapshotLease {
   static ReadSnapshotLease Acquire(IVersionManager& version_manager,
                                    GraphSnapshotStore& snapshot_store);
 
-  ReadSnapshotLease(ReadSnapshotLease&& other) noexcept
-      : version_manager_(other.version_manager_),
-        published_view_(other.published_view_),
-        active_(other.active_),
-        snapshot_(std::move(other.snapshot_)) {
-    other.version_manager_ = nullptr;
-    other.active_ = false;
-  }
-
-  ReadSnapshotLease& operator=(ReadSnapshotLease&& other) noexcept {
-    if (this != &other) {
-      release();
-      version_manager_ = other.version_manager_;
-      published_view_ = other.published_view_;
-      active_ = other.active_;
-      snapshot_ = std::move(other.snapshot_);
-      other.version_manager_ = nullptr;
-      other.active_ = false;
-    }
-    return *this;
-  }
+  ReadSnapshotLease(ReadSnapshotLease&& other) noexcept;
+  ReadSnapshotLease& operator=(ReadSnapshotLease&& other) noexcept;
 
   ReadSnapshotLease(const ReadSnapshotLease&) = delete;
   ReadSnapshotLease& operator=(const ReadSnapshotLease&) = delete;
 
-  ~ReadSnapshotLease() noexcept { release(); }
+  ~ReadSnapshotLease() noexcept;
 
-  void release() noexcept {
-    if (!active_) {
-      return;
-    }
-    active_ = false;
-    snapshot_.release();
-    version_manager_->release_read_view();
-  }
+  void release() noexcept;
 
   timestamp_t timestamp() const { return published_view_.visibility_ts; }
   uint32_t snapshot_generation() const {
@@ -82,11 +55,7 @@ class ReadSnapshotLease {
 
  private:
   ReadSnapshotLease(IVersionManager& version_manager, SnapshotGuard snapshot,
-                    PublishedReadView published_view) noexcept
-      : version_manager_(&version_manager),
-        published_view_(published_view),
-        active_(true),
-        snapshot_(std::move(snapshot)) {}
+                    PublishedReadView published_view) noexcept;
 
   IVersionManager* version_manager_;
   PublishedReadView published_view_;
--- a/include/neug/transaction/read_snapshot_lease.h
+++ b/include/neug/transaction/read_snapshot_lease.h
@@ -69,7 +69,7 @@ class ReadSnapshotLease {
     }
     active_ = false;
     snapshot_.release();
-    version_manager_->release_read_timestamp();
+    version_manager_->release_read_view();
   }
 
   timestamp_t timestamp() const { return published_view_.visibility_ts; }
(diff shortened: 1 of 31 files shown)


src/transaction/version_manager.cc
```diff
@@ -27,8 +28,16 @@ namespace neug {
...
+  if (ts == std::numeric_limits<uint32_t>::max()) {
+    THROW_RUNTIME_ERROR(
+        "Transaction timestamp space exhausted; checkpoint/reset the timeline "
+        "before reopening the database");
+  }
   write_ts_.store(ts + 1, std::memory_order_relaxed);
   read_ts_.store(ts, std::memory_order_relaxed);
+  installed_snapshot_generation_.store(0, std::memory_order_relaxed);
+  published_read_view_.store(PackPublishedReadView({ts, 0}),
+                             std::memory_order_relaxed);
```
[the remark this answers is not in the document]
zhanglei1949: resolved.

include/neug/transaction/read_snapshot_lease.h
```diff
@@ -0,0 +1,67 @@
...
+  ~ReadSnapshotLease() noexcept;
+
+  void release() noexcept;
+
+  timestamp_t timestamp() const { return published_view_.visibility_ts; }
+  uint32_t snapshot_generation() const {
+    return published_view_.snapshot_generation;
+  }
+  const GraphView& view() const { return snapshot_.get().view(); }
+  const PropertyGraph* graph() const { return snapshot_.get().mutable_graph(); }
```
liulx20: 为什么这还需要暴露PropertyGraph

include/neug/transaction/read_snapshot_lease.h
```diff
@@ -0,0 +1,67 @@
...
+  bool valid() const { return active_; }
```
liulx20: 这个似乎没有调用，是否需要暴露呢

include/neug/transaction/read_snapshot_lease.h
zhanglei1949: 对于QueryExecutionStrategy::kDirect，还是需要拿到graph。最新commit已改成ReadSnapShotLease暴露snapshot()接口，而不是直接暴露graph()和view().
```c++
  if (execution_strategy_ == QueryExecutionStrategy::kDirect) {
    if (access_mode == AccessMode::kRead) {
      auto lease =
          ReadSnapshotLease::Acquire(version_manager_, snapshot_store_);
      const auto& snapshot = lease.snapshot();
      StorageReadInterface storage(snapshot.view(), lease.timestamp());
      status = execute_on_storage(GraphStats(*snapshot.graph()), storage);
    }
}
```

include/neug/transaction/read_snapshot_lease.h
zhanglei1949: valid()/valid_已去掉，保留
```c++
  timestamp_t timestamp() const { return timestamp_; }
  const GraphSnapshotStore::SnapshotSlot& snapshot() const {
    return snapshot_.get();
  }
```

liulx20 (approved):

PR #812, opened 2026-07-30
fix: linearize transaction admission

### LINKED ISSUE #794: VersionManager quiescence protocol is not linearizable across independent atomics

zhanglei1949:
## Summary

- Replace the independent admission phase, reader count, and inserter count atomics with one packed 64-bit operation gate.
- Linearize reader/inserter admission and update/compact phase transitions through CAS on the same control word.
- Preserve adaptive backoff on contended paths and exact quiescence checks for runtime-wait lifecycle changes.
- Add deterministic delayed-admission tests plus packed-counter overflow coverage.

## Why

With independent atomics, a delayed reader or inserter could legally observe an old phase while a writer observed a zero counter. A single control-word modification order makes every operation either admitted and counted before the transition or rejected and retried after it.

Fixes #794

### DIFF
--- a/include/neug/transaction/insert_transaction.h
+++ b/include/neug/transaction/insert_transaction.h
@@ -60,9 +60,9 @@ class Schema;
  * is a correctness bug, not a performance optimization.
  *
  * **Concurrency contract** (VersionManager state machine):
- * - Insert requires admission_state_==kOpen; multiple concurrent inserts are
- *   allowed (active_inserters_ counter). Update/compact transitions the state
- *   away from kOpen, blocking new inserts.
+ * - Insert requires the packed operation gate phase to be kOpen; multiple
+ *   concurrent inserts are tracked by its active-inserter field.
+ *   Update/compact transitions the phase away from kOpen, blocking new inserts.
  * - Insert does NOT block readers, and readers do NOT block insert.
  * - The pinned slot's PropertyGraph is shared with all readers on that slot.
  *
--- a/include/neug/transaction/update_transaction.h
+++ b/include/neug/transaction/update_transaction.h
@@ -83,7 +83,7 @@ class UpdateTransaction {
    * @param logger Reference to WAL writer
    * @param snapshot_store Reference to GraphSnapshotStore for commit
    * @param cache Reference to query cache
-   * @param timestamp_lease Owner of the transaction timestamp
+   * @param timestamp_lease Owned update timestamp and admission lifecycle
    *
    * @note The caller is responsible for acquiring the timestamp lease before
    * creating the COW copy via Clone().
--- a/include/neug/transaction/version_manager.h
+++ b/include/neug/transaction/version_manager.h
@@ -98,6 +98,82 @@ class IVersionManager {
   virtual RuntimeWaitFn runtime_wait_impl() const noexcept = 0;
 };
 
+namespace detail {
+
+enum class AdmissionState : uint8_t { kOpen, kInsertsBlocked, kAllBlocked };
+
+// Layout: [63:62] phase | [61:31] active readers | [30:0] active inserters.
+struct OperationGateWord {
+  static constexpr uint32_t kMaxCount = 0x7fffffffu;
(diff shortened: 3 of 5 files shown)


liulx20 (approved):

zhanglei1949:
------------------------------------------------------------------------------
Operation Count:                        609,981,941
Duration:                               01:53:49.381.000 (h:m:s.ms.us)
Throughput:                             89,317.31 (op/s)
Start Time (China Standard Time):       2026-08-03 - 16:06:46.683
Finish Time (China Standard Time):      2026-08-03 - 18:00:36.064
------------------------------------------------------------------------------
https://github.com/alibaba/neug/actions/runs/30792042796/job/91617446673

PR #814, opened 2026-07-30
fix: isolate query plans by schema generation

### LINKED ISSUE #775: Query cache can be repopulated with stale plans after schema changes

zhanglei1949:
## Summary

- tag each published graph snapshot with a schema generation that remains stable while readers keep that snapshot pinned
- key global and execution-slot query caches by `{schema_generation, query}` instead of query text alone
- advance the generation only for schema changes, while ordinary DML keeps the same generation
- add a regression that lets an old-snapshot reader repopulate the cache after DDL and verifies a new reader cannot reuse that plan

Cache clearing remains useful for reclamation, but cache correctness no longer depends on clear timing. This PR is independently based on `main`; it intentionally excludes the coherent read-view protocol, transaction admission gate, and snapshot/WAL preparation changes split from #803.

Fixes #775

## Validation

- `cmake --build build --target test_db_svc graph_snapshot_store_test tp_index_test transaction_test -j4`
- `MODERN_GRAPH_DATA_DIR=$PWD/example_dataset/modern_graph build/tests/unittest/test_db_svc --gtest_filter=NeugDBServiceTest.QueryCacheSeparatesSchemaGenerations` (passed)
- `MODERN_GRAPH_DATA_DIR=$PWD/example_dataset/modern_graph ctest --test-dir build -R '^test_db_svc$' --output-on-failure` (34/34 passed; run outside the sandbox for local-port tests)\n- `ctest --test-dir build -R 'graph_snapshot_store_test|tp_index_test|transaction_test' --output-on-failure` (3/3 passed)


include/neug/execution/execute/query_cache.h
```diff
@@ -89,30 +103,38 @@ class GlobalQueryCache {
...
   void clear() {
     std::unique_lock<std::shared_mutex> write_lock(mutex_);
-    version_.fetch_add(1);
+    cache_epoch_.fetch_add(1);
     cache_.clear();
   }
 
  private:
-  GlobalQueryCache() : version_(0) {}
+  using GenerationCache =
```
liulx20: 感觉有点复杂，有没有可能比如cache维护一个schema版本号，然后schema也带一个版本号，不一致就直接调compiler，此时plan直接返回，也不进cache

zhanglei1949:
```
------------------------------------------------------------------------------
Operation Count:                        610,035,132
Duration:                               01:53:49.368.000 (h:m:s.ms.us)
Throughput:                             89,325.27 (op/s)
Start Time (China Standard Time):       2026-08-03 - 20:34:50.212
Finish Time (China Standard Time):      2026-08-03 - 22:28:39.580
------------------------------------------------------------------------------
```
https://github.com/alibaba/neug/actions/runs/30809603280/job/91672945910

src/main/execution_slot.cc
```diff
@@ -147,9 +160,9 @@ InsertTransaction ExecutionSlot::GetInsertTransaction() {
 
 UpdateTransaction ExecutionSlot::GetUpdateTransaction() {
   UpdateTimestampLease timestamp_lease(version_manager_);
-  auto cow_graph = snapshot_store_.CurrentSnapshot().Clone();
-  return UpdateTransaction(std::move(cow_graph), alloc_, *wal_writer_,
-                           snapshot_store_, pipeline_cache_,
+  auto [cow_graph, schema_generation] = snapshot_store_.CloneCurrentForUpdate();
+  return UpdateTransaction(std::move(cow_graph), schema_generation, alloc_,
+                           *wal_writer_, snapshot_store_, pipeline_cache_,
                            std::move(timestamp_lease));
```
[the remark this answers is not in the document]
zhanglei1949: UpdateTimestampLease will block any writers and drain writers when it is constructed.

include/neug/execution/execute/query_cache.h
zhanglei1949: resolved.

liulx20 (dismissed):

liulx20 (dismissed):

Issue #816, opened 2026-07-31
code-graph-wiki-pipeline tutorial: Leiden community count drifts and reproduces #807 on released v0.1.3


longbinlai:
## Summary

End-to-end run of `doc/source/tutorials/code-graph-wiki-pipeline.md` against the released `neug 0.1.3` (PyPI) and `gds v0.1.3` extension turns up **two distinct problems**, both reproducible on a clean `tempfile.mkdtemp()` DB with no special config.

Parent context: the v0.2 Tutorial deliverable under **#461**.

---

## Finding 1 — Tutorial drift: Step 4 community count is 91, not 30

**Expected** (tutorial): *"30 communities"*, *"community 0 size 191"*.
**Actual** with the literal query from the tutorial (`CALL leiden('code_graph', {concurrency: 1})`):
- 91 distinct communities (not 30)
- community 0 size 291 (not 191)
- 15 singleton communities

I tried to recover 30 by tuning the algorithm:
- `resolution` sweep over `{0.1, 0.3, 0.5, 0.7, 1.0, 2.0, 5.0}` → no value gives 30 (closest is 74 at `resolution=2.0`)
- `threshold` sweep over `{1e-3, 1e-5, 1e-7, 1e-9, 1e-12}` → all return 91 (threshold is the convergence criterion, not a resolution knob)
- Reinstalling the `gds` extension from OSS → still 91. OSS only ships one `gds v0.1.3` binary, so this confirms the binary is identical to the cached one.

The graph data is correct: `MATCH (f:file) RETURN count(f)` returns **1332** and `MATCH ()-[d:depends]->() RETURN count(d)` returns **2982** — both match the tutorial exactly. So the drift is not data.

Notably, **Step 5 global PageRank top-5 matches the tutorial byte-for-byte** (`binder.h 0.015460, operator.h 0.015152, database.py 0.011829, string_format.h 0.010160, types.h 0.009774` — all five files, ranks, and ordering identical). PageRank is a deterministic function of the graph, so this is strong evidence the graph and projection are correct. Only Leiden is drifting.

Likely root cause: the default leiden parameters (or the implementation itself) have changed since the tutorial was written — either gds was rebuilt with new defaults, or the published dataset is a different snapshot than the one used when the tutorial numbers were captured. Either way, the tutorial should be updated to (a) cite the actual current community count and (b) explicitly pass any parameter that pins the result, OR the gds extension should be rebuilt to match the prior behavior.

[...]

Issue #807, opened 2026-07-30
mutable_csr edge capacity locks after COPY FROM bulk edge load


BingqingLyu:
## Bug Description

After `CREATE NODE TABLE` + `COPY FROM` (nodes) + `COPY FROM` (edges), the mutable_csr edge storage capacity is frozen at N nodes. Subsequent `CREATE` of new nodes succeeds (count increases), but adding edges involving the new node fails with:

```
Source vertex id out of range: N >= N
```

## Reproduction

```python
import tempfile
from neug import Database

db = Database(db_path=tempfile.mkdtemp(), mode="w")
conn = db.connect()

conn.execute("CREATE NODE TABLE file(id STRING, label STRING, PRIMARY KEY(id))")
conn.execute("CREATE REL TABLE depends(FROM file TO file, weight INT64)")

# Bulk load 3 nodes
ncsv = tempfile.mktemp(suffix=".csv")
with open(ncsv, "w") as f:
    f.write("id,label\na.py,a\nb.py,b\nc.py,c\n")
conn.execute(f"COPY file FROM \x27{ncsv}\x27 (header=true, delim=\x27,\x27)")

# Bulk load 2 edges — this locks CSR capacity at 3
ecsv = tempfile.mktemp(suffix=".csv")
with open(ecsv, "w") as f:
    f.write("src,dst,weight\na.py,b.py,1\nb.py,c.py,1\n")
conn.execute(f"COPY depends FROM \x27{ecsv}\x27 (header=true, delim=\x27,\x27)")

# CREATE a 4th node — succeeds
conn.execute("CREATE (n:file {id: \x27d.py\x27, label: \x27d\x27})")
print("count:", list(conn.execute("MATCH (f:file) RETURN count(f)"))[0][0])  # 4

# Try to add an edge from d.py -> a.py — FAILS
conn.execute("""
    MATCH (a:file {id: \x27d.py\x27}), (b:file {id: \x27a.py\x27})
    CREATE (a)-[:depends {weight: 1}]->(b)
""")
# RuntimeError: Source vertex id out of range: 3 >= 3
```

## Key Observations

1. **COPY FROM nodes only → no bug**: loading 1332 nodes without edges, then CREATE + edge works fine.
2. **COPY FROM nodes + edges → bug**: after bulk edge load, CSR capacity is frozen.
3. **Node count is not the trigger**: 3 nodes + 2 edges fails; 100 nodes + 0 edges works; 1332 nodes + 2982 edges fails.
[...]

PR #819, opened 2026-08-03
docs: use COPY append for incremental updates in code wiki tutorial

BingqingLyu:
## Related Issues
#807 #816 

## What does this PR do?
Fixes the incremental-update section of the "Building a Code Wiki with Graph" tutorial so the full runnable script works end-to-end on v0.1.3.

## What changes in this PR?
- Step 6.1 and the full runnable script: replace per-edge `MATCH ... CREATE (a)-[:depends]->(b)` with `COPY depends FROM ...` appending to the existing edge table. On v0.1.3, creating an edge right after inserting new vertices fails with `Source vertex id out of range` (`mutable_csr.h` `put_edge`).
- New nodes are also appended via `COPY file FROM ...` (CSV carrying all current columns, including `community`) for a consistent COPY-based incremental workflow.
- Fix the OSS dataset URLs in the intro to match the full runnable script.

### DIFF
--- a/doc/source/tutorials/code-graph-wiki-pipeline.md
+++ b/doc/source/tutorials/code-graph-wiki-pipeline.md
@@ -285,7 +285,7 @@ conn.execute(f"""
     )
 """)
 
-# Insert edges: each new file is related to existing files. Append them to the
+# Append edges: each new file is related to existing files. Append them to the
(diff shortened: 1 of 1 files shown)


longbinlai (dismissed):
lgtm

longbinlai (approved):

PR #844, opened 2026-08-10
fix: handle LIST/ARRAY columns in C++ QueryResult API

### LINKED ISSUE #607: C++ QueryResult API does not handle kListArray protobuf type for LIST/ARRAY columns
## Description

The `sink.cc` `add_column` function serializes both `kList` and `kArray` execution columns into the `list_array` protobuf field (see `sink.cc` kArray branch at L383-401). However, the C++ `QueryResult` API in `src/main/query_result.cc` does not handle the `kListArray` case in its `get_value`, `IsNull`, and `GetCurrentRowAsString` functions.

## Impact

When using the C++ embedded API to read query results that contain LIST or ARRAY columns, the values are returned as null/empty. The Python bindings (`py_query_result.cc`) already handle `list_array` correctly, so this only affects C++ API users.

## Location

- `src/main/query_result.cc` — `get_value()`, `IsNull()`, `GetCurrentRowAsString()` switch statements lack a `kListArray` case
- `src/execution/common/operators/retrieve/sink.cc` — `add_column()` serializes both kList and kArray to `list_array`

## Suggested Fix

Add `kListArray` case to the switch statements in `query_result.cc`, recursively calling `get_value` to decode the nested elements and offsets.

## Context

Discovered during code review of PR #507 (Array column support). This is a pre-existing issue that also affects LIST columns, not just the new ARRAY type.

liulx20:
Decode kListArray (and kStructArray) in get_value/IsNull so LIST and ARRAY result columns are not silently returned as null.

Fixes #607

### DIFF
--- a/tests/unittest/test_query_result.cc
+++ b/tests/unittest/test_query_result.cc
@@ -394,4 +394,52 @@ TEST(QueryResultTest, GetCurrentRowAsStringWithNull) {
   EXPECT_EQ(result.GetCurrentRowAsString(), "null");
 }
 
+// Build a result with a LIST column (serialized as list_array protobuf).
+// Row 0: [1, 2], row 1: NULL, row 2: [3]
+static QueryResult BuildListResult() {
+  neug::QueryResponse response;
+  response.set_row_count(3);
+
+  auto* schema = response.mutable_schema();
+  schema->add_name("nums");
+
+  auto* col = response.add_arrays();
+  auto* list_arr = col->mutable_list_array();
+  // offsets: [0, 2, 2, 3] → lengths 2, 0, 1
+  list_arr->add_offsets(0);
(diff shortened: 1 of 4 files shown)


tests/unittest/test_query_result.cc
```diff
@@ -394,4 +394,52 @@ TEST(QueryResultTest, GetCurrentRowAsStringWithNull) {
   EXPECT_EQ(result.GetCurrentRowAsString(), "null");
 }
 
+// Build a result with a LIST column (serialized as list_array protobuf).
```
zhanglei1949, line 397: A python test could verify the end-to-end correctness, maybe we should add one?

zhanglei1949 (approved):
LGTM

PR #845, opened 2026-08-10
fix: server crash when reading large ZSTD Parquet files

### LINKED ISSUE #839: [BUG] COPY FROM large ZSTD parquet crashes server: use-after-free in buffered stream page-header read (Arrow 18)
## Describe the bug

`COPY <table> FROM '<file>.parquet'` crashes the server with glibc heap-corruption aborts (`free(): invalid size` / `malloc(): invalid next size (unsorted)`) when importing a large ZSTD-compressed Parquet file through the Parquet extension.

Valgrind shows the root cause is a **use-after-free inside the bundled Arrow 18.0.0 Parquet reader when `BUFFERED_STREAM` is enabled** — the extension's default, see `extension/parquet/include/parquet_options.h`.

During `SerializedPageReader::NextPage()`, the Thrift page-header deserializer:

```text
ThriftDeserializer::DeserializeUnencryptedMessage
  -> TCompactProtocolT::readFieldBegin
```

keeps reading from the `string_view` returned by `BufferedInputStream::Peek()` after the stream's internal buffer has been reallocated or freed. In other words, the peeked view becomes dangling.

Valgrind output:

```text
Invalid read of size 1
  at TCompactProtocolT::readFieldBegin(...)
  by parquet::ThriftDeserializer::DeserializeUnencryptedMessage(...)
  by parquet::SerializedPageReader::NextPage()
  by parquet::ColumnReaderImplBase::HasNextInternal()
  by parquet::internal::TypedRecordReader::ReadRecords(long)
  by parquet::arrow::LeafReader::LoadBatch(long)
  ...

Address 0x12a13664 is 98,404 bytes inside an unallocated block of size 288,192 in arena "client"
```

The crash is deterministic and **not a race** — it still reproduces with `OMP_NUM_THREADS=1`.

A closely related symptom on a slightly smaller file is the clean Arrow error:

```text
Cannot shrink read buffer if buffered data remains
```

thrown at:

```text
arrow/io/buffered.cc:292
```

This confirms the `BufferedInputStream::Peek` / `SetBufferSize` path is entering an inconsistent state.

> Note: plain **pyarrow 18.0.0 reads the identical file successfully** with buffered stream enabled — dataset API, same 1 MB buffer, non-nullable schema.
>
> So the issue appears specific to how the extension configures `ReaderProperties`. In particular, `extension/parquet/src/parquet_options.cc:84` passes `ReadOptions::batch_size`, which is a **row count** with default `1 << 20`, into `ReaderProperties::set_buffer_size`, which expects **bytes**. This mismatch, combined with this file's page layout, appears to trigger the bug.

## To Reproduce

1. Prepare a large Parquet vertex file.
   - Example: `company.parquet`
   - 997,641 rows
   - 218 MB
   - ZSTD-compressed
   - Written by `parquet-cpp-arrow 22.0.0`
   - 44 columns, many nullable strings
2. Load parquet data
```shell
neug > COPY Company FROM './company.parquet';
[1]    35066 segmentation fault  neug-cli open ./test
```

BingqingLyu:
## Related Issues
fix #839

## What does this PR do?

Fixes the server crash (glibc heap corruption / `free(): invalid pointer`) when reading large ZSTD Parquet files with the buffered stream enabled.

Root cause: Arrow 18.0.0's `BufferedInputStream::SetBufferSize()` caps the new buffer size at `bytes_buffered_ + remaining` without accounting for `buffer_pos_`. When `bytes_buffered_ == 0` and `buffer_pos_ > 0`, the buffer is resized too small and the next fill writes past the end of the buffer, corrupting the heap. This was fixed upstream in apache/arrow GH-48311 / PR #48322 (merged for Arrow 20).

This PR backports the upstream fix as a source patch applied at Arrow fetch time.

## What changes in this PR?

- `cmake/BuildArrowAsThirdParty.cmake`: backport the GH-48311 fix to `arrow/io/buffered.cc::SetBufferSize()` as a configure-time source patch, guarded by a sentinel (`NEUG_GH48311_PATCHED`) for idempotency and a `FATAL_ERROR` if the expected Arrow 18.0.0 code shape changes in the future.
- `extension/parquet/tests/parquet_test.cc`: add two regression tests:
  - `TestBufferedStream_PeekAfterExhaustingBuffer`: unit test covering the `buffer_pos_` edge case.
  - `TestIntegration_ReadZstdWithSmallBufferedStream`: reads a ZSTD parquet file with a small (4KB) `batch_size` buffer; without the patch it fails with the exact error from the issue (`Cannot shrink read buffer if buffered data remains`).
- `extension/parquet/include/parquet_options.h` and `doc/source/extensions/load_parquet.md`: clarify that the buffered-stream buffer size in bytes is controlled by the generic `batch_size` option.

## Verification

- Red/green test: with the patch removed, `TestIntegration_ReadZstdWithSmallBufferedStream` fails with the exact issue error; with the patch, all 35 parquet extension tests pass.
- Verified end-to-end with a real-world 44-column parquet file via `COPY FROM` (all values match pyarrow ground truth).


ShunyangLi (commented):
parquet功能已修复

longbinlai (approved):
LGTM

PR #854, opened 2026-08-11
fix: parquet data loading not using multithreading

### LINKED ISSUE #853: [BUG] Parquet Data Loading Does Not Use Multithreading
**Describe the bug**
When loading data from Parquet files, NeuG does not enable multithreading. The loading process appears to use only a single thread, which results in slower data loading performance.

**To Reproduce**

Prepare a Parquet dataset.
Load the dataset using NeuG.
Monitor CPU utilization during data loading.
Observe that only one thread/core is actively used.

**Expected behavior**
Parquet data loading should utilize multiple threads to improve loading performance.

BingqingLyu:
## Related Issues
fix #853

## What does this PR do?
Makes the `parallel` option actually take effect for Parquet data loading (`COPY ... FROM ... (format="parquet", parallel=true)`), so the dataset scanner reads with multiple threads instead of always running single-threaded.

## What changes in this PR?
Two root causes were fixed:

1. **The `parallel` option never reached the dataset scanner.** It was only forwarded to `ArrowReaderProperties::set_use_threads()`, which affects column decoding inside a single scan task but not the scanner itself. `ScanOptions::use_threads` was never set, so scans always ran single-threaded. Now the option is propagated to `ScanOptions::use_threads`, and the scanner is given an `IOContext` with an executor (`arrow::io::default_io_context()`).

2. **A single Parquet file had nothing to parallelize over.** One file maps to one fragment, i.e. one scan task, so a threaded scanner would still process it sequentially. When `use_threads` is on, Parquet fragments are now split by row group (`ParquetFileFragment::SplitByRowGroup`) so each row group becomes an independent scan task.

Changed files:
- `extension/parquet/src/parquet_options.cc`: propagate `parallel` to `ScanOptions::use_threads` and set `ScanOptions::io_context`.
- `extension/parquet/src/arrow_reader.cc`: split Parquet fragments by row group before building the scanner when threading is enabled.


ShunyangLi (commented):
已经能正常开启多线程

longbinlai (commented):
The diagnosis and fix layering are solid: `parallel` previously only reached `ArrowReaderProperties::set_use_threads` (column decode), never `ScanOptions::use_threads`; and the row-group split is the right way to create scan tasks for a single file. The third commit correctly catches the real follow-on bug — libcurl easy handles are not thread-safe, so the shared per-file `curl_handle_` would corrupt the heap once `ReadRange()` is called concurrently; thread-local handles + caller-owned per-request header lists is the right fix, and `curl_global_init` is properly guarded with a static `once_flag`.

Two things I'd like addressed or answered before merge:
1. **Default behavior change**: `parallel` defaults to `true`, so after this PR every parquet COPY splits by row group and scans concurrently — output row order becomes non-deterministic where it was previously stable. Worth documenting.
2. **The httpfs concurrency fix has zero test coverage** (the integration test uses local files). Was it at least verified manually against a real HTTP/OSS endpoint with parallel=true?

Smaller notes inline.

extension/parquet/src/arrow_reader.cc
```diff
@@ -125,6 +128,46 @@ std::shared_ptr<arrow::dataset::Scanner> ArrowReader::createScanner(
...
+              split_any = true;
+              split_fragments.insert(split_fragments.end(),
+                                     rg_fragments->begin(),
+                                     rg_fragments->end());
+              continue;
+            }
+          }
+          split_fragments.push_back(std::move(fragment));
+        }
+        if (split_any) {
```
longbinlai, line 160: Semantic change worth documenting: `parallel` defaults to `true` (options.h), so after this PR every parquet COPY takes this path by default. Batches now arrive in scan-task completion order instead of file order — previously `parallel=true` was effectively single-threaded and order-stable.

COPY FROM generally doesn't promise order, but if anything relies on file order (implicit row numbering, resumable/repeatable loads producing identical output), this changes behavior silently. Suggest a note in the parquet docs that `parallel=true` does not preserve row order.

extension/parquet/src/arrow_reader.cc
```diff
@@ -13,7 +13,10 @@
  * limitations under the License.
  */
 
+#include <arrow/compute/expression.h>
+#include <arrow/dataset/dataset_internal.h>
```
longbinlai, line 19: `FragmentDataset` comes from `dataset_internal.h` — an internal Arrow API. It exists in the bundled Arrow 18, but this is exactly the kind of thing that breaks on an Arrow version bump (cf. the GH-48311 patch in #845). A one-line comment here saying "internal API — recheck when bumping Arrow" would save the next person some time.

extension/parquet/src/arrow_reader.cc
```diff
@@ -125,6 +128,46 @@ std::shared_ptr<arrow::dataset::Scanner> ArrowReader::createScanner(
   }
   auto dataset = dataset_result.ValueOrDie();
 
+  // A single Parquet file corresponds to one fragment, i.e. one scan task,
+  // so a threaded scanner would have nothing to parallelize over. Split
+  // Parquet fragments by row group to create independent scan tasks.
+  if (scan_opts->use_threads) {
+    auto fragments_result = dataset->GetFragments();
+    if (fragments_result.ok()) {
+      auto fragments_vec_result = fragments_result.ValueOrDie().ToVector();
```
longbinlai, line 139: Nit: if `ToVector()` fails, the error is silently swallowed — only the outer `GetFragments()` failure gets a LOG(WARNING). Consider logging this branch too, or restructuring so both failures share the warning path.

extension/httpfs/src/http_filesystem.cc
```diff
@@ -394,37 +402,59 @@ arrow::Result<int64_t> HTTPRandomAccessFile::ReadRange(int64_t offset,
     return 0;
   }
 
+  // Parquet parallel scan may call ReadRange() concurrently from multiple
+  // worker threads.  libcurl easy handles are NOT thread-safe, so each
+  // thread keeps its own handle instead of sharing a per-file handle.
+  // Sharing one handle previously caused heap corruption
+  // ("double free or corruption") under concurrent reads.
+  thread_local CURL* thread_curl = nullptr;
```
longbinlai, line 415: This thread-local handle change is the highest-risk part of the PR but has no test coverage — the integration test reads local files, so the concurrent `ReadRange()` path never runs in CI. Two asks:
1. Was this verified manually against a real HTTP/OSS/S3 endpoint with `parallel=true` (ideally the workload that originally hit "double free or corruption")? If so, noting the verification in the PR description would help.
2. Minor: `thread_curl` is never cleaned up — intentional for connection reuse and bounded by pool size, but if the executor ever destroys worker threads, each dead thread leaks its handle. A comment acknowledging the trade-off would be enough.

extension/parquet/tests/parquet_test.cc
```diff
@@ -740,6 +774,52 @@ TEST_F(ParquetTest, TestIntegration_BatchReadMode) {
...
+  auto reader = createParquetReader(sharedState);
+  auto localState = std::make_shared<reader::ReadLocalState>();
+  execution::Context ctx;
+  reader->read(localState, ctx);
+
+  int64_t totalRows = 0;
+  for (size_t i = 0; i < ctx.chunk_num(); ++i) {
+    totalRows += static_cast<int64_t>(ctx.chunk(i).chunk().row_num());
+  }
+  EXPECT_EQ(totalRows, kNumRows)
```
longbinlai, line 819: Nit: this only asserts the row count. Since splitting changes batch boundaries, verifying at least the aggregate of a value column (e.g. sum of `value`) would catch duplicates/dropped rows more reliably than a count, and spot-checking a few values costs nothing.

longbinlai:
Addressed the review comments directly in 8d8710bb:

1. **Row-order semantics**: documented in `load_data.md` / `import_data.md` that `parallel=true` does not preserve row order. Also fixed the documented default — it is `true` (per `ReadOptions::use_threads`), not `false`.
2. **`dataset_internal.h`**: added a note that `FragmentDataset` comes from internal Arrow API and must be rechecked when bumping the bundled Arrow.
3. **`ToVector()` failure**: now logs a `LOG(WARNING)` instead of being silently swallowed.
4. **thread-local CURL**: added a comment documenting the lifetime trade-off (never cleaned up for connection reuse, bounded by pool size). The manual verification against a real HTTP/OSS endpoint is still for the author to confirm in the PR description.
5. **Test assertion**: `TestIntegration_ParallelReadMultiRowGroup` now also reads in full (non-batch) mode and asserts per-column aggregates (sum of `id`, sum of `value`) so duplicated/dropped rows are caught, not just the count.

These changes were not locally compiled (parquet/httpfs targets are not in my local build config) — relying on CI for the build/test check.

longbinlai:
Follow-up on the `parallel` default (supersedes part of the previous comment): instead of fixing the docs to match the code default of `true`, 61af772b makes parallel Parquet scanning **opt-in**:

- `ReadOptions::use_threads` default changed `true` → `false`. Before this PR the `true` default was inert (ScanOptions::use_threads was never set); activating it unconditionally would silently turn every Parquet load into a parallel, order-unstable scan.
- Docs default reverted to `false`; the row-order note is kept for users who opt in with `parallel=true`.
- `TestOptionsTranslation_DefaultValues` updated to expect `parallel=false`.

longbinlai (approved):
LGTM

PR #861, opened 2026-08-11
WIP: expose generic operation admission leases

zhanglei1949:
## Summary

- Expose shared/exclusive operation-admission RAII leases through IVersionManager.
- Add AP operation guards without reserving TP timestamps.
- Keep TP coherent read acquisition on its dedicated read-view lease.

## Scope

This WIP contains only PR-02A code and tests. Explicit-transaction design documents are intentionally excluded.

## Validation

- Built read_view_publication_test and transaction_test.
- read_view_publication_test passed.
- transaction_test did not return a terminal result in the local runner and remains to be re-run.


zhanglei1949:

Issue #743, opened 2026-07-15
[Tracking] Support running NeuG on Windows


lnfjpt:
Enable native Windows support for NeuG.

This could be achieved through the following adaptation tasks:
* **Build System Adaptation (CMake)**:
  * Update `CMakeLists.txt` to properly detect and support the MSVC compiler and Windows SDK.
  * Fix compiler-specific flags (e.g., replacing `-Wall -Werror` with `/W4 /WX`, handling `/O2` vs `-O3`).
* **OS Abstraction Layer (System Calls & I/O)**:
  * Abstract POSIX-specific memory mapping (e.g., `mmap` / `munmap`) and provide Win32 equivalents (`CreateFileMapping` / `MapViewOfFile`).
  * Replace or wrap Linux-specific file I/O and directory operations with `std::filesystem` or cross-platform wrappers.
  * Adapt threading and synchronization primitives if `pthread` is heavily used (migrate to `std::thread`, `std::mutex`, or Windows `SRWLock`).
* **Dependency Management**:
  * Ensure third-party C++ dependencies are cross-platform.
  * (Optional) Provide `vcpkg` or `conan` integration to easily fetch Windows-compatible binaries for dependencies.
* **CI/CD Integration**:
  * Add a Windows runner (e.g., `windows-latest` with MSVC) to GitHub Actions to continuously verify Windows builds and prevent future regressions.

PR #862, opened 2026-08-11
fix: Fix Index Reopen Issue when extension is unloaded

### LINKED ISSUE #852: HNSW index checkpoint cannot be reopened after process restart
## Description

An HNSW index provided by the `vector_search` extension can be persisted and reopened within the same process, but a database containing that index cannot reliably be reopened in a new process.

The HNSW index is stored as a checkpoint module with `module_type = "hnsw_index"`. During a fresh process startup, checkpoint recovery runs as part of database initialization, before a connection can be created and before the user can execute `LOAD vector_search`. The new process has an empty `ModuleFactory`, so the creator for `hnsw_index` has not been registered when checkpoint validation attempts to restore the module.

The database open then fails with an error similar to:

```text
Checkpoint ... references unknown module type hnsw_index
from module index_item_embedding_hnsw
```

Loading the extension after constructing `Database` cannot help because checkpoint recovery has already failed.

## Steps to reproduce

1. Start a process and create a database.
2. Install and load the vector search extension.
3. Create a node table with a vector property.
4. Create an HNSW index on that property.
5. Checkpoint and close the database.
6. Exit the process.
7. Start a new process and open the same database path.

For example, the database is initially prepared with statements equivalent to:

```cypher
INSTALL vector_search;
LOAD vector_search;

CREATE NODE TABLE Item(
    id STRING PRIMARY KEY,
    embedding FLOAT[4]
);

CREATE INDEX item_embedding_hnsw
ON Item USING HNSW (embedding)
WITH (metric = "ip");
```

The second process fails while constructing the database, before it can run:

```python
db = Database(db_path, mode="w")
conn = db.connect()
conn.execute("LOAD vector_search;")
```

## Actual behavior

- Reopening in the same process succeeds because the extension module registration remains in process memory.
- Reopening after a process restart fails because `hnsw_index` is unknown during checkpoint recovery.
- A missing or not-yet-loaded optional extension may cause the checkpoint to be treated as invalid even though the core graph data is still valid.

## Expected behavior

A database containing persisted HNSW index metadata should remain reopenable after a process restart. The absence of an in-memory extension registration at the beginning of database initialization should not make the core graph checkpoint unusable or cause valid graph data to be discarded.

The behavior should be consistent between same-process reopen and cross-process restart.

## Impact

- Persisted HNSW indexes cannot currently be used reliably across application restarts.
- Applications cannot load `vector_search` early enough through the normal query interface because database open happens first.
- Treating an unavailable extension-owned index as checkpoint corruption risks losing access to otherwise valid core graph data.

## Test coverage

`tools/python_bind/tests/test_hnsw_index.py` contains two relevant cases:

- `test_index_persistence_after_checkpoint_and_reopen`, which reopens in the same process and passes.
- `test_index_persistence_after_process_restart`, which reproduces the fresh-process failure and is currently marked as a strict expected failure.

shirly121:
## Summary

This PR resolves index reopen issue by defering extension-backed index recovery until the required extension is loaded.

- Preserve unopened extension-backed indexes as pending entries in `StorageIndexManager`, including their descriptors and index metadata.
- Activate pending indexes after `LOAD EXTENSION` registers their module factories, then reopen and rebind them to graph property columns.
- Reject create/drop/access conflicts involving pending indexes and prevent checkpoints while pending indexes remain, avoiding metadata or index-data loss.
- Add a `required` flag to `ModuleDescriptor`. Unknown modules explicitly marked as optional no longer prevent checkpoint open.
- Let each concrete index decide whether its persisted module is required.
- Add a cross-process reopen test using the `vec_index` test extension.

## Tests

- `ap_index_test`
- `tp_index_test`
- `e2e_index_test`

Fixes #852


src/storages/index/storage_index_manager.cc
```diff
@@ -24,6 +24,15 @@ namespace neug {
 
 static constexpr const char* kIndexPrefix = "index_";
 
+static Status PendingIndexError(const std::string& name,
+                                const std::string& operation) {
+  return Status(
+      StatusCode::ERR_ILLEGAL_OPERATION,
+      "Cannot " + operation + " pending index '" + name +
+          "' before its module is loaded. If this is an HNSW index, execute "
+          "LOAD vector_search first.");
```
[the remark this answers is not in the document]
shirly121: It's better to give the concrete extension name in error messages.

src/storages/index/storage_index_manager.cc
```diff
@@ -90,6 +102,12 @@ Status StorageIndexManager::DropIndex(const std::string& name) {
 
 neug::result<std::vector<StorageIndex*>> StorageIndexManager::GetIndex(
     label_t label_id, const std::string& property_name) const {
+  for (const auto& [name, pending] : pending_indexes_) {
+    if (pending.meta.schema.label_id == label_id &&
+        pending.meta.schema.property_name == property_name) {
+      return tl::unexpected(PendingIndexError(name, "access"));
```
longbinlai: This error propagates into the WAL-replay path: `UpdateTransaction::IngestWal` → `addVertexIndexData` (src/transaction/update_transaction.cc) calls `GetIndex` per property and fails the replay on a pending index.

That creates a recovery deadlock: checkpoint with a persisted extension index → keep writing with the extension loaded → crash → restart. On reopen the index is pending (extension not loaded yet) and WAL replay runs before any connection can execute `LOAD`, so the DB can never open — a new failure mode that's harder to escape than the original one.

Consider skipping pending indexes during WAL replay and rebuilding index contents at activation time (e.g. via `BulkBuild` after rebind), or at minimum fail with an actionable message.

Also note this error effectively blocks all DML/DDL on the affected table (insert/update, `DROP TABLE` via `dropVertexIndex`, property rename) until the extension is loaded — worth documenting that behavior.

src/storages/index/storage_index_manager.cc
```diff
@@ -80,6 +89,9 @@ neug::result<StorageIndex*> StorageIndexManager::CreateIndex(
 }
 
 Status StorageIndexManager::DropIndex(const std::string& name) {
+  if (pending_indexes_.count(name) > 0) {
+    return PendingIndexError(name, "drop");
```
longbinlai: No escape hatch: if the extension can never be loaded (incompatible version, uninstalled, corrupted download), the pending index can't be removed — `DROP INDEX` is rejected here, and `DROP TABLE` / property rename fail via `GetIndex` on the same table. The table becomes unmanageable.

Consider allowing `DROP INDEX` on a pending entry as a metadata-only removal (nothing to unload on disk since the module was never instantiated).

src/storages/index/storage_index_manager.cc
longbinlai: Core storage shouldn't hardcode a specific extension name — this message will be misleading for any future extension-backed index type. Consider referencing `descriptor.module_type` (or the owning extension name persisted alongside it) instead.

src/storages/graph/property_graph.cc
```diff
@@ -84,6 +85,29 @@ StorageIndexManager& PropertyGraph::mutable_index_manager() {
...
+  for (auto* index : activated.value()) {
+    const auto& meta = index->GetMeta();
+    if (meta.schema.label_id >= vertex_tables_.size() ||
+        !schema_.is_vertex_label_valid(meta.schema.label_id)) {
+      return Status::InternalError("Invalid label for pending index: " +
+                                   meta.name);
+    }
+    auto* column = vertex_tables_[meta.schema.label_id].GetPropertyColumnBase(
+        meta.schema.property_name);
+    RETURN_IF_NOT_OK(index->Rebind(IndexBindContext{column}));
```
longbinlai: If `Rebind` fails here, the index has already been moved from `pending_indexes_` into `indexes_` by `StorageIndexManager::ActivateIndexes`, so it remains registered but unbound — later `GetIndexByName` / `SHOW_INDEXES` will expose a broken index while the pending slot is gone. Consider removing the failed entry from `indexes_` (or rolling back the whole activation) before returning the error.

tests/storage/test_e2e_index.cc
```diff
@@ -15,26 +15,102 @@
...
+  config.checkpoint_on_close = false;
+
+  NeugDB reopened;
+  ASSERT_TRUE(reopened.Open(config));
+  auto connection = reopened.Connect();
+  ASSERT_NE(connection, nullptr);
 
+  auto show_before_load = connection->Query("CALL SHOW_INDEXES() RETURN *;");
+  ASSERT_TRUE(show_before_load) << show_before_load.error().ToString();
+  EXPECT_EQ(show_before_load->response().row_count(), 0);
```
longbinlai: UX concern baked in here: pending indexes are invisible in `SHOW_INDEXES` (0 rows), yet `DROP INDEX` / `CREATE INDEX` with the same name are rejected as conflicts — from a user's perspective the index both doesn't exist and does. Consider surfacing pending indexes with a status (e.g. `pending` / `unloaded`) so the state is discoverable.

src/storages/module/module_broker.cc
```diff
@@ -37,6 +37,9 @@ void ModuleBroker::Open(Checkpoint& checkpoint, const CheckpointManifest& meta,
     }
     auto module = factory.Create(desc.module_type);
     if (!module) {
+      if (!desc.required) {
```
longbinlai, line 42: Minor: silently skipping optional unknown modules makes misconfiguration hard to diagnose. A `LOG(INFO)` here (similar to the "Deferred index" log in `StorageIndexManager::Open`) would help.

longbinlai (commented):
Overall the design is sound: deferring extension-backed index recovery, guarding create/drop/checkpoint conflicts, and the `required` flag (default `true`, backward-compatible with old checkpoints) are the right call, and the fork+exec cross-process reopen test is excellent.

Two things I'd want resolved or explicitly discussed before merge:
1. **Crash recovery deadlock**: `IngestWal` maintains indexes during replay, which now errors on pending indexes — a crash after checkpoint (with post-checkpoint writes to the indexed table) makes the DB permanently unopenable, since WAL replay happens before any `LOAD` can run. See inline.
2. **No escape hatch**: if the extension can never be loaded, the pending index blocks `DROP INDEX`, `DROP TABLE`, and property rename on that table forever. See inline.

Follow-up note: the real `vector_search` extension still needs its own `Dump` override setting `required = false` (as the test `VecIndex` does here) for this fix to take effect end-to-end.
Also heads-up: the fork+execl subprocess test is POSIX-only — will need an alternative once Windows support (#743) lands.

src/main/checkpoint_coordinator.cc
```diff
@@ -148,6 +148,13 @@ Status CheckpointCoordinator::execute(Reason reason) {
             -> Status {
           auto& live_graph = maintenance.MutableCurrentSnapshot();
 
+          if (live_graph.HasPendingIndexes()) {
```
zhanglei1949: This guard also rejects the **shutdown** checkpoint: `NeugDB::createCheckpointOnClose()` → `PublishShutdownCheckpoint()` (src/main/neug_db.cc:586) hits the same path, and `Close()` then throws (`THROW_INTERNAL_EXCEPTION`) whenever the graph was modified while a pending index exists — even if the writes touched only *unrelated* tables, since the pending DML block is per label+property. With the default `checkpoint_on_close=true` that is a new and surprising `Close()` failure mode.

It also interacts badly with the WAL-replay deadlock raised in the earlier review: modified graph → close throws → data stays in WAL → next open replays WAL with the extension still unloaded → open fails if any redo record touches the pending table. Consider letting the shutdown checkpoint proceed (skip dumping pending indexes, keep their descriptors in the manifest), or at minimum making the close-path failure message explain that data remains safe in WAL.

src/storages/index/storage_index_manager.cc
```diff
@@ -149,7 +180,40 @@ void StorageIndexManager::Open(std::shared_ptr<Checkpoint> ckp,
...
+    if (!module) {
+      ++it;
+      continue;
+    }
+    if (!dynamic_cast<StorageIndex*>(module.get())) {
+      RETURN_STATUS_ERROR(
+          StatusCode::ERR_SCHEMA_MISMATCH,
+          "Module is not an index type: " + it->second.descriptor.module_type);
+    }
+    module->Open(*ckp_, ckp_->GetMeta(), it->second.descriptor, memory_level_);
```
zhanglei1949: Two error-handling gaps here:

1. `Module::Open` reports failures by throwing, so a single broken/incompatible pending index throws straight through `ActivateIndexes` (past the `result<>` API) and surfaces as a LOAD failure via the pipeline's catch-all — while *other* pending indexes registered by the same extension are never attempted. Indexes activated earlier in the loop stay activated, so the user sees `LOAD failed` yet `SHOW_INDEXES` changes partially on retry.

2. Related to the earlier `Rebind` comment: the index is moved out of `pending_indexes_` here, before `PropertyGraph::ActivateIndexes` has successfully rebound it, so any later failure leaves a registered-but-unbound index.

Consider per-index try/catch with error aggregation (activate what can be activated, report the rest), or roll back the whole activation on first failure.

src/storages/graph/graph_interface.cc
```diff
@@ -710,4 +710,13 @@ Status StorageAPUpdateInterface::DropIndex(const std::string& name) {
   return Status::OK();
 }
 
+Status StorageAPUpdateInterface::ActivateIndexes() {
+  RETURN_IF_NOT_OK(graph_.ActivateIndexes());
+  mut_view_.Rebuild(graph_);
```
zhanglei1949, line 746: This rebuilds the view and fires `on_planning_changed_` on *every* `LOAD <extension>` — including unrelated extensions and repeated loads — even when no index was activated. `Rebuild` walks the graph and the planning callback invalidates cached plans, so workloads that load extensions frequently pay a real cost for a no-op. Consider only rebuilding when `graph_.ActivateIndexes()` actually activated something (e.g. have it return a count/bool and guard both the rebuild and the callback).

zhanglei1949 (commented):
Thanks for tackling #852 — the pending-index design is the right shape, and the fork/exec cross-process reopen test gives real end-to-end confidence. I verified the concerns from the earlier review against the current tree and would like to add a few findings on top:

1. **The WAL-replay deadlock is real — confirmed the exact path.** `UpdateTransaction::IngestWal` calls `addVertexIndexData` (src/transaction/update_transaction.cc, kInsertVertex redo) and `updateVertexIndexData` (kUpdateVertexProp redo), both of which call `index_manager.GetIndex(label, property)` per property. With this PR a pending index makes that call fail, and `THROW_STORAGE_EXCEPTION_STATUS` then aborts redo — before any connection can run `LOAD`. Skipping index maintenance for pending indexes during replay and rebuilding contents at activation time (e.g. `BulkBuild` after rebind) looks like the only sound fix. Blocking DML/DDL on the affected table until activation is otherwise acceptable, but should be documented.

2. **The checkpoint guard also makes `Close()` throw** (inline): the shutdown checkpoint hits the same `HasPendingIndexes()` check, so closing a database that was modified while an index is pending fails — even when the writes touched only unrelated tables. That is a new, surprising `Close()` contract and pushes more users into the un-merged-WAL state that feeds the replay deadlock above.

3. **Activation has no per-index error isolation** (inline): one broken pending index aborts activation of the rest, and `module->Open()` throws through the `result<>`-based API.

4. **Unconditional view rebuild on every `LOAD`** (inline): `ActivateIndexes()` rebuilds the view and fires `on_planning_changed_` even when nothing was activated.

Minor: hardening `StorageIndexManager::Open` from LOG(WARNING)+skip to THROW when `index_meta` is absent changes recovery behavior for malformed/older manifests (inline nit).

Non-blocking follow-ups:
- **Service/TP mode**: the pending state is only resolved via `StorageIndexDDLInterface::ActivateIndexes`, which TP storage does not implement. If a TP instance ever opens a checkpoint containing an extension-backed index, the index stays pending forever with no activation path. Worth at least a follow-up issue.
- **Docs**: the pending-state restrictions (DML/DDL blocked on the table, checkpoint refused, close-checkpoint refused) should be documented alongside `LOAD` semantics.

src/storages/index/storage_index_manager.cc
```diff
@@ -138,7 +159,17 @@ void StorageIndexManager::Open(std::shared_ptr<Checkpoint> ckp,
 
     auto index = store.TakeModule<StorageIndex>(key, false);
     if (!index) {
-      LOG(WARNING) << "Index module not found in broker: " << key;
+      auto index_meta = desc.get("index_meta");
+      if (!index_meta) {
+        THROW_RUNTIME_ERROR("Index module '" + key +
```
zhanglei1949: Nit: this hardens the previous tolerant behavior (`LOG(WARNING)` + skip) into a hard failure to open the database whenever an index module is absent from the broker *and* its descriptor lacks `index_meta`. Today's `StorageIndex::Dump` always writes `index_meta`, so this should only trigger for malformed or older manifests — but on a recovery path, throwing converts 'index silently dropped' into 'DB won't open'. Consider `LOG(ERROR)` + skipping the entry (the old behavior) here, since the pending path below already covers the well-formed case.

src/storages/index/storage_index_manager.cc
shirly121: Fixed in 868a5a40. WAL replay now checks whether the affected label/property has a pending index before index maintenance. For related vertex insert/update/delete records, it records the mutation in WAL order instead of calling GetIndex and failing recovery. ActivateIndexes replays those mutations after Open and Rebind, so post-checkpoint writes are applied without a full BulkBuild.

src/storages/index/storage_index_manager.cc
shirly121: Fixed in 868a5a40. DROP INDEX now detects a pending entry and removes its metadata directly. If it was the last pending index, the shared pending-mutation list is also cleared, providing an escape hatch when the extension cannot be loaded.

src/storages/index/storage_index_manager.cc
shirly121: Fixed in 998ca2ca. The pending-index error no longer hardcodes vector_search. It reports descriptor.module_type and asks the user to load the extension that registers that module type.

src/storages/graph/property_graph.cc
shirly121: Fixed in 868a5a40. ActivateIndexes now opens, rebinds, and replays pending mutations on temporary candidate indexes first. It moves candidates into indexes_ and erases pending_indexes_ only after all activation steps succeed, so a Rebind/Replay failure leaves the original pending state unchanged.

tests/storage/test_e2e_index.cc
shirly121: Fixed in 998ca2ca. SHOW_INDEXES now combines active and pending indexes and adds a state column with active/pending values, so pending indexes remain discoverable before their extension is loaded.

src/storages/module/module_broker.cc
shirly121: Fixed in 998ca2ca. ModuleBroker::Open now emits LOG(INFO) when it skips an optional unknown module, including both descriptor.module_type and the manifest entry name.

src/main/checkpoint_coordinator.cc
shirly121: Fixed. The checkpoint guard now checks pending mutations rather than merely the presence of pending indexes. If a pending index has no related WAL mutations, its unchanged module descriptor/files (including referenced modules) are linked into the new checkpoint, so shutdown/manual checkpoints can proceed. If related pending mutations exist, the checkpoint is rejected to avoid losing unapplied index updates; Close() preserves the existing contract by logging the checkpoint-on-close failure and completing cleanup normally. Cross-process tests cover both the safe unrelated-write checkpoint and the pending-mutation explicit/close-checkpoint paths.

src/storages/index/storage_index_manager.cc
shirly121: Fixed with all-or-nothing activation. StorageIndexManager::ActivateIndexes now opens, rebinds, and replays pending mutations into a temporary candidate list first. It does not mutate pending_indexes_ or indexes_ until every candidate has completed successfully. Therefore an exception/status failure leaves all pending/active state unchanged, and successful candidates are committed together only after the preparation phase finishes.

src/storages/graph/graph_interface.cc
shirly121: Fixed in 4db4ad64. StorageIndexManager::ActivateIndexes and PropertyGraph::ActivateIndexes now return the number of indexes actually activated. StorageAPUpdateInterface::ActivateIndexes only rebuilds the graph view and invokes on_planning_changed_ when that count is non-zero, so unrelated or repeated LOAD operations remain no-ops for view/planning state.

src/storages/index/storage_index_manager.cc
shirly121: Fixed in 4db4ad64. When an unavailable index module has no persisted index_meta, StorageIndexManager::Open now emits LOG(ERROR) and skips the malformed/legacy entry instead of throwing and preventing database recovery. Well-formed extension-backed descriptors still enter the pending-index path.

longbinlai (approved):
LGTM

PR #863, opened 2026-08-11
WIP: unify COW update interface for AP and TP

### LINKED ISSUE #840: Unify COW update transaction lifecycle for AP and TP as foundation for explicit transactions
## Background

Currently, AP update/schema statements use an in-place write path through `StorageAPUpdateInterface`, while TP update/schema statements use a copy-on-write (COW) path via `StorageTPUpdateInterface` and `UpdateTransaction`. The two paths diverge in behavior, resource ownership, and rollback semantics:

- AP updates cannot be aborted safely because they mutate the live graph directly.
- `UpdateTransaction` is constructed with allocator, WAL writer, and snapshot store references, making its resource boundary unclear and preventing it from being reused across multiple statements.
- Index DDL logic is embedded in `StorageUpdateInterface`, which complicates sharing it between in-place and COW contexts.

## Goal

Refactor the update transaction layer so that AP and TP share a single COW-based update path, with explicit control over whether WAL redo is written. This unification is a structural prerequisite for explicit transaction control (`BEGIN` / `COMMIT` / `ABORT`) planned for NeuG v0.3.

## What this PR does

- Introduce `TransactionDurability { kNoWal, kWal }` and factory methods `UpdateTransaction::Begin` / `BeginUntil`.
- Make `UpdateTransaction` own only the COW workspace and timestamp lease; allocator and WAL writer are supplied at statement/commit time.
- Rename `StorageTPUpdateInterface` to `StorageCOWUpdateInterface` and add `StorageAPCOWUpdateInterface` for AP update/schema statements.
- Switch AP update/schema execution to the COW path with `kNoWal`, so aborts discard the clone and behavior matches TP.
- Extract index DDL into a dedicated `StorageIndexDDLInterface` and a reusable `index_ddl` helper namespace.
- Update transaction documentation (`transaction.md`, `README.md`) to reflect the new model and v0.3 scope.
- Adapt existing unit tests to the new `Begin` / `Commit` API.

## Value for explicit transactions

A unified COW workspace, clear `Begin` / `Commit` / `Abort` lifecycle, and optional WAL durability are exactly what an explicit transaction session needs:

1. The session can hold one `UpdateTransaction` across multiple `execute()` calls.
2. Statement-local resources (allocator, WAL writer) can be borrowed per statement rather than owned by the transaction.
3. `kNoWal` vs `kWal` maps naturally to embedded-analytics vs service-transactional modes.
4. Abort semantics are consistent regardless of execution mode.

## Assignees

- @zhanglei1949

zhanglei1949:
Fix #840

### DIFF
--- a/doc/source/transaction/transaction.md
+++ b/doc/source/transaction/transaction.md
@@ -25,7 +25,19 @@ conn.execute("CREATE (p:Person {name: 'Alice'})")  # Transaction 1
 conn.execute("CREATE (p:Person {name: 'Bob'})")    # Transaction 2
 ```
 
-> **Coming in v0.2:** Multi-statement execution (multiple statements separated by `;` in a single `execute()` call) and explicit transaction control (`BEGIN`/`COMMIT`/`ABORT`) will be supported in NeuG v0.2.
+> **Planned for NeuG v0.3:** Explicit transaction control (`BEGIN`/`COMMIT`/`ABORT`).
+> This is not available in current releases.
+
+### Explicit Transaction Scope (v0.3)
+
+NeuG does **not** plan to support multi-statement batches in the near term:
+
+- One `execute()` call continues to accept exactly one Cypher statement.
+- Statements separated by `;` in one `execute()` call are not a transaction feature and are not on the v0.3 roadmap.
+- Savepoints are out of scope for v0.3.
+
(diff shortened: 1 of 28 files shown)


zhanglei1949:

PR #867, opened 2026-08-12
fix: Manage loaded extensions to avoid loading multiple times

### LINKED ISSUE #866: bug: LOAD EXTENSION can initialize the same extension multiple times
## Problem

`LOAD EXTENSION` does not check whether the requested extension has already been loaded in the current process. Executing the statement multiple times loads or initializes the same extension repeatedly and invokes its `Init()` function each time.

For example:

```cypher
LOAD vector_search;
LOAD vector_search;
LOAD vector_search;
```

Each invocation may execute extension registration logic again, including function, optimizer rule, module, or other global registrations. Depending on the extension implementation, this may cause duplicate registrations, conflicting state, memory/resource leaks, or behavior that depends on how many times `LOAD EXTENSION` was executed.

## Expected behavior

Extension loading should be idempotent within a process. Before loading and initializing an extension, NeuG should check the loaded-extension registry. If the extension is already loaded, it should avoid reopening the library and must not execute `Init()` again.

The repeated statement may either:

- return success as an idempotent no-op, preferably with an informational log; or
- return a clear `already loaded` status.

In either case, extension initialization must run at most once per process.

## Proposed solution

- Normalize the extension identity/name before checking its state.
- Check the loaded-extension registry before calling the dynamic library load and `Init()` path.
- Make the check and state transition thread-safe so concurrent LOAD requests cannot initialize the same extension twice.
- Record an extension as loaded only after initialization succeeds. Failed initialization should remain retryable.

## Acceptance criteria

- Repeated `LOAD EXTENSION` statements do not invoke the extension `Init()` function more than once.
- Repeated loads produce a deterministic success/no-op or explicit already-loaded result.
- Concurrent attempts cannot initialize the same extension multiple times.
- A failed initialization does not permanently mark the extension as loaded and can be retried.
- Tests cover sequential duplicate loads, concurrent duplicate loads, and retry after initialization failure.

shirly121:
Manage loaded extensions to avoid loading multiple times.

Another fix is to run init functions when the database is reopened. Before this fix, functions registered during the init phase were lost when a database reopen happened. This PR fixes that by rerunning init functions when the database is reopened.

Add one more test that switches from AP to TP and checks whether the extension still exists. The test uses the GDS extension to verify this.

Fixes #866


src/execution/extension/extension.cc
```diff
@@ -477,37 +478,43 @@ Status load_extension(const std::string& extension_name) {
...
-                        ". Error: " + std::string(e.what()));
-    } catch (...) {
-      dlclose(handle);
-      return Status(StatusCode::ERR_IO_ERROR,
-                    "Extension initialization failed with unknown error: " +
-                        extension_name);
+    void* handle = ExtensionManager::GetLoadedExtensionHandle(extension_name);
+    bool newly_loaded = handle == nullptr;
+    if (newly_loaded) {
+      handle = dlopen(userLibPath.c_str(), RTLD_NOW | RTLD_LOCAL);
```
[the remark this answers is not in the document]
shirly121: Addressed in 93d126ed. Added deterministic C++ coverage for sequential duplicate acquisition, 16-way concurrent duplicate acquisition (exactly one owner), and retry after a failed load. All three ExtensionLoadCoordinator tests pass.

src/compiler/extension/extension_manager.cpp
```diff
@@ -29,11 +29,50 @@
 namespace neug {
 namespace extension {
 
+std::mutex ExtensionManager::loaded_extensions_mutex_;
+std::unordered_map<std::string, ExtensionManager::LoadedExtension>
+    ExtensionManager::loaded_extensions_;
```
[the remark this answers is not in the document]
shirly121: Addressed in 93d126ed. extension_manager.cpp now explicitly includes both <utility> and <vector>.

src/execution/extension/extension.cc
```diff
@@ -477,37 +478,43 @@ Status load_extension(const std::string& extension_name) {
...
+      if (!handle) {
```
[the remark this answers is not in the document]
shirly121: Addressed in 93d126ed with a CAS-based load coordinator. The registry uses immutable atomically-published snapshots, and each extension entry has atomic LOADING/LOADED/FAILED state. Exactly one caller owns dlopen/dlsym/Init; concurrent callers wait for release/acquire publication, and FAILED can be claimed by CAS for retry. No mutex is held around extension code.

include/neug/compiler/planner/gopt_planner.h
```diff
@@ -28,6 +29,7 @@ class GOptPlanner : public neug::IGraphPlanner {
   GOptPlanner() : IGraphPlanner() {
     database = std::make_unique<neug::main::MetadataManager>();
     neug::main::MetadataRegistry::registerMetadata(database.get());
+    neug::extension::ExtensionManager::InitLoadedExtensions();
   }
```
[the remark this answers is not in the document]
shirly121: Clarified and addressed in 93d126ed. The API is now named ReplayLoadedExtensions, with a planner-constructor comment explaining that replay targets the newly-created catalog and does not reload the library. First-load Init remains guarded by the CAS load coordinator. Extension registration APIs are catalog-idempotent, so replay safely repopulates a fresh planner catalog.

src/compiler/extension/extension_manager.cpp
```diff
@@ -29,11 +29,50 @@
...
+    std::lock_guard<std::mutex> lock(loaded_extensions_mutex_);
+    init_functions.reserve(loaded_extensions_.size());
+    for (const auto& [_, extension] : loaded_extensions_) {
+      init_functions.push_back(extension.init);
+    }
+  }
+  for (auto init : init_functions) {
+    init();
+  }
+}
```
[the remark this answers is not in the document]
shirly121: Addressed by the existing 23320bea follow-up and retained in 93d126ed. Replay catches both std::exception and unknown exceptions and rethrows with the normalized extension name for startup diagnostics.

include/neug/compiler/planner/gopt_planner.h
```diff
@@ -28,6 +29,9 @@ class GOptPlanner : public neug::IGraphPlanner {
   GOptPlanner() : IGraphPlanner() {
     database = std::make_unique<neug::main::MetadataManager>();
     neug::main::MetadataRegistry::registerMetadata(database.get());
+    // A new planner owns a new catalog. Replay extension registration into
+    // that catalog without reloading the extension library.
+    neug::extension::ExtensionManager::ReplayLoadedExtensions();
```
longbinlai: This changes the `Init()` contract in a way worth documenting: extension `Init()` now re-runs on **every** planner recreation (i.e. each AP↔TP switch / reopen), not just on `LOAD`. I verified the in-tree extensions (gds, pattern_matching, test_out_of_tree) are pure registration and idempotent, and `registerFunction`/`registerRule` have contains-guards while `registerExtension` is a name-keyed map, so replay is safe today. But a third-party extension whose `Init()` performs one-shot side effects (allocating resources, starting threads, registering global callbacks) will now break silently. Suggest documenting the idempotency requirement in the extension development docs (and ideally in a comment above `ReplayLoadedExtensions`).

Also note the interaction with #855's proposed clear-GraphDataCache-on-LOAD: if that lands as-is, the cache would be cleared on every AP→TP switch too, not just on explicit LOAD.

src/compiler/extension/extension_manager.cpp
```diff
@@ -22,18 +22,114 @@
...
+      extension->state.wait(state, std::memory_order_acquire);
+      state = extension->state.load(std::memory_order_acquire);
+    }
+    if (state == LoadState::LOADED) {
+      init_functions.emplace_back(name, extension->init);
+    }
+  }
+  for (const auto& [name, init] : init_functions) {
+    try {
+      init();
```
longbinlai: Replay safety depends on every registration API being idempotent. `registerFunction`/`registerRule` have contains-guards and `registerExtension` is name-keyed, but `ExtensionAPI::registerFileSystem` (extension_api.h:69) has **no** duplicate guard, unlike its siblings. An extension registering a filesystem in `Init()` would re-register on every replay. Either add the same contains-guard there, or document why it's safe.

src/compiler/extension/extension_manager.cpp
```diff
@@ -22,18 +22,114 @@
...
+  ticket.extension->state.notify_all();
+}
+
+void ExtensionManager::ReplayLoadedExtensions() {
+  std::vector<std::pair<std::string, InitFunc>> init_functions;
+  auto snapshot = loaded_extensions_.load(std::memory_order_acquire);
+  init_functions.reserve(snapshot->size());
+  for (const auto& [name, extension] : *snapshot) {
+    auto state = extension->state.load(std::memory_order_acquire);
+    while (state == LoadState::LOADING) {
```
longbinlai: Nit: `ReplayLoadedExtensions` runs inside the `GOptPlanner` constructor, so planner construction blocks here on any extension still in `LOADING` state (e.g. a concurrent `LOAD` in flight). Probably acceptable in practice — just noting that planner creation latency is now coupled with extension loading.

tests/unittest/test_extension.cc
```diff
@@ -1,9 +1,71 @@
...
+  ASSERT_TRUE(first.owns_load);
+  neug::extension::ExtensionManager::CompleteLoad(
+      first, reinterpret_cast<void*>(1), &NoopExtensionInit);
+
+  auto duplicate = neug::extension::ExtensionManager::AcquireLoad(
+      "COORDINATOR_SEQUENTIAL_TEST");
+  EXPECT_FALSE(duplicate.owns_load);
+}
+
+TEST(ExtensionLoadCoordinator, ConcurrentDuplicateLoadHasSingleOwner) {
```
longbinlai: These coordinator tests exercise the ticket machinery with fake handles/init funcs, but nothing asserts the end-to-end invariant at the `load_extension` level: that a duplicate `LOAD` runs `Init()` exactly once. The Python e2e test covers catalog survival across reopen, but a C++ test using a counting `Init` (e.g. in test_out_of_tree or a dedicated fixture) doing `LOAD x; LOAD x;` and asserting init count == 1 would directly pin down the #866 regression.

longbinlai (commented):
Review summary: this is the right root-cause fix for #866 — the lock-free CoW registry with single-owner loading is well designed (memory orders, notify_all, FailLoad on every error path all check out), and the Python e2e AP→TP test directly covers the bug scenario. Main asks: (1) document the new Init() idempotency contract, since Init() now re-runs on every planner recreation; (2) add the missing idempotency guard to `registerFileSystem`; (3) consider a load_extension-level test asserting Init runs exactly once across duplicate LOADs. Minor: UNINSTALL leaves a stale LOADED entry in the registry, so a subsequent LOAD skips dlopen/Init but the library path may have changed — worth a note if uninstall is ever made functional.

include/neug/compiler/planner/gopt_planner.h
shirly121: Thanks for pointing this out. This replay is no longer necessary. The loss of registered extension functions during an AP-to-TP switch was already fixed by #754 (`fix: preserve loaded extensions when starting TP service`). Registered extension functions now have the same lifetime as the database, and switching from AP to TP does not close the database. Therefore, the registrations remain available across the switch and `Init()` does not need to be replayed. The replay interface has been removed in this PR.

src/compiler/extension/extension_manager.cpp
shirly121: Fixed. `FileSystemRegistry::Register` now rejects duplicate protocols and keeps the existing factory, while `ExtensionAPI::registerFileSystem` logs a warning when registration is skipped. I also added `FileSystemRegistry.DuplicateRegistrationKeepsExistingFactory` to cover this behavior.

src/compiler/extension/extension_manager.cpp
shirly121: Thanks for noting this. `ReplayLoadedExtensions` has been removed, so planner construction is no longer coupled to an in-flight extension load through replay. AP/TP transitions no longer wait on or rerun extension initialization.

tests/unittest/test_extension.cc
shirly121: Added an end-to-end C++ test, `ExtensionLoad.DuplicateQueryInitializesExtensionOnce`, using the out-of-tree test extension. It executes `LOAD test_out_of_tree;` twice and checks the exported initialization counter to verify that `Init()` runs exactly once.

include/neug/main/execution_slot.h
```diff
@@ -237,6 +240,7 @@ class ExecutionSlot {
                 QueryExecutionStrategy execution_strategy,
                 IWalWriter* wal_writer,
                 CheckpointCoordinator& checkpoint_coordinator,
+                extension::ExtensionManager& extension_manager,
```
zhanglei1949: ExtensionManager的namespace extension是否可以去掉？

src/main/execution_slot.cc
```diff
@@ -308,7 +354,7 @@ Status ExecutionSlot::executeCore(const std::string& query,
 
   // EXPLAIN CHECKPOINT is non-mutating; skip the checkpoint access-mode
   // validation so it works on read-only databases and with access_mode=read.
-  if (NEUG_UNLIKELY(analysis.checkpoint() &&
+  if (NEUG_UNLIKELY(analysis.isAdmin() &&
```
zhanglei1949, line 386: 这里对吗

src/main/execution_slot.cc
```diff
@@ -237,6 +239,50 @@ Status ExecutionSlot::validateCheckpointRequest(AccessMode access_mode) const {
...
+    auto& slot = write_scope.Snapshot();
+    StorageAPUpdateInterface storage(
+        *slot.mutable_graph(), slot.mutable_view(), write_scope.Timestamp(),
+        alloc_, [&write_scope]() { write_scope.MarkPlanningChanged(); });
+    RETURN_IF_NOT_OK(storage.ActivateIndexes());
+  } else {
+    LOG(WARNING) << "[Admin] TP storage does not support extension index "
+                    "activation yet; skipping pending index activation for "
+                 << load_result->canonical_name;
+  }
```
zhanglei1949, line 310: 对于既非checkpoint，又非extension的查询（现在虽然没有，但是后续又可能支持），这里似乎会silently ignore掉，建议考虑报错。

include/neug/main/execution_slot.h
shirly121: 已修改。`ExtensionManager` 和 `ExtensionEntry` 均已从 `neug::extension` 移到 `neug` 根命名空间，相关声明、成员类型及调用点已同步更新。见 a7b1a1e4。

src/main/execution_slot.cc
shirly121: 已修改。这里原先在 `analysis.isAdmin()` 条件下调用 `validateCheckpointRequest()`，语义不一致；现统一为 `validateAdminRequest()`，校验所有 AdminRequest 的类型/扩展参数、access mode 和只读模式，并将错误信息统一为 Admin request。相关 CHECKPOINT 测试断言也已更新，测试 3 passed。见 a7b1a1e4。

src/main/execution_slot.cc
shirly121: 已修改。`executeAdmin()` 现在仅接受 CHECKPOINT、INSTALL/LOAD/UNINSTALL EXTENSION；遇到其他 AdminType 会明确返回 `ERR_NOT_SUPPORTED`（包含类型值），不会再 silently ignore。`validateAdminRequest()` 中也增加了同样的兜底校验。见 a7b1a1e4。

longbinlai (approved):
LGTM

PR #873, opened 2026-08-13
WIP: implement explicit transaction for both embedded mode and service mode

zhanglei1949:
## Summary

- Unify current-graph reads on ReadTransaction and narrow the writer guard to current-graph publication.
- Classify AP COPY FROM at physical planning time and run it under exclusive in-place admission without WAL.
- Publish successful AP bulk loads through an incremental checkpoint, preserving the consuming mmap rename fast path while adding an explicit non-consuming snapshot path.
- Retain the current graph and block further service after failed or incompletely published unlogged mutations.
- Drain active execution slots during service shutdown.
- Add W6 TP explicit transactions: a private ServiceTransactionManager owns opaque HTTP session IDs, capacity reservation, same-ID single-flight, request-scoped slot leases, absolute deadlines/reaper, rollback-only handling, and shutdown drain.
- Add dedicated transaction routes for begin/query/commit/rollback/schema; transaction IDs stay in X-Transaction-Id, while ordinary /cypher remains auto-commit.
- Add W6 Java Session control APIs and private token/state handling, including rollback-only and commit-unknown behavior without transparent transport retry.

## Stack

This draft contains the W2 foundation, W3 AP COPY/checkpoint work, and the completed W6 TP service/Java-client explicit-transaction implementation. It is intended to be reviewed after #863 and the W2B private-COW commit.

## Validation

- Built transaction_test, storage_test, test_db_svc, test_column, test_mmap_container, and gopt_test with -j8.
- 23 focused transaction/WAL/checkpoint-admission tests passed.
- 26 checkpoint manager/file-manager tests passed.
- 5 LIST column and 6 mmap-container tests passed.
- 4 service lifecycle/cache/timeline tests passed with MODERN_GRAPH_DATA_DIR=example_dataset/modern_graph.
- 3 compiler COPY/LOAD flag tests passed with TEST_RESOURCE=tests/compiler.
- W6 adds focused C++ service/connection coverage and Java session/mock-server coverage for cross-request transactions, lifecycle, deadline, and unknown-response behavior.
- git diff --check passed.
- graphify update . completed.

## Deferred

- W1 WAL framing, checkpoint epoch, and strict crash recovery are intentionally out of scope. The current WAL format is assumed for this stage; this PR does not claim full crash durability, commit receipts/idempotency, or final commit-unknown recovery semantics.
- Additional remote bindings remain follow-up work; Java is the first W6 client.
- Local design documents are intentionally not included in this PR.


zhanglei1949:

Issue #779, opened 2026-07-27
[Tracking] Support explicit transaction control (BEGIN / COMMIT / ROLLBACK) end-to-end


longbinlai:
## Background

Today, every statement submitted to NeuG is implicitly wrapped in a single-statement transaction by the connection layer. Users have no way to:

- Open a transaction explicitly with `BEGIN TRANSACTION` (or `BEGIN TRANSACTION READ ONLY`).
- Bundle multiple statements into one atomic unit (e.g. create a node and a relationship based on it without committing the create in between).
- Explicitly `COMMIT` or `ROLLBACK` an in-flight transaction.
- Get atomicity guarantees across `CREATE` + `MATCH` + `DELETE` chains that span multiple queries.

The Cypher grammar and parser already accept the keywords (see `Cypher.g4:203-207` and `parser/transform/transform_transaction.cpp`), and the binder/planner emit a `BoundTransactionStatement` for them. The gap is everything **below** the compiler: session/connection lifecycle, runtime execution, the Python binding, and the HTTP service.

This milestone tracks the end-to-end delivery of explicit transaction control so users can drive long-running, multi-statement transactions safely.

### In scope

1. **Session-level transaction state** on `Connection` (and Python `Session`):
   - Default `auto_commit = true`; `BEGIN` switches the session to an active transaction.
   - Subsequent statements participate in the active transaction until `COMMIT` / `ROLLBACK` / connection close.
   - `Connection::Close()` rolls back any in-flight transaction (no silent partial commit).
2. **Multi-statement execution path** that supports `BEGIN; stmt1; stmt2; ...; COMMIT;` in a single `Query`/`execute` call.
   - Detect `TransactionAction` statements while iterating statements and steer session state accordingly.
   - Reject mixing read-only txn control with DDL/DML (per #760 follow-up).
3. **Runtime execution** of `BoundTransactionStatement`:
   - `BEGIN_WRITE` / `BEGIN_READ`: open a `Transaction` on the storage layer, set the active read/write timestamp, attach to the connection's `ClientContext`.
   - `COMMIT`: drive the two-phase commit (per #86 staging), flush WAL, publish the new snapshot.
   - `ROLLBACK`: discard staged mutations, restore prior snapshot, no WAL durability required.
4. **Read-only enforcement**: `BEGIN TRANSACTION READ ONLY` rejects any statement classified as write/schema.
5. **AP/TP integration** (with #86, #762):
[...]

Issue #715, opened 2026-07-09
Avoid unnecessary checkpointing when starting TP service after AP open without changes


zhanglei1949:
## Background

NeuG supports both AP embedded mode and TP service mode. Today, Python users who want to start a TP service from an existing database directory usually create a `Database` object first, open the directory, and then call `serve()`.

## Problem

Before starting the TP service, `serve()` closes and reopens the database. During close, the default configuration may trigger a checkpoint so that changes made in AP mode, which are not protected by WAL, are persisted before TP transactions are accepted.

This safety guard is necessary, but it currently does not distinguish whether any AP-side modification actually happened after the directory was opened. If a user only opens a directory and then immediately starts the TP service without making AP writes, NeuG still pays the cost of checkpointing and reopening. For large graphs or large database directories, this can significantly increase service startup latency.

## Impact

- Starting a TP service from an existing directory can be slower than necessary.
- Users pay checkpoint cost even when the opened database has not been changed in AP mode.
- The current API does not clearly express the intent of "start a TP service directly from this database directory".
- The AP-to-TP lifecycle boundary lacks explicit state semantics, so the implementation has to use a conservative path.

## Requirements

- Avoid unnecessary checkpointing when no AP-side modification needs to be persisted before TP service startup.
- Preserve the invariant that TP WAL must never depend on AP changes that have not been persisted.
- Clarify lifecycle semantics when starting a TP service from a database directory.
- Ensure resources and state are handled correctly for errors, service shutdown, and repeated startup attempts.

## Non-goals

This issue only describes the problem and constraints. It does not discuss or propose a specific implementation.

zhanglei1949:
Since both AP and TP use WAL for persistence, no checkpointing is required when switching between them.

PR #887, opened 2026-08-17
refactor: unify AP and TP COW write foundations

### LINKED ISSUE #886: Refactor AP auto-commit writes onto private COW transactions
## Background

The COW transaction work in #840 evolved beyond the original `UpdateTransaction` shape. AP ordinary writes, TP snapshot writes, bulk/index in-place operations, and reusable graph mutation logic now need explicit ownership boundaries before public explicit-transaction APIs are added.

This issue tracks the focused implementation that consolidates those foundations without exposing `BEGIN` / `COMMIT` / `ROLLBACK` yet.

## Scope

- Extract reusable COW mutation state, storage adaptation, and logical WAL replay from transaction lifecycle ownership.
- Add `CurrentCowWriteTransaction` for ordinary AP auto-commit writes.
- Add `SnapshotCowWriteTransaction` for TP/WAL-backed snapshot updates.
- Add `CurrentGraphWriteGuard` so AP query classification and execution retain one writer-exclusive guard, timestamp, snapshot, and prepared plan.
- Keep AP bulk import and index DDL on the existing in-place path, under the same guard used for classification.
- Use explicit detach points through `CowDetachLedger`.
- Preserve reusable index DDL behavior and coherent planning-generation publication.

## Correctness requirements

- Ordinary AP writes publish atomically on commit and discard private COW state on failure.
- Readers never observe partially published graph/schema state.
- TP updates retain WAL-backed commit behavior.
- AP in-place bulk/index operations do not release and reacquire writer admission between classification and execution.
- Existing auto-commit, read-view publication, index persistence, and COW commit/abort tests remain green.

## Out of scope

- Public explicit-transaction APIs.
- Checkpoint-directory redesign.
- WAL framing changes.
- W2 performance optimization.

## Related work

- Original COW lifecycle proposal: #840
- Explicit transaction tracking: #779

zhanglei1949:
Closes #886, #715

Related: #840, #779

## Summary

- Unify AP and TP COW mutation foundations around `CowGraphWorkspace`,
  `CowGraphStorage`, lazy detach state, and logical redo.
- Use `CurrentCowWriteTransaction` for AP direct auto-commit writes and
  `SnapshotCowWriteTransaction` for TP/versioned writes. Both reuse the same
  storage mutation implementation while retaining their own publication rules.
- Keep one `CurrentGraphWriteGuard` across AP classification and execution.
  Ordinary AP writes mutate a private clone, prepare the view, append logical
  WAL, and atomically replace the current snapshot. Failed preparation discards
  the private workspace.
- Route AP COPY/batch insert and index DDL through `BulkCowGraphStorage` over the
  same private workspace. Bulk statements keep the native batch/index path, do
  not emit per-row transaction WAL, and commit through a statement-level
  checkpoint before atomically replacing the current graph.
- Remove the AP in-place workspace mode, partial-apply semantics, pending
  checkpoint barrier, and their publication helpers.

## Persistence and safety boundary

- Ordinary AP DML/DDL is private COW plus logical WAL.
- AP COPY/batch insert and index DDL is private COW plus checkpoint; success
  publishes the checkpoint, installs the private graph, and rotates active WAL
  writers to the new epoch before releasing the write guard.
- Validation and staging failures occur before consuming checkpoint work and can
  discard the private workspace. The dirty-module dump is consuming; failures
  after it begins must be fail-stop until every checkpointed payload is private.
- `VecColumn` payload storage remains intentionally shared by ordinary COW
  detach. This is a documented risk for consuming bulk checkpoints and is not
  hidden by suppressing failures or reporting a rollback.

## Capability boundary

- `CowGraphStorage` supports ordinary DML/small DDL and logical redo, but does
  not expose index DDL and rejects batch insertion.
- `BulkCowGraphStorage` adds COPY/batch insertion and index
  create/drop/activation only for AP direct auto-commit execution.
- TP auto-commit and explicit transactions do not receive bulk/index
  capabilities; unsupported operations are rejected before mutation.
- `COPY TEMP` uses private COW and an internal transient publication path without
  WAL or checkpoint.

## Scope boundaries

- No public explicit-transaction API in this PR.
- No new checkpoint format, checkpoint-directory redesign, or non-consuming
  snapshot primitive; checkpoint storage uses the implementation already merged
  through #904.
- No WAL framing or commit-receipt redesign.

## Validation

- Built `neug` and the focused transaction, AP-index, TP-index, and service test
  targets.
- Passed `transaction_test`, `read_view_publication_test`, `ap_index_test`,
  `tp_index_test` (15/15), and `test_db_svc` (37/37, with
  `MODERN_GRAPH_DATA_DIR=example_dataset/modern_graph`).
- `git diff --check upstream/main...HEAD` passed.


include/neug/transaction/transaction_utils.h
```diff
@@ -14,52 +14,23 @@
...
+#include <string>
+#include <vector>
 
 #include "glog/logging.h"
 #include "neug/storages/graph/property_graph.h"
-#include "neug/storages/graph_snapshot_store.h"
-#include "neug/transaction/timestamp_lease.h"
 #include "neug/utils/likely.h"
 #include "neug/utils/property/types.h"
 #include "neug/utils/serialization/in_archive.h"
```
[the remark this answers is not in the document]
zhanglei1949: resolved

liulx20 (changes requested):
I found four issues that should be addressed before merging:

1. **[P1] Schema-inferred persistent `COPY FROM` cannot commit.**

   When the target table does not exist, the direct COPY path executes `CreateVertexType`/`CreateEdgeType` through `BulkCowGraphStorage`. The shared create implementation appends logical schema redo, while `CommitWithCheckpoint` rejects any transaction containing logical redo. As a result, a persistent `COPY Foo FROM ...` with inferred schema aborts after execution. The bulk create path needs to be checkpoint-only, and an empty inferred COPY still needs to mark the schema mutation so it is not discarded.

   Relevant code: [schema redo](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L257-L269), [checkpoint commit rejection](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/main/checkpoint_coordinator.cc#L113-L128).

2. **[P1] Transient labels destabilize the numeric label IDs stored in durable WAL.**

   Temporary type creation is published without WAL, but later DML redo serializes numeric label IDs. For example, after `COPY TEMP T ...; CREATE NODE TABLE P ...; CREATE (:P {id: 1});`, recovery omits `T`, assigns `P` a different ID, and then applies the insert using the stale ID. WAL should use a stable identifier such as the label name, durable IDs should be reserved independently, or transient labels should use a separate ID namespace.

   Relevant code: [transient create path](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L265-L269), [numeric label IDs in redo](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/transaction_utils.cc#L200-L217).

3. **[P1] Mutations of an existing temporary label are written to durable WAL.**

   Only temporary type creation/deletion is classified as transient. Property ALTER operations and vertex/edge insert, delete, and update paths append logical redo unconditionally, so a valid mutation of a temporary table can commit WAL that recovery later tries to apply after that temporary type has disappeared. Every mutation touching a temporary label needs to be classified as transient (with mixed transient/durable mutations rejected), or mutations of temporary labels should be rejected.

   Relevant code: [schema mutation logging](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L331-L349), [DML logging](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L620-L655).

4. **[P2] Shutdown checkpoint GC runs before the WAL writer is closed.**

   `createCheckpointOnClose()` calls `CollectGarbage()` while `wal_writers_` is still alive; the writer is reset only after the function returns. GC can therefore try to remove a retired WAL epoch with an open file descriptor, which fails on platforms such as Windows and unlinks a still-open file on POSIX. Close/reset the writer before collecting retired epochs.

   Relevant code: [checkpoint and GC order](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/main/neug_db.cc#L683-L706), [writer reset](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/main/neug_db.cc#L284-L286).

Requesting changes because the first three issues affect supported persistence and recovery paths.

src/transaction/snapshot_cow_write_transaction.cc
```diff
@@ -0,0 +1,129 @@
...
+    return false;
+  }
+  auto prepared = std::move(prepared_result).value();
+
+  logical_redo.finalize(timestamp());
+  if (!wal_writer_.append(logical_redo.data(), logical_redo.size())) {
+    LOG(ERROR) << "Failed to append wal log";
+    Abort();
+    return false;
+  }
```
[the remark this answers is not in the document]
zhanglei1949: resolved.

zhanglei1949:
> I found four issues that should be addressed before merging:
> 
> 1. **[P1] Schema-inferred persistent `COPY FROM` cannot commit.**
>    When the target table does not exist, the direct COPY path executes `CreateVertexType`/`CreateEdgeType` through `BulkCowGraphStorage`. The shared create implementation appends logical schema redo, while `CommitWithCheckpoint` rejects any transaction containing logical redo. As a result, a persistent `COPY Foo FROM ...` with inferred schema aborts after execution. The bulk create path needs to be checkpoint-only, and an empty inferred COPY still needs to mark the schema mutation so it is not discarded.
>    Relevant code: [schema redo](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L257-L269), [checkpoint commit rejection](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/main/checkpoint_coordinator.cc#L113-L128).
> 2. **[P1] Transient labels destabilize the numeric label IDs stored in durable WAL.**
>    Temporary type creation is published without WAL, but later DML redo serializes numeric label IDs. For example, after `COPY TEMP T ...; CREATE NODE TABLE P ...; CREATE (:P {id: 1});`, recovery omits `T`, assigns `P` a different ID, and then applies the insert using the stale ID. WAL should use a stable identifier such as the label name, durable IDs should be reserved independently, or transient labels should use a separate ID namespace.
>    Relevant code: [transient create path](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L265-L269), [numeric label IDs in redo](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/transaction_utils.cc#L200-L217).
> 3. **[P1] Mutations of an existing temporary label are written to durable WAL.**
>    Only temporary type creation/deletion is classified as transient. Property ALTER operations and vertex/edge insert, delete, and update paths append logical redo unconditionally, so a valid mutation of a temporary table can commit WAL that recovery later tries to apply after that temporary type has disappeared. Every mutation touching a temporary label needs to be classified as transient (with mixed transient/durable mutations rejected), or mutations of temporary labels should be rejected.
>    Relevant code: [schema mutation logging](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L331-L349), [DML logging](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/transaction/cow_graph_storage.cc#L620-L655).
> 4. **[P2] Shutdown checkpoint GC runs before the WAL writer is closed.**
>    `createCheckpointOnClose()` calls `CollectGarbage()` while `wal_writers_` is still alive; the writer is reset only after the function returns. GC can therefore try to remove a retired WAL epoch with an open file descriptor, which fails on platforms such as Windows and unlinks a still-open file on POSIX. Close/reset the writer before collecting retired epochs.
>    Relevant code: [checkpoint and GC order](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/main/neug_db.cc#L683-L706), [writer reset](https://github.com/alibaba/neug/blob/de6f8e9876335f7b797b72dcfdf298059af94eb1/src/main/neug_db.cc#L284-L286).
> 
> Requesting changes because the first three issues affect supported persistence and recovery paths.

Resolved.

include/neug/common/columns/edge_columns.h
```diff
@@ -35,6 +35,10 @@ class IEdgeColumn : public IContextColumn {
 
   virtual EdgeRecord get_edge(size_t idx) const = 0;
 
+  // COW may relocate bundled edge payloads while preserving the edge column's
+  // topology. Refresh the pointer retained by this execution-context row.
+  virtual void set_edge_data_ptr(size_t idx, const void* data) = 0;
```
liulx20: [P1] Rebuild the edge column instead of mutating cached pointers

This row-local setter leaks CSR COW relocation into every execution-column implementation and cannot restore coherence for aliases. The same physical edge can appear in multiple rows or chunks after a join. Updating the first occurrence detaches the bundled adjacency list and refreshes only that row; a later occurrence still carries the pre-detach property pointer. For parallel edges, `record_to_csr_offset_pair()` then disambiguates with stale property bytes and can return the sentinel offset (or select the wrong peer).

Please avoid repairing this through a mutable `IEdgeColumn` API. A safer shape is to resolve the affected rows and RHS values before the first write, apply the updates from those cached locations, then rebuild/replace the affected edge column from the updated CSR so every duplicate row is refreshed together.

src/storages/graph/schema.cc
```diff
@@ -2708,9 +2743,18 @@ Schema Schema::StripTemporary() const {
   stripped.elabel_triplet_tomb_.resize(
       stripped.e_schemas_.empty() ? 0 : max_e_triplet_index + 1);
 
-  // Projected graph metadata has lower priority than the physical schema.
-  // Keep it intact and defer missing-label validation to query binding.
-  stripped.graph_entry_set_ = graph_entry_set_;
+  // A projection that references session-scoped schema is session-scoped too.
+  // Persisting it would leave dangling label references after stripping.
+  for (const auto& [name, entry] : graph_entry_set_.Entries()) {
```
liulx20, line 2755: [P1] Remap persistent schema and index IDs when stripping temporary labels

`StripTemporary()` packs surviving names into a fresh `LabelIndexer`, but it copies each `VertexSchema` and `EdgeSchema` without rewriting their embedded IDs. If a temporary label at ID 0 precedes a persistent label at ID 1, the persistent label moves to slot 0 while its `VertexSchema::label_id` remains 1; edge `entry_id/src_label_id/dst_label_id/edge_label_id` values have the same problem. A manual checkpoint immediately reopens the in-memory manifest, so catalog validation can stop seeing the persistent table.

The index modules are also staged before this stripped schema is installed, and `IndexBindSchema` persists the old numeric label ID. After restart, rebinding can therefore fail with `invalid index label id`. Please either preserve the original slots or apply one remapping to all embedded schema IDs and index metadata, as `Compact()` does for schema entries. A regression should create the temporary label before the persistent indexed label and exercise checkpoint/reopen.

liulx20 (changes requested):

include/neug/common/columns/edge_columns.h
zhanglei1949: resolved.

src/storages/graph/schema.cc
zhanglei1949: remapping implemented.

src/execution/execute/ops/insert/create_edge.cc
```diff
@@ -56,13 +56,9 @@ class CreateEdgeOpr : public IOperator {
...
       expr_properties.emplace_back(std::move(expr_props));
     }
-    return ctx.apply_chunks(
-        [&](ContextChunk&& chunk) -> neug::result<ContextChunk> {
-          return CreateEdge::insert_edge(
-              dynamic_cast<StorageInsertInterface&>(graph_interface),
-              std::move(chunk), labels_, src_dst_tags_,
-              std::move(expr_properties), alias_);
-        });
+    return CreateEdge::insert_edge(
```
liulx20: 这个改动是为啥

src/execution/execute/ops/edge_column_rebuild.cc
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+  return builder.finish();
+}
+
+std::shared_ptr<IContextColumn> RebuildEdgeColumn(
+    const IEdgeColumn& original, const std::vector<EdgeRecord>& records) {
+  assert(original.size() == records.size());
+  switch (original.edge_column_type()) {
+  case EdgeColumnType::kSDSL: {
```
liulx20: 考虑最general的情形，BDML就够了

src/execution/execute/ops/insert/create_edge.cc
zhanglei1949: `apply_chunks()` 会逐个 chunk 执行并立即修改存储。对于 bundled edge，`AddEdge()` 可能触发 CSR 的 COW detach 或扩容，使后续 chunk 中 edge column 保存的属性指针失效。因此必须在第一次写入前捕获整个 Context 中受影响的列，写入完成后再统一刷新所有别名；这个跨 chunk 生命周期无法由 `apply_chunks()` 表达。

src/execution/execute/ops/edge_column_rebuild.cc
zhanglei1949: 已修复

src/execution/execute/ops/insert/create_edge.cc
zhanglei1949: 已撤回 create_edge相关改动，之后再Fix #927

liulx20 (dismissed):

PR #889, opened 2026-08-17
fix: Correct the width setting when opening a string column

### LINKED ISSUE #888: [BUG] VARCHAR max_length not honored for writes in sessions after table creation (silent truncation to 256 bytes)
**Describe the bug**

`VARCHAR(max_length)` is not honored for writes performed in a DB-open session *after* the one that created the table. The persisted schema still reports the declared `max_length`, but new writes are silently truncated to **256 bytes**. Rows written during the creating session are unaffected. Both write paths are affected: parameterized `CREATE` and `COPY`. The truncation is checkpointed and survives a read-only reopen.

Additionally, the declared cap appears to be ignored entirely on the cross-session write path: declaring `VARCHAR(100)` and writing 150 characters reads back 150 (no enforcement). So the behavior is equivalent to a fixed 256-byte buffer: values over 256 bytes are silently cut, and declared caps below 256 are not enforced.

**To Reproduce**

```python
import shutil
from pathlib import Path
import neug

ROOT = Path("/tmp/neug_varchar_repro")
DB = str(ROOT / "db.neug")
VAL = "x" * 400  # > 256, < 512

shutil.rmtree(ROOT, ignore_errors=True)
ROOT.mkdir(parents=True)

# session 1: create table + write
db = neug.Database(DB, mode="w")
c = db.connect()
c.execute("CREATE NODE TABLE T (id VARCHAR(64), val VARCHAR(512), PRIMARY KEY(id))", "schema")
c.execute("CREATE (:T {id:$id, val:$v})", "insert", {"id": "s1", "v": VAL})
r = list(c.execute("MATCH (n:T {id:'s1'}) RETURN n.val", "read"))
print("session1 write+readback len:", len(r[0][0]))
c.close(); db.close()

# session 2: reopen, write a NEW row
db = neug.Database(DB, mode="w")
c = db.connect()
r = list(c.execute("MATCH (n:T {id:'s1'}) RETURN n.val", "read"))
print("old row readback len:", len(r[0][0]))
c.execute("CREATE (:T {id:$id, val:$v})", "insert", {"id": "s2", "v": VAL})
r = list(c.execute("MATCH (n:T {id:'s2'}) RETURN n.val", "read"))
print("session2 write+readback len:", len(r[0][0]))
c.close(); db.close()

# session 3: read-only reopen
db = neug.Database(DB, mode="r")
c = db.connect()
for k in ("s1", "s2"):
    r = list(c.execute("MATCH (n:T {id:'%s'}) RETURN n.val" % k, "read"))
    print("%s persisted len:" % k, len(r[0][0]))
c.close(); db.close()
```

Observed output:

```
session1 write+readback len: 400
old row readback len: 400
session2 write+readback len: 256     <- silently truncated
s1 persisted len: 400
s2 persisted len: 256                <- truncation is checkpointed
```

`Connection.get_schema()` on the reopened database still reports `val` as `var_char(max_length=512)`.

`COPY` variant: creating the table in session 1 and `COPY ... (HEADER=true, DELIMITER=',')`-ing a 400-char value in session 2 also reads back 256.

Cap-below-256 variant: `CREATE NODE TABLE T3 (id VARCHAR(64), small VARCHAR(100), ...)`, then cross-session insert of 150 characters reads back 150 — the declared cap is not enforced at all.

**Expected behavior**

Session 2 writes should honor the persisted column definition: a 400-character value into `VARCHAR(512)` should read back 400; a 150-character value into `VARCHAR(100)` should be rejected or truncated to the declared limit. Silent truncation to an undeclared 256-byte boundary is data loss.

**Environment**

 - NeuG version: 0.1.3 (pip package "GraphScope NeuG for python")
 - OS: macOS (Darwin 25.3.0, arm64)
 - Python Version: 3.10.20

**Additional context**

Suspected area: the catalog loaded at open time may not propagate `var_char.max_length` to the column writers used by subsequent write transactions (falling back to a default 256-byte buffer).

Impact: any incremental workload that opens an existing database and appends data silently corrupts string values longer than 256 bytes, including long `VARCHAR(512)` primary keys. The only workaround we found is dropping and recreating the affected tables and writing them within the same DB-open session.

zhanglei1949:
Fix #888

### DIFF
--- a/include/neug/utils/property/column.h
+++ b/include/neug/utils/property/column.h
@@ -257,6 +257,9 @@ class TypedColumn<std::string_view> : public ColumnBase {
 
   void Open(Checkpoint& ckp, const ModuleDescriptor& desc,
             MemoryLevel level) override {
+    if (auto width = desc.get("width")) {
+      width_ = static_cast<uint16_t>(std::stoul(*width));
+    }
     items_buffer_ = ckp.OpenFile(
         desc.get_path(ModuleDescriptor::kItemsPath).value_or(""), level);
     data_buffer_ = ckp.OpenFile(
@@ -292,6 +295,7 @@ class TypedColumn<std::string_view> : public ColumnBase {
             const std::string& key) override {
     ModuleDescriptor desc;
     desc.module_type = ModuleTypeName();
+    desc.set("width", std::to_string(width_));
     if (!items_buffer_ || !data_buffer_) {
       THROW_RUNTIME_ERROR("Buffers not initialized for dumping");
     }
--- a/tools/python_bind/tests/test_call_string_literal.py
+++ b/tools/python_bind/tests/test_call_string_literal.py
@@ -193,6 +193,76 @@ def test_varchar_column_with_explicit_max_length_truncates_at_that_length(conn):
     assert rows[0][0] == "z" * 80
 
 
+def test_varchar_max_length_survives_reopen_for_create_and_copy(tmp_path):
+    """Issue #888: the checkpoint schema, not the factory default, sets caps."""
+    db_dir = tmp_path / "varchar_reopen"
+    db = Database(db_path=str(db_dir), mode="w")
+    conn = db.connect()
+    conn.execute(
+        "CREATE NODE TABLE t (id VARCHAR(512), val VARCHAR(512), " "PRIMARY KEY(id));"
+    )
+    conn.execute(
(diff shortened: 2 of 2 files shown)


BingqingLyu (approved):
LGTM

liulx20 (approved):

PR #891, opened 2026-08-17
fix: support null values in CSV collections and aggregation function

### LINKED ISSUE #890: Fix null value handling for CSV format and aggregation function
Follow-up to #885.

This issue tracks the remaining null-value handling work for CSV collections and the `collect()` aggregate function.

## CSV null-handling comparison with DuckDB

The default behavior of CSV null-handling in NeuG and DuckDB is almost identical, but there are some shortcomings in the configuration items.
https://duckdb.org/docs/lts/data/csv/overview#parameters

| Scenario | DuckDB vs. NeuG |
|---|---|
| Empty field: `1,,20` | Same: interpreted as `NULL` by default |
| Quoted empty field: `1,"",20` | Same: interpreted as `NULL` by default |
| `force_not_null` | DuckDB supports preserving empty fields as an empty string; NeuG does not support |
| `allow_quoted_nulls` | DuckDB can preserve quoted empty fields as an empty string; NeuG does not support |
| Custom NULL values | DuckDB uses `nullstr`; NeuG supports `null_values`, but not through all CSV interfaces |
| Missing trailing columns | DuckDB errors by default; NeuG pads them with `NULL` |
| NULLs in collections | Supported in #891 |

For the default empty-field behavior, NeuG is aligned with DuckDB. The remaining differences are mainly optional CSV configuration and missing-column handling.

## Aggregate NULL-handling comparison

| Example | DuckDB | Kùzu | Neo4j | NeuG |
| --- | --- | --- | --- | --- |
| `COUNT([1, 2])` | `2` | `2` | `2` | ✅ `2` |
| `COUNT([1, 2, NULL])` | `2` | `2` | `2` | ✅ `2` |
| `COUNT(DISTINCT [1, 2, 2, NULL])` | `2` | `2` | `2` | ✅ `2` |
| `COUNT(*)` over `[1, 2, NULL]` | `3` | `3` | `3` | ✅ `3` |
| `COUNT([])` | `0` | `0` | `0` | ✅ `0` |
| `AVG([1, 2, NULL])` | `1.5` | `1.5` | `1.5` | ✅ `1.5` |
| `MIN([1, 2, NULL])` | `1` | `1` | `1` | ✅ `1` |
| `MAX([1, 2, NULL])` | `2` | `2` | `2` | ✅ `2` |
| `AVG/MIN/MAX([NULL, NULL])` | `NULL` | `NULL` | `NULL` | ✅ `NULL` |
| `SUM([1, 2, NULL])` | `3` | `3` | `3` | ✅ `3` |
| `SUM([NULL, NULL])` | `NULL` | `NULL` | `0` | ⚠️ `NULL` |
| `collect/list([1, 2, NULL])` | `[1, 2, NULL]` | `[1, 2]` | `[1, 2]` | ❓ `[1, 2]` |
| `collect/list(DISTINCT [1, 2, 2, NULL])` | `[1, 2, NULL]` | `[1, 2]` | `[1, 2]` | ❓ `[1, 2]` |
| `collect/list([NULL, NULL])` | `[NULL, NULL]` | `NULL` | `[]` | ❓ `[]` |
| `collect/list(DISTINCT [NULL, NULL])` | `[NULL]` | `NULL` | `[]` | ❓ `[]` |
| `AVG/MIN/MAX([])` | `NULL` | `NULL` | `NULL` | ⚠️ No rows |
| `SUM([])` | `NULL` | `NULL` | `0` | ⚠️ No rows |
| `collect/list([])` | `NULL` | `NULL` | `[]` | ⚠️ No rows |
| Group `[0, NULL, NULL]` by value | `{0: 1, NULL: 2}` | `{0: 1, NULL: 2}` | `{0: 1, NULL: 2}` | ⚠️ `{0: 3}` |
| Group `[]` by value | `[]` | `[]` | `[]` | ✅ `[]` |

Items recommended to remain unchanged:

- ❓ `collect/list`: DuckDB's `list` and Cypher's `collect` have different definitions. Cypher's `collect` filters out NULL values, so NeuG's current behavior is correct and should be kept.
- ❓ `collect` over an all-NULL input: After filtering out NULL values, the correct Cypher result is an empty list (`[]`). NeuG and Neo4j behave correctly; Kuzu's `NULL` result is an implementation limitation.
- ❓ `collect` over an empty input: This can be treated the same as collecting an all-NULL input after all NULL values have been filtered out. No values remain, so the correct Cypher result is an empty list (`[]`). NeuG's current behavior should be kept.

Items recommended to change:

- ⚠️ `SUM` over an all-NULL input: Regardless of the chosen semantics, this behavior should at least be consistent with `SUM` over an empty input. Following Neo4j's definition, NeuG should return `0`, the identity element of addition, instead of `NULL`.
- ⚠️ Aggregation over an empty input: NeuG currently returns no rows for `SUM`, `AVG`, `MIN`, `MAX`, and `collect` when they are queried individually. This is dangerous because, when these functions are combined with another aggregation that produces a row (such as `COUNT`), their zero-length result columns can be exposed as initialized default values (`0`, `0.0`, or `[]`). NeuG should align with Neo4j and always produce one row for an ungrouped aggregate over an empty input: `SUM` should return `0`, `AVG/MIN/MAX` should return `NULL`, and `collect` should return `[]`. See: https://neo4j.com/developer/kb/understanding-aggregations-on-zero-rows/
- ⚠️ Grouping by NULL: NULL should be treated as a distinct grouping value. It must form its own group instead of being converted to a type-specific default value such as `0` and merged with that group.

Spockkk0225:
## Summary
- support null child elements in CSV LIST and ARRAY values, including nested collections
- document and test that collect filters null inputs while UNWIND preserves null elements
- add regression coverage for null propagation through builders, RETURN, UNWIND, and shuffle

## Testing
- execution_test: 113 passed
- targeted Python LIST/ARRAY null tests passed
- Black and git diff checks passed

Fixes #890

### DIFF
--- a/tests/execution/test_runtime_column.cc
+++ b/tests/execution/test_runtime_column.cc
@@ -1464,6 +1464,46 @@ TEST_F(ArrowColumnTest, DataChunkSupplierBasic) {
   EXPECT_GT(total_rows, 0);
 }
 
+TEST_F(ArrowColumnTest, CsvCollectionNullElements) {
+  const char* var = std::getenv("TEST_PATH");
+  std::string test_path = var ? var : "/workspaces/neug/tests";
+  std::string file_path =
+      test_path + "/execution/resources/collection_null.csv";
+
+  CsvReadConfig config;
+  put_boolean_option(config);
+  config.delimiter = '|';
+  config.quoting = false;
+  config.skip_rows = 1;
+  config.column_names = {"id", "list_values", "array_values"};
+  config.include_columns = config.column_names;
+  config.column_types.emplace("id", DataType(DataTypeId::kInt64));
+  config.column_types.emplace("list_values",
+                              DataType::List(DataType(DataTypeId::kInt64)));
+  config.column_types.emplace("array_values",
+                              DataType::Array(DataType(DataTypeId::kInt64), 3));
+  config.null_values = {"NULL"};
+
+  CSVChunkSupplier supplier(file_path, std::move(config));
+  auto chunk = supplier.GetNextChunk();
+  ASSERT_NE(chunk, nullptr);
+  ASSERT_EQ(chunk->row_num(), 1);
+
+  auto list_value = chunk->get(1)->get_elem(0);
+  const auto& list_children = ListValue::GetChildren(list_value);
+  ASSERT_EQ(list_children.size(), 3);
+  EXPECT_EQ(list_children[0].GetValue<int64_t>(), 1);
+  EXPECT_TRUE(list_children[1].IsNull());
+  EXPECT_EQ(list_children[2].GetValue<int64_t>(), 3);
+
+  auto array_value = chunk->get(2)->get_elem(0);
+  const auto& array_children = ArrayValue::GetChildren(array_value);
+  ASSERT_EQ(array_children.size(), 3);
+  EXPECT_EQ(array_children[0].GetValue<int64_t>(), 4);
+  EXPECT_TRUE(array_children[1].IsNull());
+  EXPECT_EQ(array_children[2].GetValue<int64_t>(), 6);
+}
+
 }  // namespace test
 }  // namespace execution
 }  // namespace neug
--- a/tests/execution/test_runtime_column.cc
+++ b/tests/execution/test_runtime_column.cc
@@ -42,7 +42,7 @@ class VertexColumnTest : public ::testing::Test {
     col_builder.push_back_elem(Value::VERTEX(VertexRecord(label, kVid0)));
     col_builder.push_back_vertex(VertexRecord(label, kVid1));
     if (is_optional) {
-      col_builder.push_back_null();
+      col_builder.push_back_elem(Value());
     }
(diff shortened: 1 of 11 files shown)


tests/execution/test_runtime_column.cc
```diff
@@ -42,7 +42,7 @@ class VertexColumnTest : public ::testing::Test {
     col_builder.push_back_elem(Value::VERTEX(VertexRecord(label, kVid0)));
     col_builder.push_back_vertex(VertexRecord(label, kVid1));
     if (is_optional) {
-      col_builder.push_back_null();
+      col_builder.push_back_elem(Value());
```
liulx20: 这个必须换吗，其实NULL也是带类型的，或者可以push_back_elem(Value(kVertex))

Spockkk0225: 已修改为 `push_back_elem(Value(kVertex))`，保留类型信息，更接近真实的数据流。

longbinlai (approved):
LGTM

liulx20 (approved):

Issue #883, opened 2026-08-17
Refactor checkpoint storage into immutable objects and versioned manifests


zhanglei1949:
## Motivation

The current checkpoint layout uses a complete directory as both the publication unit and the reclamation unit:

```text
db/
├── checkpoint-N/
└── checkpoint-(N+1).next/
```

A checkpoint is published by renaming the staging directory, and startup discovers the highest valid published generation. This works for full checkpoints, but it is a poor long-term foundation for incremental checkpoints:

- every generation still needs a complete directory;
- unchanged files must be linked or copied into the new directory;
- live mmap/container objects may continue referencing files in retired directories, so old generations cannot be reclaimed promptly;
- checkpoint publication, WAL generation, immutable snapshot data, and runtime resource lifetimes remain coupled to directory generations;
- adding a generic non-consuming `WriteSnapshot()` virtual API throughout the storage hierarchy would expose this layout problem to every container/module.

We should separate a **logical checkpoint generation** from the **physical immutable objects** it references.

## Proposed layout

The exact names are internal, but the ownership boundaries should be equivalent to:

```text
db/
├── checkpoint/
│   ├── CURRENT
│   ├── manifests/
│   │   ├── <checkpoint-id>.manifest
│   │   └── ...
│   └── objects/
│       ├── <object-id>
│       └── ...
├── wal/
│   ├── <checkpoint-id>/
│   └── ...
└── runtime/
    ├── <open-epoch>/
    └── ...
```

- **Objects are immutable after publication.** Dirty modules/containers create new objects; clean data is reused by reference instead of being linked/copied into another full checkpoint directory.
- **Each manifest is immutable, complete, and self-contained.** It records the format version, `checkpoint_id`, schema/module metadata, the full object reference set, and the WAL epoch required for recovery. A parent id may be retained for diagnostics, but recovery must not require walking a manifest chain.
- **`CURRENT` is the only normal publication selector.** It identifies the currently published manifest. Multiple old manifests may remain temporarily for readers, recovery safety, and GC.
- **WAL is owned by checkpoint id.** Every successful ordinary TP checkpoint allocates a new `checkpoint_id` and starts a fresh WAL epoch. AP and TP writers continue to share the database transaction timeline; this issue does not define a new WAL record format.
- **Runtime files are not checkpoint contents.** Allocator and other mutable scratch files remain under instance/open-epoch-owned runtime directories and are never referenced by a checkpoint manifest.

This generalizes the allocator/runtime separation completed in #758.

[...]

PR #892, opened 2026-08-18
refactor: Store checkpoints as immutable objects and versioned manifests

zhanglei1949:
## What do these changes do?

This PR refactors checkpoint storage and publication around immutable objects and complete, versioned manifests. It does **not** add an incremental-checkpoint entry point; manual `CHECKPOINT` keeps the existing full-checkpoint lifecycle.

- Replace numbered `checkpoint-N` directories with:
  - `checkpoint/CURRENT` as the sole publication selector;
  - immutable `checkpoint/objects/<object-id>` files;
  - complete `checkpoint/manifests/<id>.manifest` files;
  - per-checkpoint `wal/<id>` epochs;
  - per-open `runtime/open-<epoch>` workspaces.
- Publish through one staging protocol: persist new objects, initialize and sync the WAL epoch, persist the complete manifest, and only then atomically replace `CURRENT`.
- Attempt to restore the previous `CURRENT` selector when replacement durability cannot be confirmed. If the selector cannot be confirmed or restored, publication is commit-unknown.
- Keep manual, recovery, and shutdown checkpoints on the compacting, destructive `DumpAndClear` path.
- Preserve manual `CHECKPOINT` behavior: reopen the graph and allocator, rotate Service-mode WAL writers, and reset the transaction timeline before admitting new transactions.
- Let dirty graph/index modules create new immutable objects while clean module descriptors reuse existing object references. “Full checkpoint” describes the runtime lifecycle boundary; it does not require rewriting every clean object.
- Reclaim manifests, objects, WAL epochs, and runtime workspaces independently using reachability and live checkpoint pins. GC is skipped while a staging checkpoint owns unpublished objects.
- Narrow and unify checkpoint APIs and naming, centralize graph dump logic, and validate published manifests and referenced objects strictly.
- Reject the legacy `checkpoint-N` layout without modifying it; migration is not included in this PR.

### Checkpoint behavior

| Path | Compact + destructive dump | Reopen graph / allocator | WAL / timeline |
|---|---|---|---|
| Manual `CHECKPOINT` | Yes | Yes | Rotate Service WAL epoch and reset timeline |
| Recovery checkpoint | Yes | Yes | Recover, publish a full baseline, then initialize the timeline |
| Shutdown checkpoint | Yes | No | Persist and close; no online reset |

The manifest format retains `base_ts` for future incremental checkpoint support, but all checkpoint entry points in this PR publish full baselines with `base_ts=0`. A future explicit-transaction bulk-load path can add incremental publication on top of this storage protocol without changing the on-disk ownership model.

## API migration notes for out-of-tree extensions

Checkpoint APIs were renamed to match the new publication model. Out-of-tree extensions that touch checkpoint or manifest APIs need the following mechanical renames:

`CheckpointManager` / `StagingCheckpoint` (`neug/storages/checkpoint_manager.h`):

| Old | New |
|---|---|
| `CreateStagingCheckpoint()` | `CreateStaging()` |
| `StagingCheckpoint::Commit()` | `StagingCheckpoint::Publish()` |
| `CurrentCheckpoint()` | `Current()` |
| `HasCurrentCheckpoint()` | removed; use `Current() != nullptr` |
| `RestoreCurrentCheckpoint(...)` | removed; `CURRENT` is replaced atomically by `Publish` only |
| `CleanupPublishedCheckpoint(...)` / `CleanupRetiredCheckpoints()` | removed; superseded by `CollectGarbage()` reachability |
| `CommitStagingCheckpoint(...)` | `PublishStagingCheckpoint(...)` |
| `db_dir()` | `database_dir()` |
| `Open(db_dir, recover_workspace)` | `Open(database_dir, create_if_missing)` |

`CheckpointManifest` (`neug/storages/checkpoint_manifest.h`):

| Old | New |
|---|---|
| `module(key)` returning `std::optional` | `FindModule(key)` returning a pointer |
| `set_module(...)` | `SetModule(...)` |
| `has_module(key)` | `HasModule(key)` |
| `remove_module(key)` | removed |
| `LinkModuleFrom(...)` | `ReuseModuleClosureFrom(...)` |
| `modules()` | `Modules()` |
| `mutable_modules()` | removed; use `FindMutableModule(key)` |

`Checkpoint` (`neug/storages/checkpoint.h`):

| Old | New |
|---|---|
| `UpdateMeta(meta)` | `SetManifest(manifest)` |
| `GetMeta()` | `manifest()` |
| `path()` | `manifest_path()` |

The in-tree adaptation in `tests/extension/vec_index` (`mutable_modules().at(key)` to `FindMutableModule(key)`) shows the typical migration.

## Related issue

Related to #883

## Testing

- `cmake --build build --target storage_test transaction_test -j4`
- `build/tests/storage/storage_test '--gtest_filter=CheckpointFormatTest.*'` (13 tests passed)
- Focused coordinator/WAL recovery tests (6 tests passed)
- `git diff --check`
- `graphify . --code-only`

### DIFF
--- a/doc/source/transaction/checkpoint.md
+++ b/doc/source/transaction/checkpoint.md
@@ -111,9 +111,15 @@ data_dir/
                                 # <id> is an opaque unique suffix (a UUID)
 ```
 
-`CURRENT` is the sole publication selector. Its content is the selected manifest's decimal ID followed by a single trailing newline (for example `3\n`), written via an atomic rename; operators may inspect or rewrite it manually with a plain text editor. A published manifest has the required fields `v`, `id`, `base_ts`, `schema`, and `modules`; it may also contain `scalars`. Module descriptors persist object IDs, not absolute paths. The same ID names the manifest and its WAL epoch. `base_ts` is the highest transaction timestamp already represented by the manifest, so recovery replays the selected epoch from `base_ts + 1`. Full checkpoints use `base_ts=0` and reset the transaction timeline after reopening.
+`CURRENT` is the sole publication selector. Its content is the selected manifest's decimal ID followed by a single trailing newline (for example `3\n`), written via an atomic rename; operators may inspect or rewrite it manually with a plain text editor. A published manifest has the required fields `v`, `base_ts`, `schema`, and `modules`; it may also contain `scalars`. Module descriptors persist object IDs, not absolute paths. The same ID names the manifest and its WAL epoch. `base_ts` is the highest transaction timestamp already represented by the manifest, so recovery replays the selected epoch from `base_ts + 1`. Full checkpoints use `base_ts=0` and reset the transaction timeline after reopening.
 
-Checkpoint objects are immutable and may be referenced by several manifests. Runtime files are not checkpoint data: each database open receives its own `runtime/open-<id>/` directory (the suffix is an opaque unique ID, not a timestamp or manifest ID), and closing that database removes only its own unpinned workspace. Existing `checkpoint-N` directories are unsupported and rejected without modification; this format does not migrate or read them. Recover or rebuild such a database with a version that supports the legacy format first.
+Checkpoint objects are immutable and may be referenced by several manifests. Runtime files are not checkpoint data: each database open receives its own `runtime/open-<id>/` directory (the suffix is an opaque unique ID, not a timestamp or manifest ID), and closing that database removes only its own unpinned workspace.
+
+### Upgrading legacy checkpoint directories
+
+When `CURRENT` is absent, the first read-write open automatically upgrades the newest valid released v1 `checkpoint-N` generation. NeuG imports its immutable snapshot files into `checkpoint/objects/`, preserves its generation as the new manifest and WAL epoch ID, and publishes a v2 manifest with `base_ts=0`; normal recovery then replays every legacy WAL record. Files are hardlinked when safe and copied otherwise.
+
+The old directories are not changed before the new `CURRENT` is durably published. A crash before publication leaves the legacy checkpoint usable and the next read-write open retries. After the database has opened and recovered successfully, normal garbage collection removes the old `checkpoint-N` and `checkpoint-N.next` directories, making the upgrade one-way. A legacy-only database cannot be opened read-only: open it once in read-write mode to perform the upgrade. Legacy `meta` versions other than v1 are rejected rather than guessed.
 
 ## What a checkpoint does
 
@@ -162,7 +168,7 @@ persistence succeeded. If you set `checkpoint_on_close=False`:
 
 ## Failure and recovery
 
-On startup, NeuG opens only the manifest named by `CURRENT`; it does not scan for the newest directory. An incomplete staging manifest or object is unreachable until `CURRENT` changes and is therefore never selected. In Service mode, NeuG validates the selected WAL epoch and replays only records with timestamps greater than that manifest's `base_ts`.
+On startup, NeuG opens only the manifest named by `CURRENT`; it never falls back to a legacy directory when that selector exists. If `CURRENT` is absent, a read-write open may perform the one-time v1 migration described above. An incomplete staging manifest or object is unreachable until `CURRENT` changes and is therefore never selected. In Service mode, NeuG validates the selected WAL epoch and replays only records with timestamps greater than that manifest's `base_ts`.
 
 A **manual** `CHECKPOINT` fails in one of two ways:
 
--- a/doc/source/transaction/checkpoint.md
+++ b/doc/source/transaction/checkpoint.md
@@ -103,16 +103,17 @@ A persistent database uses one manifest selector, immutable objects shared by ma
 ```text
 data_dir/
 ├── checkpoint/
-│   ├── CURRENT                 # decimal ID of the selected manifest
+│   ├── CURRENT                 # decimal manifest ID + trailing newline
 │   ├── manifests/<id>.manifest
 │   └── objects/<object-id>
 ├── wal/<id>/                   # WAL epoch for manifest <id>
-└── runtime/open-<epoch>/       # temporary files for one database open
+└── runtime/open-<id>/          # temporary files for one database open;
+                                # <id> is an opaque unique suffix (a UUID)
 ```
 
-`CURRENT` is the sole publication selector. A published manifest has the required fields `v`, `id`, `base_ts`, `schema`, and `modules`; it may also contain `scalars`. Module descriptors persist object IDs, not absolute paths. The same ID names the manifest and its WAL epoch. `base_ts` is the highest transaction timestamp already represented by the manifest, so recovery replays the selected epoch from `base_ts + 1`. Full checkpoints use `base_ts=0` and reset the transaction timeline after reopening.
+`CURRENT` is the sole publication selector. Its content is the selected manifest's decimal ID followed by a single trailing newline (for example `3\n`), written via an atomic rename; operators may inspect or rewrite it manually with a plain text editor. A published manifest has the required fields `v`, `id`, `base_ts`, `schema`, and `modules`; it may also contain `scalars`. Module descriptors persist object IDs, not absolute paths. The same ID names the manifest and its WAL epoch. `base_ts` is the highest transaction timestamp already represented by the manifest, so recovery replays the selected epoch from `base_ts + 1`. Full checkpoints use `base_ts=0` and reset the transaction timeline after reopening.
 
-Checkpoint objects are immutable and may be referenced by several manifests. Runtime files are not checkpoint data: each database open receives its own `runtime/open-<epoch>/` directory, and closing that database removes only its own unpinned workspace. Existing `checkpoint-N` directories are unsupported and rejected without modification; this format does not migrate or read them. Recover or rebuild such a database with a version that supports the legacy format first.
+Checkpoint objects are immutable and may be referenced by several manifests. Runtime files are not checkpoint data: each database open receives its own `runtime/open-<id>/` directory (the suffix is an opaque unique ID, not a timestamp or manifest ID), and closing that database removes only its own unpinned workspace. Existing `checkpoint-N` directories are unsupported and rejected without modification; this format does not migrate or read them. Recover or rebuild such a database with a version that supports the legacy format first.
 
 ## What a checkpoint does
 
@@ -123,6 +124,12 @@ After publication, NeuG reopens the graph and allocators from the new checkpoint
 Recovery and shutdown checkpoints use the same compacting, destructive dump. Recovery reopens the graph and allocators before the database starts serving; shutdown persists without reopening. Garbage collection removes manifests, WAL epochs, and objects only when they are neither current nor retained by a live checkpoint reference.
 
 "Full" describes the runtime lifecycle boundary; it does not require rewriting every clean immutable object. Checkpoint disk growth is therefore driven by rewritten modules plus objects retained by live references. Schedule checkpoints based on the acceptable replay work after a crash and WAL growth, rather than a fixed tight interval.
+
+### Disk space reclamation
+
+Retired manifests, WAL epochs, immutable objects, and abandoned `runtime/open-<id>/` workspaces are removed by garbage collection, which runs only at three points: read-write database open, a successful manual `CHECKPOINT`, and database close. Deleting rows or dropping tables therefore does not shrink disk usage until one of those points is reached.
+
+Read-only opens never run garbage collection. A pure read-only deployment accumulates the stale `runtime/open-<id>/` workspaces left behind by crashed read-only processes; the next read-write open reclaims them.
 
 ### Concurrency
 
--- a/doc/source/transaction/checkpoint.md
+++ b/doc/source/transaction/checkpoint.md
@@ -9,8 +9,8 @@ plays a different role in each mode:
 | Question | Embedded mode | Service mode |
 |---|---|---|
 | Is `CHECKPOINT` required for durability? | Yes; un-checkpointed changes are lost when the database closes | No; committed changes are already durable in the WAL |
-| What is it for? | Persist changes and create a recovery point | Optional maintenance: fold WAL records into a fresh snapshot |
-| What is recovered after a restart? | The last successful checkpoint | The last checkpoint plus WAL replay |
+| What is it for? | Persist changes and create a recovery point | Optional maintenance: publish a checkpoint that bounds WAL replay |
+| What is recovered after a restart? | The checkpoint selected by `CURRENT` | The `CURRENT` checkpoint plus WAL records after its `base_ts` |
 
 For transaction boundaries and concurrency outside checkpoint operations, see
 [Transaction Management](transaction.md).
@@ -85,7 +85,7 @@ session.execute(
     access_mode="insert",
 )
 
-# Optional maintenance: fold committed WAL records into a fresh snapshot.
+# Optional maintenance: publish a checkpoint that bounds future WAL replay.
 session.execute("CHECKPOINT")
 session.close()
 ```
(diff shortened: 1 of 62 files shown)


longbinlai:
**Backward compatibility: consider auto-upgrading legacy `checkpoint-N` layouts instead of rejecting them**

The current behavior — `has_legacy_checkpoint()` throws `NOT_SUPPORTED` on any `checkpoint-*` directory — makes every existing database directory unusable after upgrade, forcing users to recover/re-export with an old binary first. That's a heavy migration cost for what is fundamentally a metadata/packaging change: the on-disk data files themselves (column dumps, index dumps) haven't changed, only the container (per-generation directories → objects + manifests).

I'd like us to attempt an automatic upgrade path before settling on hard rejection:

1. **On the first RW open of a legacy database** (we already hold the exclusive DB lock there):
   - Pick the highest-N `checkpoint-N` generation (the old "current" semantics).
   - Synthesize a `CheckpointManifest` (kFormatVersion 3) describing its existing files — move or hardlink them into `objects/`. They were immutable-once-written under the old model too, so hardlinks should be safe under the new immutability assumption; if that doesn't hold for some file kinds, fall back to copy.
   - Write `manifests/<id>.manifest`, then atomically publish `CURRENT`.
   - Handle the WAL: either replay legacy WAL records into the new checkpoint during the upgrade, or place them under `wal/<id>/` with `base_ts = 0`, matching the recovery boundary the new code already computes.
2. **Crash-safe by construction**: never delete old `checkpoint-N` directories until `CURRENT` is durably published; if we crash mid-upgrade, the next open retries. Old generations are then removed by the existing `CollectGarbage`.
3. **Fall back to an error only when we truly cannot interpret the data** (corrupt/partial state, unmappable old format) — the bar should be "there is no way to read this layout", not "we chose not to support it".

If after investigation the auto-upgrade is genuinely infeasible for some class of legacy layouts, then the incompatibility needs much stronger user-facing documentation than what's currently in the PR:

- a concrete migration recipe in the docs (old binary → open → export via `COPY TO` → new binary → import), not just "recover or rebuild such a database with a version that supports the legacy format first";
- the version number that introduces the format change called out in the release notes / upgrade guide, so users hit it *before* opening production data with the new binary.

The new test `RejectsLegacyCheckpointDirectoriesWithoutMutation` would then be inverted into upgrade-path coverage: legacy dir → first open auto-upgrades → reopen → data intact → old directories cleaned by GC.

This isn't blocking the core refactor — the object/manifest model itself looks right — but the zero-migration cliff is the single biggest user-facing cost of this change, so I'd like us to either close it or document our way through it explicitly.

doc/source/transaction/checkpoint.md
```diff
@@ -96,26 +96,33 @@ closed with `checkpoint_on_close=True`, any outstanding WAL records are
...
+A persistent database uses one manifest selector, immutable objects shared by manifests, a WAL epoch per manifest ID, and a temporary workspace per database open:
+
+```text
+data_dir/
+├── checkpoint/
+│   ├── CURRENT                 # decimal ID of the selected manifest
+│   ├── manifests/<id>.manifest
+│   └── objects/<object-id>
+├── wal/<id>/                   # WAL epoch for manifest <id>
+└── runtime/open-<epoch>/       # temporary files for one database open
```
longbinlai: Doc says `runtime/open-<epoch>/`, but the implementation names these workspaces with a UUID (`create_runtime_workspace` uses the UUID generator), while the tests call them "open epochs" (`OpenEpochWorkspacesAreDistinctAndIndependentlyOwned`). Users will `ls` this directory when debugging, so doc and code should agree — either describe it as an opaque unique suffix (`runtime/open-<id>/`) or align the naming in code/tests.

doc/source/transaction/checkpoint.md
```diff
@@ -96,26 +96,33 @@ closed with `checkpoint_on_close=True`, any outstanding WAL records are
...
+```
+
+`CURRENT` is the sole publication selector. A published manifest has the required fields `v`, `id`, `base_ts`, `schema`, and `modules`; it may also contain `scalars`. Module descriptors persist object IDs, not absolute paths. The same ID names the manifest and its WAL epoch. `base_ts` is the highest transaction timestamp already represented by the manifest, so recovery replays the selected epoch from `base_ts + 1`. Full checkpoints use `base_ts=0` and reset the transaction timeline after reopening.
```
longbinlai: Two operator-facing details missing from this section:

1. **When is space reclaimed?** GC runs only on RW open, manual `CHECKPOINT`, and close — deleting data won't shrink disk usage until one of those happens. Also, stale `runtime/open-*` workspaces left by crashed read-only processes are only cleaned up by the next RW open (a pure-RO deployment accumulates them). Worth a sentence each.
2. **The exact content of `CURRENT`** (bare decimal manifest ID? trailing newline?) — one sentence here would let operators inspect/repair it manually during incident response.

include/neug/storages/checkpoint_manifest.h
```diff
@@ -31,79 +32,53 @@ class Checkpoint;
...
-  static constexpr const char* kMetaFileName = "meta";
-
-  /// Current on-disk format version for the meta JSON.
+  /// Current on-disk format version for the manifest JSON.
   ///
   /// Bump only on breaking changes to the JSON layout (renamed/removed
   /// fields, changed value semantics).  Additive changes (new optional
   /// fields) do not require a bump.  Readers must reject unknown versions.
-  static constexpr int kFormatVersion = 1;
+  static constexpr int kFormatVersion = 3;
```
longbinlai: `kFormatVersion` jumps 1 → 3. If 2 was never written by any shipped build, taking 2 here would be the cleaner next number; if 2 does exist somewhere (branch builds, intermediate releases), a one-line comment explaining the skip would save future archaeology.

src/storages/graph/checkpoint_manager.cc
```diff
@@ -364,266 +178,296 @@ CheckpointManager::StagingCheckpoint::StagingCheckpoint(
...
+  }
+  if (publish_result != file_utils::AtomicFileWriter::CommitResult::kDurable) {
+    if (restore_current(selector, previous_id)) {
+      THROW_IO_EXCEPTION(
+          "CheckpointManager::Publish: CURRENT replacement durability failed "
+          "and was rolled back");
+    }
+    THROW_IO_EXCEPTION(
+        "CheckpointManager::Publish: CURRENT replacement durability and "
+        "rollback are commit-unknown");
```
longbinlai, line 353: This commit-unknown path (rename succeeded, fsync failed → keep the new CURRENT, throw) is the most safety-critical branch in publish, but I don't see a test reaching it: `FailedCurrentReplacementKeepsInMemoryCurrent` forces failure *before* the rename (CURRENT replaced by a directory), which exercises the rollback-confirmed path instead.

Consider a small fault-injection seam (e.g., an injectable fsync hook used only by tests) so this branch is executable in a test, or at least a comment on this line stating the branch is review-verified only and why.

tests/storage/test_checkpoint.cc
```diff
@@ -14,2504 +14,981 @@
...
+  meta->schema.property_name = "age";
+  meta->schema.property_type = schema->property_types[property_id];
+  auto* column = graph.get_vertex_table(label).GetPropertyColumnBase("age");
+  auto created = graph.mutable_index_manager().CreateIndex(
+      std::move(meta), std::make_unique<neug::DefaultIndexIDAccessor>(), column,
+      graph.GetVertexSet(label));
+  ASSERT_TRUE(created) << created.error().ToString();
+}
+
+TEST(CheckpointFormatTest, RejectsLegacyCheckpointDirectoriesWithoutMutation) {
```
longbinlai: The new suite covers the format invariants well. A few gaps versus the deleted suite worth closing before merge:

- **Memory-level coverage narrowed**: the old `TYPED_TEST` over `AllMemoryLevels` ran every scenario at both `kInMemory` and `kSyncToFile`; the new tests are mostly `kInMemory` with only `SyncToFileCheckpointRecoversAcrossOpenEpochs` covering `kSyncToFile`. The object-store path (hardlink vs copy, mmap'd objects) differs by level — please restore typed coverage for at least the core roundtrip + GC tests.
- **Data-level correctness is thinner**: `compact` and the old ~300-line `recover_from_checkpoint` scenario are only partially replaced; most new tests verify format/lifecycle invariants rather than recovered data content.
- **No multi-process coverage**: RO processes holding live checkpoints while an RW open runs GC (the pin guarantee), and two concurrent opens each creating `runtime/open-*` workspaces.
- **No explicit assertion** that `CollectGarbage` removes unreachable `wal/<id>/` epochs (runtime workspaces are covered by `GarbageCollectionReclaimsAbandonedOpenEpochs`, WAL epochs don't seem to have an equivalent).
- **No python/nodejs binding test changes** in the PR — the end-to-end open → write → close → reopen regression now depends entirely on pre-existing CI suites; a pointer to which existing test covers it would help.

tests/storage/test_ap_index.cc
```diff
@@ -217,9 +217,10 @@ class APIndexTest : public ::testing::Test {
   }
 
   void CheckpointGraph() {
-    auto staging = checkpoint_mgr_.CreateStagingCheckpoint();
+    auto staging = checkpoint_mgr_.CreateStaging();
+    graph_->Compact();
```
longbinlai, line 221: Adding `graph_->Compact()` before `DumpAndClear()` (same pattern in test_tp_index.cc and test_temporary_graph.cc) suggests the dump contract changed — previously `DumpAndClear` absorbed in-memory deltas itself. If `DumpAndClear` now requires a compacted graph, or no longer compacts implicitly, that is a contract change for out-of-tree `StorageIndex::Dump` implementers and deserves an explicit note in the headers (`checkpoint_file_manager.h` / PropertyGraph docs) and a line in the PR description.

include/neug/storages/checkpoint_manager.h
```diff
@@ -78,69 +58,40 @@ class NEUG_API CheckpointManager {
...
-  void CleanupRetiredCheckpoints();
-
-  std::string db_dir() const;
+  /// Return the checkpoint selected by CURRENT, or nullptr before first
+  /// publication.
+  std::shared_ptr<Checkpoint> Current() const;
+
+  /// Create an unpublished staging checkpoint. Destroying its move-only handle
+  /// discards it unless Publish() has made its manifest visible.
+  StagingCheckpoint CreateStaging();
```
longbinlai, line 75: These renames (`CreateStagingCheckpoint`→`CreateStaging`, `Commit`→`Publish`, `UpdateMeta`→`SetManifest`, `CurrentCheckpoint`→`Current`, plus snake→PascalCase across `CheckpointManifest`) are internal APIs but they break out-of-tree extensions that touch checkpoint APIs — this PR already had to adapt `tests/extension/vec_index`. Please include the full rename list in the PR description/changelog so extension authors can migrate mechanically.

longbinlai (commented):
Review notes on docs, test coverage, and API impact (design itself looks solid — the publish sequence, GC reachability model, and WAL epoch boundary all check out against the code).

doc/source/transaction/checkpoint.md
zhanglei1949: Good catch — the code names workspaces `open-<UUID>` (`UUIDGenerator::Generate()` in `create_runtime_workspace`), so `<epoch>` was misleading. Fixed in 30125704: the layout now reads `runtime/open-<id>/` and explicitly notes the suffix is an opaque unique ID (a UUID), not a timestamp or manifest ID.

doc/source/transaction/checkpoint.md
zhanglei1949: Both added in 30125704. A new "Disk space reclamation" section documents the three GC points (RW open, manual `CHECKPOINT`, close), that deleting rows or dropping tables does not shrink disk usage until one of them runs, and that RO opens never run GC — so a pure-RO deployment accumulates stale `runtime/open-<id>/` workspaces from crashed RO processes until the next RW open. The `CURRENT` paragraph now states its exact content: the selected manifest's decimal ID plus a single trailing newline (e.g. `3\n`), written via atomic rename, so operators can inspect or rewrite it manually.

include/neug/storages/checkpoint_manifest.h
zhanglei1949: OK.

src/storages/graph/checkpoint_manager.cc
zhanglei1949: Fair point — `FailedCurrentReplacementKeepsInMemoryCurrent` only exercises the pre-rename failure path (rollback confirmed), and there is no seam to fail the parent-directory fsync *after* a successful rename without a fault-injection hook inside `AtomicFileWriter`. Went with the comment option in 30125704: the branch now documents why the new CURRENT is deliberately kept when rollback cannot be confirmed (it names a manifest that was fully persisted above, strictly safer than an unverified rollback) and states the branch is review-verified rather than test-covered. Happy to add an injectable fsync hook as a follow-up if you think the seam is worth carrying.

tests/storage/test_checkpoint.cc
zhanglei1949: Partially addressed in 30125704, plus two clarifications:

- **Memory-level coverage — restored**: new `CheckpointRoundtripTest` runs the core roundtrip (`ManualCheckpointRoundTripsEvolvedVertexAndEdgeDescriptors`), `ManualCheckpointPreservesDeletedVertexState`, and a new `ReopenReclaimsRetiredCheckpointsAndWalEpochs` at both `kInMemory` and `kSyncToFile` via an `AllMemoryLevels` instantiation.
- **WAL epoch GC — already covered, easy to miss**: `CurrentSelectsManifestAndGcRespectsCheckpointPins` asserts `wal/0` survives GC while the checkpoint is pinned and is removed once unpinned (the test name only mentions manifest/pins). The new reopen-GC test additionally asserts end-to-end that only the current manifest and its WAL epoch remain after a reclaiming reopen, at both memory levels.
- **Data-level scenarios — partially agree**: the old 410-line `recover_from_checkpoint` and `compact` scenarios were folded into narrower data-asserting tests (deleted-vertex state, drop/recreate, descriptor evolution, index object reuse). Can add a dedicated multi-round write → checkpoint → verify scenario if you want more depth.
- **Multi-process coverage — agreed, filed as follow-up**: absent before this PR too; RO pin + concurrent RW GC and concurrent opens need a multi-process harness, proposed as follow-up work rather than a merge blocker.
- **Bindings — Python did change**: `tools/python_bind/tests/test_tp_service.py` in this PR asserts `CURRENT == f"{id}\n"`, the manifests/objects layout, WAL rotation across CHECKPOINTs, and plan-cache invalidation after label compaction (the open → write → checkpoint → reopen regression lives there). Node.js does not touch checkpoint internals, so it is unchanged.

tests/storage/test_ap_index.cc
zhanglei1949: Checked the premise against both revisions: the dump contract did not change. `DumpAndClear` never compacted implicitly — old and new implementations differ only in renames (`LinkToSnapshot`→`ReuseCheckpointModules`, `UpdateMeta`→`SetManifest`, `GetMeta`→`manifest`) — and the production path already called `Compact()` immediately before `DumpAndClear` in `CheckpointCoordinator` (unchanged by this PR; see the identical sequence on main). The test helpers adding `graph_->Compact()` align the tests with that production sequence rather than reflecting a new requirement, so out-of-tree `StorageIndex::Dump` implementers need no migration. That said, the precondition was undocumented, so 30125704 states it on the `DumpAndClear` declaration in `property_graph.h`.

include/neug/storages/checkpoint_manager.h
zhanglei1949: Done — the PR description now has an "API migration notes for out-of-tree extensions" section with three tables covering the `CheckpointManager`/`StagingCheckpoint`, `CheckpointManifest`, and `Checkpoint` renames plus removals (`mutable_modules()`, `remove_module`, `RestoreCurrentCheckpoint`, the `Cleanup*` pair), with the vec_index adaptation called out as the typical migration.

longbinlai (dismissed):
LGTM

include/neug/storages/checkpoint.h
```diff
@@ -121,55 +83,60 @@ class Checkpoint {
...
-  std::unique_ptr<CheckpointManifest> meta_;
+  void initialize(bool load_manifest);
+  void create_dirs() const;
+  void resolve_object_paths();
+  void persist_manifest();
+
+  std::string database_dir_;
+  std::string manifest_path_;
+  std::string object_dir_;
+  std::shared_ptr<const std::string> runtime_workspace_;
```
liulx20: 这个东西是否需要封装一下呢，感觉看起来语义上有点奇怪

zhanglei1949: https://github.com/alibaba/neug/pull/892/commits/fd42c1c9541f4daf616cf9ae339738aa43bc0b31

liulx20 (approved):

PR #898, opened 2026-08-19
refactor: decouple S3/HTTP IO from Arrow via a curl-based client in httpfs

### LINKED ISSUE #872: Remove Arrow S3 dependency in httpfs extension

The S3/OSS support of the httpfs extension is currently built entirely on `arrow::fs::S3FileSystem`. Enabling it requires `ARROW_S3=ON`, which makes Arrow statically bundle the whole AWS toolchain into `arrow_bundled_dependencies.ar`:

- aws-sdk-cpp (s3 / core) and the aws-c-* libraries
- aws-lc / BoringSSL

This causes:

1. Significantly larger extension artifacts and longer build time;
2. TLS symbol conflicts between Arrow's bundled aws-lc/BoringSSL and the system OpenSSL 3.0, which currently requires a "prefer static libcurl + static OpenSSL" workaround (the first HTTPS request segfaults inside `OPENSSL_sk_value()`);
3. Extra CURL detection logic in the Arrow build; the build fails when libcurl dev headers are missing (issue #480-type errors).

In addition, the HTTP part of httpfs (`HTTPFileSystem` / `HTTPRandomAccessFile`) directly inherits `arrow::fs::FileSystem` / `arrow::io::RandomAccessFile`, so the extension also depends on the base Arrow library.

A full decoupling in three layers:

1. **Core VFS**: add an Arrow-agnostic remote IO abstraction (`fsys::RemoteFileSystem` / `fsys::RandomAccessStream`) to replace the opaque Arrow handle contract of `getArrowFileSystem()`; the core library itself introduces no Arrow dependency.
2. **httpfs extension**: become Arrow-free. Reimplement the S3 protocol directly with libcurl (similar to DuckDB's httpfs approach), replacing `arrow::fs::S3FileSystem`, and set `ARROW_ENABLE_S3=OFF`; `HTTPFileSystem` / `HTTPRandomAccessFile` no longer inherit Arrow classes, and the extension links no Arrow library at all. The write path (parquet export to s3/oss) must be kept.
3. **parquet extension**: add an Arrow adapter layer that wraps `fsys::RemoteFileSystem` into `arrow::fs::FileSystem` (read: `arrow::io::RandomAccessFile`; write: `arrow::io::OutputStream`), so `ArrowReader` / `ExportWriter` stay unchanged.

Expected gains: drop the aws-sdk-cpp toolchain, make httpfs completely Arrow-free, simplify the TLS workaround, and remove the CURL detection logic from the Arrow build.

1. Partial decoupling only: keep exposing the new S3 implementation as an `arrow::fs::FileSystem` subclass (httpfs still links the base Arrow library). Simpler, but the Arrow dependency remains in httpfs;
2. Drop S3/OSS support entirely: minimal change, but unacceptable since OSS read/write is required;
3. Keep the status quo: the AWS SDK size and TLS conflict issues remain.

**Additional context**

- Files involved: core VFS (`include/neug/utils/io/vfs/file_system.h`, `src/utils/io/vfs/file_system.cc`), httpfs extension (`extension/httpfs/**`), parquet extension (`extension/parquet/src/arrow_fs_resolver.cc` + new adapter), `cmake/BuildArrowAsThirdParty.cmake`, root `CMakeLists.txt` (`ARROW_ENABLE_S3` switch)
- Note: the virtual-interface change of `fsys::FileSystem` is a breaking change (vtable); only httpfs/parquet use this interface (gds/pattern_matching/test_out_of_tree are unaffected), and the core CSV/JSON loading path only uses `glob()` and is unaffected

BingqingLyu:
## What do these changes do?

Fully decouple the httpfs extension (S3/OSS/HTTP IO) from Arrow's filesystem
stack and replace it with a self-contained curl-based client, then fix a series
of correctness issues in the remote read/write paths that were exposed by the
migration.

### Core refactor

- New curl-based `S3Client` with AWS SigV4 signing (`s3_sigv4`, covered by
  golden-vector tests): GET/HEAD/PUT plus S3 multipart upload
  (Create/UploadPart/Complete/Abort).
- `HTTPFileSystem` / `S3FileSystem` now implement the NeuG VFS interfaces
  (`fsys::RandomAccessStream` / `fsys::OutputStream`) directly; `arrow::fs` is
  no longer used by the extension.
- `ARROW_S3` is forced OFF in the bundled Arrow build; the retired
  `ARROW_ENABLE_S3` flag now warns and is ignored instead of being silently
  accepted.
- The parquet extension reads remote files through new Arrow stream adapters
  (`arrow_fs_adapter`) backed by the VFS streams, and the AWS-SDK / static-cURL
  linker workarounds were removed.

### Remote reads and exports

- `LOAD FROM` for remote CSV/JSON now streams through `io::InputStream`
  instead of downloading whole files first.
- CSV/JSON/JSONL export is routed through the VFS so `oss://` / `s3://`
  targets work; the output stream is opened lazily after validation and after
  the query results are sunk, so a failed export never truncates the target.
- Failed remote exports Abort the stream (the S3 multipart upload is aborted)
  instead of finalizing a partial object.

### Testing

- SigV4 golden vectors, offline option/filesystem unit tests, a loopback
  no-Range-support HTTP server test, parquet adapter tests, stream
  error-propagation tests, lazy-open/abort writer tests, and empty-file
  regression tests.
- Verified: utils tests (290), httpfs extension tests (45), parquet extension
  tests (42), Python load/export regression, CI-style `simple_example.py` /
  `complex_test.py` (55/55) against the local build.

## Migration notes (behavior changes vs. the Arrow-based implementation)

- **Credential resolution is simplified.** Credentials are resolved only from
  query options and environment variables (`OSS_ACCESS_KEY_ID` /
  `AWS_ACCESS_KEY_ID`, `OSS_ACCESS_KEY_SECRET` / `AWS_SECRET_ACCESS_KEY`).
  The Arrow default provider chain is gone: `~/.aws/credentials` files, STS
  tokens, and EC2/ECS IAM roles are **not supported**. With
  `CREDENTIALS_KIND=Default` and no credentials found, the query now fails
  with a clear error instead of silently downgrading to anonymous access.
- **`ARROW_ENABLE_S3` is retired** and only warns if passed; `ARROW_S3` is
  forced OFF in the bundled Arrow build. No build flag is needed anymore to
  enable S3 support — it ships with the httpfs extension.
- **HTTP/HTTPS sources are read-only** and must support `HEAD` (size probe)
  and `Range` requests; servers that ignore `Range` and answer `200` are
  rejected with an error instead of being silently misread.
- **New options:** `VERIFY_SSL`, `CA_CERT_FILE`, `PATH_STYLE` (S3/OSS);
  see `doc/source/extensions/load_httpfs.md`.
- **Remote exports (`COPY TO`)** now use multipart uploads for objects larger
  than 16 MiB and abort the multipart upload on failure, so a failed export
  never leaves a truncated/partial object.

## Related issue number

Fixes https://github.com/alibaba/neug/issues/872


shirly121 (dismissed):
Thanks for the substantial VFS refactor. Decoupling httpfs from Arrow and adding lazy-open/abort behavior are good improvements. I found several issues that should be addressed before merging.

## Required changes

### 1. Remote JSON array input is not streaming and can consume multiple times the object size in memory

`parse_json_array_file()` calls `read_stream_to_string()`, then builds a complete RapidJSON DOM. Thus a remote JSON array retains both the whole source string and the parsed document. `BATCH_READ`/`BATCH_SIZE` does not bound this memory usage.

Suggested fix: parse array elements incrementally with a SAX/iterative parser and emit chunks as elements arrive. If that is intentionally out of scope, reject/document streaming mode for JSON arrays and make the limitation explicit in the user-facing docs and option descriptions. Add a large-input peak-memory regression test.

### 2. Remote CSV and JSONL perform an avoidable full counting pass before parsing

`JsonChunkSupplier` opens the stream once to count every non-empty line, then opens it again to parse. The PR description indicates the remote CSV path similarly performs a streamed row-count pass. This doubles transferred bytes and requests, and the two passes can observe different content if a URL/object changes between reads.

Suggested fix: allow suppliers to start with an unknown row count and update it while parsing, or make row count lazy/optional. If two passes are unavoidable, pin and validate an object version/ETag between passes. Add a test using a counting stream opener asserting a single data pass.

### 3. Remote JSONL does not normalize CRLF line endings

`StreamLineReader::nextLine()` removes `\n` but leaves a trailing `\r`, unlike `io::readFirstLine()`. JSONL produced with CRLF can therefore fail parsing on the remote stream path or behave inconsistently with local/header reads.

Suggested fix: strip one trailing `\r` before returning a line, both when `\n` is found and when EOF returns the final line. Add CRLF tests for normal rows and a final row without `\n`.

### 4. Filesystem scheme lookup is incorrectly case-sensitive

`FileSystemRegistry::Provide()` extracts the URI scheme verbatim, while providers are registered under lowercase names. URI schemes are case-insensitive, so paths such as `HTTPS://...` and `S3://...` currently fail with `Unsupported file system protocol`.

Suggested fix: normalize schemes to lowercase in both `Register()` and `Provide()`, and add mixed/uppercase scheme tests.

## API/behavior clarifications and recommended follow-ups

- `BATCH_SIZE` does not have a unified meaning: CSV/JSON treat it as rows (capped at 65,536), while Parquet treats it as an I/O buffer size in bytes; Parquet row batching is separately controlled by `PARQUET_BATCH_ROWS`. Please split or rename this option (for example `BATCH_ROWS` and `IO_BUFFER_SIZE`) or at minimum document the format-specific semantics clearly.
- The current CSV/JSON `batch_read()` appends every produced chunk to the same output `Context`, and filters force `full_read()`/`read_all_chunks()`. Please verify with a peak-memory test that this is genuinely end-to-end bounded-memory streaming; otherwise describe it as chunked parsing rather than streaming.
- HTTP/HTTPS is read-only, has no glob expansion, and requires successful HEAD plus Range support. The support matrix should explicitly distinguish read from write: local and S3/OSS support both, while HTTP/HTTPS does not support export.
- Remote Parquet requires both the `parquet` and `httpfs` extensions. When httpfs has not been loaded, HTTP/S3/OSS currently produces only `Unsupported file system protocol: <scheme>`. Please consider an actionable error suggesting `LOAD EXTENSION httpfs` and add an unloaded-extension end-to-end test.

These points do not undermine the VFS direction, but the first four are correctness/performance-contract issues in the newly introduced paths and should be resolved before merge.

extension/httpfs/src/http_filesystem.cc
```diff
@@ -131,456 +124,409 @@ size_t WriteCallbackDirect(void* contents, size_t size, size_t nmemb,
...
   }
 
-  // Handle zero-length read
-  if (nbytes == 0) {
-    return std::make_shared<arrow::Buffer>(nullptr, 0);
+  result<std::shared_ptr<fsys::OutputStream>> openOutputStream(
+      const std::string& path) override {
+    RETURN_STATUS_ERROR(
+        neug::StatusCode::ERR_NOT_SUPPORTED,
+        "Writing over HTTP is not supported (read-only filesystem): " + path);
```
shirly121, line 601: HTTP/HTTPS output is explicitly unsupported here, while CSV/JSON/JSONL and Parquet exports now route through the VFS. If HTTP is intentionally read-only, could we document this in a protocol/format capability matrix so users do not interpret unified VFS support as read+write support for every protocol? It would also help to add an export test that asserts this actionable `ERR_NOT_SUPPORTED` result.

shirly121 (commented):
A few non-blocking review suggestions around the HTTP/VFS capability boundary. One additional issue cannot be attached inline because the relevant registry lookup line is unchanged by this PR: `FileSystemRegistry::Provide()` extracts the scheme verbatim and looks it up against lowercase registrations, so URI schemes such as `HTTPS://` or `S3://` fail even though URI schemes are case-insensitive ([code](https://github.com/alibaba/neug/blob/04600e2081c8b30858ee05f58cc3293a71078bdc/src/utils/io/vfs/file_system.cc#L63-L87)). Suggest normalizing to lowercase in both `Register()` and `Provide()`, with mixed-case tests.

src/utils/io/read/json/json_reader.cc
```diff
@@ -200,14 +228,74 @@ Value parse_json_value(const rapidjson::Value& value,
...
+        len_ = static_cast<size_t>(*r);
+      }
+      const char* begin = buffer_ + pos_;
+      const size_t avail = len_ - pos_;
+      const char* nl =
+          static_cast<const char*>(std::memchr(begin, '\n', avail));
+      if (nl) {
+        line.append(begin, static_cast<size_t>(nl - begin));
+        pos_ += static_cast<size_t>(nl - begin) + 1;
+        return true;
```
shirly121, line 273: `StreamLineReader` strips `\n` but leaves a trailing `\r`, unlike the local `std::getline` path and `io::readFirstLine()`. Remote JSONL with CRLF may therefore parse inconsistently. Suggest removing one trailing `\r` before both successful return paths (newline found and EOF with a final line), plus tests for CRLF and a final line without `\n`.

extension/httpfs/src/http_filesystem.cc
```diff
@@ -131,456 +124,409 @@ size_t WriteCallbackDirect(void* contents, size_t size, size_t nmemb,
...
-
-    auto proxy_user_it =
-        options_.find(HTTPConfigOptionKeys::kHTTPProxyUsername);
-    auto proxy_pass_it =
-        options_.find(HTTPConfigOptionKeys::kHTTPProxyPassword);
-    if (proxy_user_it != options_.end() && proxy_pass_it != options_.end()) {
-      std::string userpass =
-          proxy_user_it->second + ":" + proxy_pass_it->second;
-      curl_easy_setopt(curl, CURLOPT_PROXYUSERPWD, userpass.c_str());
+    if (response_code == 200) {
```
shirly121, line 434: Failing instead of silently misreading a server that ignores Range is correct. However, HTTP support now assumes both a successful HEAD request (during size probing) and 206 Range responses. Many download endpoints/CDNs do not support HEAD or Range consistently. If this is an intentional requirement, please document it prominently and test representative failure messages; otherwise consider a sequential-GET fallback for offset 0/non-random-access readers and deriving size from `Content-Range`/GET when HEAD is unavailable.

extension/httpfs/src/s3_client.cc
```diff
@@ -0,0 +1,667 @@
...
+  if (!resp) {
+    return tl::unexpected(resp.error());
+  }
+  return {};
+}
+
+result<std::string> S3Client::createMultipartUpload(
+    const std::string& bucket, const std::string& key) const {
+  auto resp =
+      request("POST", bucket, key, {}, {}, nullptr, 0, true, nullptr, 0);
```
longbinlai: 这里发出的请求是 `POST /bucket/key`（query 为空），但 S3/OSS 的 CreateMultipartUpload API 需要 `?uploads` 子资源（`POST /{key}?uploads`）。缺了它，这次 POST 会被服务端当作普通 POST Object 拒绝——效果是：`S3OutputStream` 攒满 2 × part_size（默认 16 MiB）之后的远程写入（`COPY TO` s3:// / oss:// 导出大结果集）会全部失败。

建议补上：

```cpp
std::vector<std::pair<std::string, std::string>> query = {{"uploads", ""}};
auto resp = request("POST", bucket, key, query, {}, nullptr, 0, true, nullptr, 0);
```

SigV4 canonical query 里 `uploads=` 正是 AWS 对无值子资源的约定写法，现有签名逻辑不用改。

extension/httpfs/src/s3_options.cc
```diff
@@ -287,192 +324,111 @@ std::string S3OptionsBuilder::resolveEndpoint() const {
...
-    // resolveCredentialsKind()
+  if (config.hasCredentials()) {
+    LOG(INFO) << "Configured credentials (access_key: "
+              << maskCredential(config.access_key) << ")";
+  } else {
+    // Fail loudly instead of silently downgrading to anonymous: earlier
+    // versions resolved Default through ~/.aws/credentials and IAM/ECS
+    // roles, which this build does not support. A silent anonymous
+    // fallback would turn those deployments into hard-to-diagnose 403s.
     THROW_INVALID_ARGUMENT_EXCEPTION(
```
longbinlai, line 418: 这个行为变化建议在 PR 描述里显式声明：之前（Arrow S3）Default 模式走 AWS SDK 默认凭证链，`~/.aws/credentials`、EC2 instance metadata（IAM role）、ECS task role 都可用；无凭证时最终退化为匿名请求，公开桶裸读是能工作的。现在这两条路都断了——前者直接抛错，后者必须显式 `CREDENTIALS_KIND='Anonymous'`。

"宁可大声失败也不静默降级"的取舍我赞成（报错文案也写得清楚），但这是用户可见的 breaking change：靠 IAM role / `~/.aws/credentials` 的部署升级后查询会直接报错。建议在 PR 描述里加一条 migration note，说明需要迁移的用法。

extension/httpfs/include/s3_options.h
```diff
@@ -128,46 +102,63 @@ struct S3ParseOptions {
...
- *   - NeuG does NOT interfere with this chain
- *
- * When CREDENTIALS_KIND=Anonymous:
- *   - No credentials used (for public buckets only)
- *
- * OSS-Specific Handling (automatic when OSS endpoint detected):
- *   - force_virtual_addressing = true (OSS requires virtual hosted-style)
- *   - Auto-detect region from endpoint (e.g., oss-cn-hangzhou.aliyuncs.com ->
- * oss-cn-hangzhou)
+ * Credential resolution (simplified; no STS / IAM role / default chain):
```
longbinlai, line 153: 本 PR 改了 51 个文件但没有同步任何文档，而 `doc/source/extensions/load_httpfs.md` 的 "Credential Modes → Default" 章节明确写着支持 `~/.aws/credentials` → EC2 IAM role → ECS task role 这条链，文档里的 `COPY TO` 示例还在用 `CREDENTIALS_KIND='Default'`——合并后这页文档就与实现相悖了。

建议随本 PR 一并重写该章节，顺带记录新增的 `VERIFY_SSL` / `CA_CERT_FILE` / `PATH_STYLE` 选项。两个 tutorial 用的都是 Anonymous/Explicit，不受影响。

extension/httpfs/tests/s3_test.cc
```diff
@@ -103,859 +90,394 @@ TEST(S3URIParserTest, ParseURIWithGlobPattern) {
...
-  if (oss_access_key_.empty() || oss_secret_key_.empty()) {
-    GTEST_SKIP() << "OSS credentials not configured. "
-                 << "Set OSS_ACCESS_KEY_ID and OSS_ACCESS_KEY_SECRET "
-                    "environment variables to run this test.";
+TEST_F(S3IntegrationTest, WriteListReadRoundTrip) {
+  // Generate a deterministic pseudo-random payload (~1 MiB).
+  std::string payload;
+  payload.reserve(1 << 20);
+  std::mt19937_64 rng(42);
+  while (payload.size() < (1u << 20)) {
```
longbinlai, line 460: 这个 1 MiB payload 低于 multipart 触发阈值（2 × 8 MiB），本用例实际只走了 PutObject 单发路径——createMultipartUpload / uploadPart / completeMultipartUpload / abort 都没有被执行到（这也是 `?uploads` 那个问题没被发现的原因）。

建议：
1. 补一个 > 16 MiB 的写入用例（可以同样放在有凭证才跑的集成测试里），让 multipart 全链路至少被集成测试覆盖；
2. 目前所有 S3 集成测试在 CI 里都会被 `GTEST_SKIP` 跳过（workflow 未配置 OSS 凭证），签名请求的真实组装（virtual-host / path-style、ETag 回传、ListObjectsV2 翻页）以及全部写路径等于零自动化覆盖；CI 里唯一真实的远程用例是公开 OSS 的 HTTP parquet（`test_copy_temp.py`），覆盖不到 `s3://` 签名路径。可以考虑在 CI 里起一个 MinIO 容器，把签名 + multipart 变成常规测试，不依赖真实云凭证。

extension/httpfs/src/s3_client.cc
```diff
@@ -0,0 +1,667 @@
...
+    if (attempt > 0) {
+      int backoff_ms = 100 * (1 << (attempt - 1));  // 100, 200, 400, ...
+      LOG(WARNING) << "S3 request retry " << attempt << "/"
+                   << (max_attempts - 1) << " after " << backoff_ms
+                   << "ms: " << method << " " << url
+                   << " (last error: " << last_error << ")";
+      std::this_thread::sleep_for(std::chrono::milliseconds(backoff_ms));
+    }
+
+    CURL* curl = curl_easy_init();
```
longbinlai, line 316: 每次 `request()` 都新起一个 CURL handle，意味着每次 ReadAt 都是一次全新的 TCP + TLS 握手——没有连接复用，也没有 readahead。parquet 读一个文件是 footer + 每个列 chunk 各一次 ReadAt，几十列的文件就是几十次完整握手；对比原来 Arrow S3FileSystem 的连接池 + 预读，远程 parquet 读取会有可感知的性能回退（`http_filesystem.cc` 的 ReadAt 同样如此）。

建议：用 `curl_share_init`（CURLSH）共享 DNS / 连接缓存，或在 stream 层加一层 readahead 缓冲（按 4-8 MiB 对齐做块读取），两者结合能把小读放大的问题压下来。

longbinlai (commented):
整体方向很好——把 httpfs 从 Arrow 的 S3 栈里解耦出来、构建上去掉整个 AWS 工具链正是 #872 要的；SigV4 golden vector 测试、导出懒打开与 Abort 语义也都做得很扎实。

几个关注点（详见 inline comments）：

1. `createMultipartUpload` 缺 `?uploads` 子资源，≥16 MiB 的远程写入会失败，且现有测试恰好覆盖不到这条路径；
2. Default 凭证模式不再支持 `~/.aws/credentials` / IAM / ECS role，无凭证时直接抛错，是用户可见的 breaking change，建议在 PR 描述加 migration note；
3. 文档未同步：`load_httpfs.md` 的 Default 凭证章节在合并后会与实现相悖；
4. multipart 与真实 S3 签名路径目前没有任何自动化覆盖（集成测试在 CI 全部 skip），建议引入 MinIO 集成测试；
5. 每次读新建 CURL handle（每读一次完整 TLS 握手）且无 readahead，远程 parquet 读取性能会回退。

其余问题（远程 CSV/JSON 双遍下载、JSON 内存驻留、JSONL CRLF、scheme 大小写等）其他 reviewer 已提出，不在此重复。

extension/httpfs/tests/s3_test.cc
longbinlai: 如果 16mb 比较大耗时较长可以和 @lnfjpt 商量放到 weekly test

extension/httpfs/include/glob_utils.h
```diff
@@ -16,13 +16,8 @@
 
 #pragma once
```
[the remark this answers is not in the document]
BingqingLyu: Fixed in bdefbcb6: added `#include <algorithm>` to `glob_utils.h`.

extension/httpfs/include/glob_utils.h
```diff
@@ -48,86 +43,35 @@ inline bool MatchGlobPattern(const std::string& text,
...
-
-  if (wildcard_pos != std::string::npos) {
-    // Find last '/' before wildcard
-    size_t last_slash = base_dir.rfind('/', wildcard_pos);
-    if (last_slash != std::string::npos) {
-      base_dir = base_dir.substr(0, last_slash);
-    } else {
-      base_dir = "";
-    }
+      std::min({pattern.find('*'), pattern.find('?'), pattern.find('[')});
```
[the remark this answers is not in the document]
BingqingLyu: Same fix as above — both lines are in the same header.

extension/parquet/src/arrow_fs_adapter.cc
```diff
@@ -0,0 +1,292 @@
...
+  if (closed_) {
+    return arrow::Status::OK();
+  }
+  closed_ = true;
+  auto r = stream_->Close();
+  if (!r.has_value()) {
+    return ToArrowStatus(r.error());
+  }
+  return arrow::Status::OK();
+}
```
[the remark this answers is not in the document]
BingqingLyu: Fixed in bdefbcb6: `StreamOutputStream::Close()` now calls `stream_->Abort()` when `failed_` is true and returns an `IOError`. Added `ParquetAdapterTest.ExplicitCloseAfterWriteFailureAborts` regression test.

extension/httpfs/src/s3_sigv4.cc
```diff
@@ -0,0 +1,285 @@
...
+
+  std::time_t sign_time = (timestamp != 0) ? timestamp : std::time(nullptr);
+  signed_req.amz_date = Iso8601BasicFormat(sign_time);
+  signed_req.date_stamp = signed_req.amz_date.substr(0, 8);
+
+  // Collect all headers to sign: host + x-amz-date + x-amz-content-sha256
+  // + caller supplied extras. Lowercase names, trim/collapse values.
+  std::vector<std::pair<std::string, std::string>> canonical_headers;
+  canonical_headers.emplace_back("host", request.host);
+  canonical_headers.emplace_back("x-amz-content-sha256", request.payload_hash);
```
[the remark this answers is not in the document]
BingqingLyu: Fixed in bdefbcb6: the `payload_hash` is now resolved at the top of `SignSigV4` (after `amz_date`), and the same resolved value is used for both the `x-amz-content-sha256` header and the canonical request. Added `SigV4Test.EmptyPayloadHashResolvedConsistently`.

extension/httpfs/src/s3_options.cc
```diff
@@ -76,63 +76,28 @@ std::string resolveTlsCaFilePath() {
...
-  // Fallback to a directory next to the chosen ca-bundle, if it exists.
-  struct stat st;
-  if (::stat("/etc/ssl/certs", &st) == 0 && S_ISDIR(st.st_mode)) {
-    return "/etc/ssl/certs";
-  }
-  return "";
+std::string toLowerStr(std::string s) {
+  std::transform(s.begin(), s.end(), s.begin(), ::tolower);
+  return s;
 }
```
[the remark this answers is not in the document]
BingqingLyu: Fixed in bdefbcb6: `toLowerStr` now casts to `unsigned char` before calling `std::tolower`.

extension/httpfs/src/s3_client.cc
```diff
@@ -0,0 +1,657 @@
...
+                                    const std::string& key) const {
+  auto resp =
+      request("HEAD", bucket, key, {}, {}, nullptr, 0, false, nullptr, 0);
+  if (!resp) {
+    // Treat 404 as "not found"; propagate other errors.
+    if (resp.error().error_message().find("HTTP 404") != std::string::npos) {
+      return false;
+    }
+    return tl::unexpected(resp.error());
+  }
```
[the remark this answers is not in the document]
BingqingLyu: Fixed in bdefbcb6: `request()` now takes a `long* out_http_code` output parameter, and `objectExists()` branches on `http_code == 404` instead of parsing the error message.

extension/httpfs/include/s3_options.h
```diff
@@ -85,23 +58,24 @@ struct S3ConfigOptionKeys {
...
   static constexpr const char* kAnonymous =
       "anonymous";  // No credentials (public buckets)
   static constexpr const char* kDefault =
-      "default";  // Arrow's default chain (env/config/role)
-  static constexpr const char* kRole =
-      "role";  // Assume role (not yet supported)
-  static constexpr const char* kWebIdentity =
-      "webidentity";  // Web identity token (not yet supported)
+      "default";  // options -> environment -> anonymous
 };
```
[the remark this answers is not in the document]
BingqingLyu: Fixed in bdefbcb6: the comment for `kExplicit` now says 'credentials from options only'.

extension/httpfs/src/http_filesystem.cc
BingqingLyu: Done. `openOutputStream` now has a comment explaining HTTP is read-only, and `RemoteFileSystem_OpenOutputStream_NotSupported` already asserts `ERR_NOT_SUPPORTED`. The capability matrix is documented in `load_httpfs.md` (HTTP/HTTPS section explicitly states read-only + HEAD/Range requirements).

src/utils/io/read/json/json_reader.cc
BingqingLyu: Fixed: added `stripTrailingCR()` applied in both `StreamLineReader::nextLine` return paths (newline found + EOF with final line), the local `std::getline` counting loop, `GetNextChunk`, and schema sniff. Added 3 tests: `TestJsonLinesCRLF` (local), `TestJsonLinesCRLFStreaming` (stream), `TestJsonLinesLFUnchanged` (regression guard).

extension/httpfs/src/http_filesystem.cc
BingqingLyu: Done. `HTTPRandomAccessStream` class comment now documents HEAD + Range as mandatory server requirements, and the 200-rejection path emits 'Server ignored the Range request (returned 200); Range support is required: <url>'. `ReadAtRejectsServerWithoutRangeSupport` now asserts this message. `load_httpfs.md` HTTP section documents the requirement prominently.

extension/httpfs/src/s3_client.cc
BingqingLyu: Fixed: `createMultipartUpload` now sends `query = {{"uploads", ""}}`. Added `MultipartWriteAboveThresholdRoundTrip` (20 MiB payload, credential-gated) that exercises the full CreateMultipartUpload → UploadPart → CompleteMultipartUpload path.

extension/httpfs/src/s3_options.cc
BingqingLyu: Added a 'Migration notes' section to the PR description covering: credential resolution simplified (no more `~/.aws/credentials` / IAM role / STS), `ARROW_ENABLE_S3` retired, HTTP read-only + HEAD/Range requirement, new options, multipart export behavior.

extension/httpfs/include/s3_options.h
BingqingLyu: Rewrote `doc/source/extensions/load_httpfs.md`: Credential Modes section now documents the actual resolution order (options → env → error), added TLS/Addressing options table (VERIFY_SSL / CA_CERT_FILE / PATH_STYLE), HTTP section documents read-only + HEAD/Range requirements, Export section documents multipart threshold.

extension/httpfs/tests/s3_test.cc
BingqingLyu: Added `MultipartWriteAboveThresholdRoundTrip` (20 MiB, >16 MiB threshold) that exercises the full multipart path. MinIO CI for integration tests is tracked as a follow-up.

extension/httpfs/src/s3_client.cc
BingqingLyu: Done. Added `remote_io_utils.h/.cc` with a thread-safe `CurlShare` wrapper (DNS/TLS/TCP connection cache) and a `ReadaheadCache` (4 MiB block). S3Client and HTTPFileSystem now share one `CurlShare` across all requests. S3/HTTP `RandomAccessStream` serve small sequential reads from the readahead block. Added `ReadaheadAndConnectionReuse` loopback test that verifies 3 requests (HEAD + 2 GET) share 1 TCP connection.

extension/httpfs/src/http_filesystem.cc
BingqingLyu: A sequential-GET fallback is supported in commit `
c64ee45`

shirly121 (approved):

PR #902, opened 2026-08-20
fix: validate zvec ANTLR patch state

shirly121:
## Summary

- refactor NeuG's third-party patch handling into a reusable state-aware helper
- verify zvec's bundled ANTLR4 patch against the actual tracked files before configuring zvec
- recreate zvec's compatibility marker only after the patch is confirmed present
- fail with actionable diagnostics when a patch is neither applicable nor already applied

## Problem introduced by #851

PR #851 added the HNSW vector search extension and its zvec submodule dependency. zvec patches its bundled ANTLR4 runtime during CMake configure, but its apply_patch_once() implementation trusts an untracked .antlr4_fix_patched marker without validating the patched files.

On reused self-hosted runners, actions/checkout can reset the nested ANTLR4 submodule's tracked files while leaving that untracked marker behind. Subsequent builds then skip the patch even though it is no longer present. The original ANTLR4 CMake logic runs find_package(PkgConfig REQUIRED) on Linux, and the current neug-dev image does not provide pkg-config, so the extension build fails persistently until the runner workspace is cleaned.

This change makes NeuG inspect the real patch state on every configure. If checkout reset ANTLR4 but left the marker behind, NeuG reapplies the patch before zvec sees the marker.

## Validation

- configured vector_search from clean zvec and ANTLR4 sources; both patches were applied
- reproduced the polluted-runner state by retaining .antlr4_fix_patched while reverting the ANTLR4 tracked changes; configure detected and reapplied the patch
- configured again with the patch present; configure recognized it as already applied without applying it twice
- ran cmake -P cmake/BuildZVecAsThirdParty.cmake
- ran git diff --check

### DIFF
--- a/.github/workflows/neug-extension-test.yml
+++ b/.github/workflows/neug-extension-test.yml
@@ -52,6 +52,11 @@ jobs:
         git config --global --add safe.directory "$GITHUB_WORKSPACE/third_party/parallel-hashmap"
         git config --global --add safe.directory "$GITHUB_WORKSPACE/third_party/pybind11"
 
+    - name: Clean ZVec state from self-hosted runner
+      run: |
+        git submodule deinit --force third_party/zvec
+        git submodule update --init --force --recursive third_party/zvec
+
     - name: Clean the tool directory for self-hosted runner
       run: |
         rm -rf /__w/_tool/Python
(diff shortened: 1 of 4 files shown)


Spockkk0225 (dismissed):

longbinlai (approved):
LGTM

PR #924, opened 2026-08-25
fix: "invalid oe offset" when creating a relationship after edge deletion (#922)

### LINKED ISSUE #922: [BUG] Creating a relationship fails with "invalid oe offset" after its source node had an outgoing edge deleted
## Summary

Creating a relationship fails with `ERR_INVALID_ARGUMENT: invalid oe offset` whenever the **source node previously had an outgoing edge deleted**. The failure is deterministic and reproduces with a minimal schema (no indexes, no vector columns), so it appears to be a storage-layer issue in edge deletion bookkeeping.

## Error

```
ERR_INVALID_ARGUMENT: invalid oe offset
at src/storages/graph/edge_table.cc:633 func: UpdateEdgeProperty, Error Code: 1009
```

## Minimal reproduction (Python binding only)

Environment: macOS (arm64), Python 3.12, NeuG local build (`build-py312`), accessed via the `neug` Python binding's `conn.execute(query, mode, params)` + `get_bolt_response()`.

```python
import json, shutil, tempfile
import neug

def exec_q(conn, query, params=None, mode=''):
    result = conn.execute(query, mode, params) if params else conn.execute(query, mode)
    resp = json.loads(result.get_bolt_response())
    return [
        dict(zip(rec['keys'], rec['_fields'], strict=True))
        for rec in resp.get('raw', {}).get('records', [])
    ]

def save_edge(conn, uuid, src, dst):
    # MERGE the edge, then SET its properties (Graphiti-style edge save)
    exec_q(
        conn,
        """
        MATCH (source:N {uuid: $source_uuid})
        MATCH (target:N {uuid: $target_uuid})
        MERGE (source)-[e:R {uuid: $uuid}]->(target)
        SET e.uuid = $uuid, e.name = $name
        RETURN e.uuid AS uuid
        """,
        {'uuid': uuid, 'name': f'edge-{uuid}', 'source_uuid': src, 'target_uuid': dst},
    )

tmpdir = tempfile.mkdtemp(prefix='neug_bug_oe_offset_')
db = neug.Database(tmpdir, mode='w')
conn = db.connect()

exec_q(conn, 'CREATE NODE TABLE N (uuid STRING, PRIMARY KEY (uuid))', mode='schema')
exec_q(conn, 'CREATE REL TABLE R (FROM N TO N, uuid STRING, name STRING)', mode='schema')
exec_q(conn, 'MERGE (a:N {uuid: "a"})')
exec_q(conn, 'MERGE (b:N {uuid: "b"})')

save_edge(conn, 'e1', 'a', 'b')                                            # step 1: OK
exec_q(conn, 'MATCH (a:N {uuid: "a"})-[e:R]->(b:N {uuid: "b"}) DELETE e')  # step 2: OK
save_edge(conn, 'e2', 'a', 'b')                                            # step 3: FAILS
```

- Step 1 (create `e1` from `a`): succeeds.
- Step 2 (delete `e1`): succeeds.
- Step 3 (create `e2` from the **same source** `a`): fails with `invalid oe offset`. The target does not matter — creating to a different target node fails the same way.

## Expected behavior

Step 3 should succeed: creating a new outgoing edge from a node whose previous outgoing edge was deleted is a normal workload pattern (e.g. Graphiti's edge invalidation/re-creation cycles).

## Observations from bisection

- Only nodes that previously acted as the **SOURCE** of a deleted edge are affected. Nodes that only ever had edges deleted on their **TARGET** side work fine.
- Brand-new nodes work fine.
- The failure happens during the property update of the newly created edge (`UpdateEdgeProperty` in `edge_table.cc`), suggesting stale/corrupted outgoing-edge (OE) table offsets after deletion.
- Reproduces with a minimal schema: one node table, one rel table, no indexes, no HNSW/FTS, no vector columns — so it is unrelated to extensions or index maintenance.
- Workaround currently used: avoid creating edges from any source node that has previously had an outgoing edge deleted (e.g. order operations so all creations precede deletions).

zhanglei1949:
## What do these changes do?

Fixes the `ERR_INVALID_ARGUMENT: invalid oe offset` failure when creating a relationship from a node whose previous outgoing edge had been deleted (issue #922).

### Root cause

`EdgeTable::UpdateEdgeProperty` positioned the CSR iterator with `iter += offset`, but the two sides of that expression use different offset semantics:

- **Producers** of the offsets — `MutableCsr::put_edge`'s return value, `record_to_csr_offset_pair`, `search_other_offset_with_cur_offset`, `delete_edge`, and WAL records — all use **raw adjacency-list slot indices**.
- **`NbrIterator::operator+=`** advances over **visible edges only** — `operator++` skips tombstones of edges deleted before the view timestamp.

With no deletions the two notions coincide. Once a vertex has a deleted edge, its adjacency list keeps the tombstone slot (degree is not shrunk), so a newly appended edge sits at raw slot *N* while its visible ordinal is *N-1*. `iter += offset` then lands past the end of the list and the `end()` check throws `invalid oe offset`.

### Fix

Position the iterator by raw slot (`iter.cur = start_ptr + offset * stride`, the same approach documented in `csr_view_utils.cc`), for both the outgoing and incoming sides. Offsets that fall outside the adjacency list **or onto a tombstone** are rejected, so a stale offset can never silently overwrite another edge's data.

### Test plan

- New C++ regression tests: `EdgeTableTest.TestUpdateEdgePropertyAfterDeletion{,Bundled}` (unbundled + bundled, the latter exercises both oe and ie offset paths) — both fail on the unfixed code and pass with the fix.
- New Python regression tests in `tools/python_bind/tests/test_dml.py`: exact issue repro (MERGE + SET after DELETE), 5x delete/re-create cycles, SET on surviving edges after a mid-list deletion, and bundled-edge re-create + SET — all fail on the unfixed code and pass with the fix.
- Existing suites: `transaction_test` 124 passed (incl. `test_update_transaction`, `test_acid`, WAL replay), `csr_test` 370 passed, `edge_table_test` EdgeTableTest 18 passed, Python `test_dml.py` 20 passed, full Python suite shows no new failures vs. baseline.

## Related issue number

Fixes #922

### DIFF
--- a/src/storages/graph/edge_table.cc
+++ b/src/storages/graph/edge_table.cc
@@ -634,9 +634,6 @@ namespace {
 // tombstone so a stale offset can never overwrite another edge's data.
 bool position_iter_at_raw_slot(NbrIterator& iter, const NbrList& edges,
                                int32_t offset) {
-  if (offset < 0 || edges.start_ptr == nullptr || edges.cfg.stride <= 0) {
-    return false;
-  }
   const auto* start = static_cast<const char*>(edges.start_ptr);
   const auto* end = static_cast<const char*>(edges.end_ptr);
   const auto slot_num =
@@ -645,8 +642,8 @@ bool position_iter_at_raw_slot(NbrIterator& iter, const NbrList& edges,
     return false;
   }
   iter.cur = start + static_cast<size_t>(offset) * edges.cfg.stride;
-  // A tombstone slot holds no visible edge; writing through it would corrupt
-  // the storage of the deleted edge.
+  // A tombstone slot holds no visible edge; set_data would rewrite its
+  // timestamp and silently resurrect the deleted edge.
   return iter.cfg.ts_offset == 0 || iter.get_timestamp() <= iter.timestamp;
 }
 
--- a/src/storages/graph/edge_table.cc
+++ b/src/storages/graph/edge_table.cc
@@ -621,6 +621,37 @@ void EdgeTable::DeleteVertex(bool is_src, vid_t vid, timestamp_t ts) {
   }
 }
 
+namespace {
+
+// Offsets consumed by EdgeTable::UpdateEdgeProperty (produced by
+// CsrBase::put_edge, record_to_csr_offset_pair, or WAL records) are raw
+// adjacency-list slot indices, not visible-edge ordinals.
+// NbrIterator::operator+= advances over visible edges only — it skips
+// tombstones of edges deleted before the view timestamp — so once a vertex
+// has a deleted edge the two notions diverge and `iter += offset` lands past
+// the end of the list (issue #922). Position the iterator by raw slot
+// instead, and reject offsets that fall outside the adjacency list or onto a
+// tombstone so a stale offset can never overwrite another edge's data.
+bool position_iter_at_raw_slot(NbrIterator& iter, const NbrList& edges,
+                               int32_t offset) {
+  if (offset < 0 || edges.start_ptr == nullptr || edges.cfg.stride <= 0) {
+    return false;
+  }
+  const auto* start = static_cast<const char*>(edges.start_ptr);
+  const auto* end = static_cast<const char*>(edges.end_ptr);
+  const auto slot_num =
+      static_cast<size_t>(end - start) / static_cast<size_t>(edges.cfg.stride);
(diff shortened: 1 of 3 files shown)


BingqingLyu (approved):
LGTM

PR #934, opened 2026-08-25
fix(nodejs): align mimalloc with core build

### LINKED ISSUE #932: [BUG] HNSW distance ordering over a relationship traversal segfaults the native runtime
**Describe the bug**

A vector-distance Top-K query segfaults the native runtime when the indexed vector node is reached through a relationship pattern in the same query part. The process exits with status 139 immediately after entering the query.

The same schema, data, HNSW index, and query vector work when candidate generation starts from the vector node and the relationship pattern is matched in a second query part after `WITH ... ORDER BY ... LIMIT`.

This report only claims a reproducible native crash. I have not isolated whether the failing component is the query planner/optimizer, the HNSW extension, or their operator combination.

**To Reproduce**

Build the Node.js binding and `vector_search` extension from the referenced commit. Save the following file as `tools/nodejs_bind/repro-hnsw-traversal.mjs` and run it from the repository root.

```js
import { mkdtemp } from 'node:fs/promises'
import { tmpdir } from 'node:os'
import { join } from 'node:path'

const mode = process.argv[2]
if (mode !== 'combined' && mode !== 'hnsw-first') {
  throw new Error('usage: node repro-hnsw-traversal.mjs <combined|hnsw-first>')
}

process.env.NEUG_EXTENSION_HOME_PYENV = new URL('../../build', import.meta.url).pathname
const { Database } = await import('./lib/index.js')
const databasePath = await mkdtemp(join(tmpdir(), `neug-hnsw-traversal-${mode}-`))
const database = new Database({ databasePath, mode: 'w', maxThreadNum: 0, checkpointOnClose: false })
const connection = database.connect()

function execute(query, accessMode = undefined, parameters = null) {
  const result = connection.execute(query, accessMode, parameters)
  try {
    return [...result]
  } finally {
    result.close()
  }
}

try {
  execute('LOAD vector_search;')
  execute('CREATE NODE TABLE Source(id STRING PRIMARY KEY);')
  execute('CREATE NODE TABLE Chunk(id STRING PRIMARY KEY, embedding FLOAT[4]);')
  execute('CREATE REL TABLE HAS_CHUNK(FROM Source TO Chunk);')
  execute("CREATE (:Source {id: 'source'});", 'insert')
  execute("CREATE (:Chunk {id: 'chunk-a', embedding: [1.0, 0.0, 0.0, 0.0]});", 'insert')
  execute("CREATE (:Chunk {id: 'chunk-b', embedding: [0.0, 1.0, 0.0, 0.0]});", 'insert')
  execute("MATCH (s:Source), (c:Chunk) WHERE s.id = 'source' CREATE (s)-[:HAS_CHUNK]->(c);", 'insert')
  execute("CREATE INDEX chunk_hnsw ON Chunk USING HNSW (embedding) WITH (metric = 'cosine', m = 2, ef_construction = 10);")

  const query = mode === 'combined'
    ? `MATCH (source:Source)-[:HAS_CHUNK]->(chunk:Chunk)
       RETURN source.id, chunk.id,
              vector_distance_cosine(chunk.embedding, $embedding) AS distance
       ORDER BY distance ASC
       LIMIT 1;`
    : `MATCH (chunk:Chunk)
       WITH chunk, vector_distance_cosine(chunk.embedding, $embedding) AS distance
       ORDER BY distance ASC
       LIMIT 1
       MATCH (source:Source)-[:HAS_CHUNK]->(chunk)
       RETURN source.id, chunk.id, distance
       ORDER BY distance ASC;`

  console.log(JSON.stringify({ mode, stage: 'before-query' }))
  console.log(execute(query, 'read', { embedding: [1.0, 0.0, 0.0, 0.0] }))
} finally {
  connection.close()
  database.close()
}
```

Run the relationship-first form:

```console
$ node tools/nodejs_bind/repro-hnsw-traversal.mjs combined
{"mode":"combined","stage":"before-query"}
$ echo $?
139
```

Run the HNSW-first form:

```console
$ node tools/nodejs_bind/repro-hnsw-traversal.mjs hnsw-first
{"mode":"hnsw-first","stage":"before-query"}
[ [ 'source', 'chunk-a', 0 ] ]
$ echo $?
0
```

I reproduced the exit-139 result with only two node tables, one relationship table, two four-dimensional vectors, and no FTS, Namespace, checkpoint, or application framework involved.

**Expected behavior**

The `combined` query should return the nearest related `Chunk`, or return a normal query/planner error if this query form is unsupported. It should not terminate the host process.

**Environment**

- NeuG commit: `9a49a68350e721958c997edc1d5b80063ca2875b`
- Build: local source build of the Node.js binding and `vector_search` extension
- OS: Darwin 24.6.0, arm64
- Node.js: v26.3.1

**Additional context**

Splitting the operation into an HNSW-only first query part and a relationship match after `WITH ... LIMIT` is a working workaround, but it changes how callers must formulate otherwise equivalent retrieval queries.

Spockkk0225:
## Summary
- enable mimalloc by default for Node.js builds, matching the NeuG root CMake default
- stage the platform-specific mimalloc shared library alongside the Node addon and libneug
- add a sibling-library RPATH to libneug so packaged Node binaries can resolve mimalloc after installation

## Validation
- built neug_node_bind and vector_search from the same build tree with WITH_MIMALLOC=ON
- reproduced both combined and hnsw-first HNSW traversal queries; both return successfully with no segfault
- built an npm tarball and verified it contains neug_node_bind.node, libneug.dylib, and libmimalloc.2.dylib
- extracted the tarball into an isolated directory and successfully opened/closed a database
- npm test: 76 passed, 11 skipped; 2 unrelated failures require the missing /tmp/modern_graph fixture

fix #932


zhanglei1949 (commented):
## Review Summary

**LGTM** — the change is correct, well-scoped, and the packaging approach is sound. Verified against the root build config and a local build tree:

- Root CMake defaults `WITH_MIMALLOC=ON` (non-Windows builds mimalloc as SHARED and links it `PUBLIC` into libneug), so the old forced `-DWITH_MIMALLOC=OFF` here caused allocator drift across the shared root build tree. Aligning the default is the right fix.
- `mimalloc` is the **only** non-system dynamic dependency of `libneug.dylib` (`otool -L`), so staging `neug_node_bind.node + libneug + libmimalloc` side-by-side with `@loader_path`/`$ORIGIN` RPATH is a complete solution (npm has no delocate/auditwheel equivalent; the Python wheel handles this via `delocate-wheel` in `pyproject.toml`, so the manual approach here is necessary).
- Library names/paths check out: `libmimalloc.2.dylib` / `libmimalloc.so.2` (SOVERSION 2) under `build/third_party/mimalloc/`, and `cp` dereferences the version symlinks.
- `set_property(TARGET neug APPEND BUILD_RPATH ...)` only touches the build-tree RPATH; `INSTALL_RPATH` is unchanged, so `cmake --install` behavior is unaffected.
- `WITH_MIMALLOC` propagates correctly through the recursive make invocations (`dev→stage`, `pack→dev`), and the explicit `-D` on every configure refreshes the CMake cache.
- CI (`npm-build-test.yml`) invokes `make build/pack` without overriding the variable, so it will exercise the new default; `npm-package-test.yml` end-to-end-validates the tarball on a clean runner.
- Nice side effect: the implementation now matches what the README already claimed on main (it listed `libmimalloc.so.2` in the package contents even though staging never copied it).

### Non-blocking suggestions

1. **CMakeLists.txt**: the two `if (WITH_MIMALLOC)` blocks are duplicates except for `@loader_path` vs `$ORIGIN`; consider extracting a platform variable and keeping a single block:
   ```cmake
   if (APPLE)
     set(_neug_rpath "@loader_path")
   else()
     set(_neug_rpath "$ORIGIN")
   endif()
   if (WITH_MIMALLOC)
     set_property(TARGET neug APPEND PROPERTY BUILD_RPATH "${_neug_rpath}")
   endif()
   ```
2. **Packaging hygiene (pre-existing)**: the staged `libneug` still carries the build machine's absolute RPATH (e.g. `/home/runner/.../build/third_party/mimalloc`) into the tarball. Functionally harmless since `@loader_path` resolves first, but it leaks builder paths. Could be a follow-up (strip via `install_name_tool` during `pack`, or enable `CMAKE_BUILD_RPATH_USE_ORIGIN`).
3. **README (pre-existing nit)**: "Copy prebuilt binaries into `prebuilds/<platform>/`" doesn't match the actual staging dir `build/Release/<platform>/` (which is what `lib/binding.js` loads from).
4. **Makefile edge case**: running `make stage` standalone after `WITH_MIMALLOC=OFF make dev` fails at the `cp` (the Make variable defaults back to `ON` while the build tree has no mimalloc). Low priority — the error is understandable.

None of these block merging; thanks for the thorough validation (isolated-dir install test + HNSW repro)!

Spockkk0225:
The first and third issues have been addressed by refactoring CMakeLists.txt to use a shared _neug_node_rpath platform variable and correcting the staging path documented in README.md. https://github.com/alibaba/neug/pull/934/commits/f85041b92f9c1a3a957fe281526249aa74dbb596 https://github.com/alibaba/neug/pull/934/commits/0e436f769cd8299130169aaae6058bc2b87ea661

The second issue—the absolute build RPATH retained in packaged libneug—is pre-existing behavior inherited from the root NeuG build and is out of scope for this fix. For the fourth issue, users are currently expected to use consistent WITH_MIMALLOC options across configuration, build, and staging commands; if those options differ, failure during an incremental build or standalone staging step is considered reasonable and understandable.

zhanglei1949 (dismissed):
LGTM

zhanglei1949 (approved):
LGTM

PR #946, opened 2026-08-26
fix: reject null property updates

### LINKED ISSUE #945: Rollback unsupported NULL handling for persistent properties and indexes
## Problem

Persistent property columns do not support NULL. A NULL property is currently stored as a type default such as `0`, `0.0`, `false`, or an empty string.

However, some index code already treats NULL as a valid state. For example, `StorageIndex::Upsert()` deletes the index entry when the new value is NULL. This can make storage and indexes inconsistent.

## Changes needed

- Reject NULL property writes with a clear error until storage supports NULL.
- Remove NULL-specific behavior from indexes and related tests that imply persistent NULL is supported.
- Document that persistent properties currently do not support NULL.

Execution-layer NULL handling, such as expressions, GROUP BY, and aggregation, is separate and should remain supported.

Related: #885, #893.

Spockkk0225:
## Summary
- restore strict type checks for property updates
- assert storage index upserts receive non-NULL values
- update index tests and documentation

This issue was introduced in https://github.com/alibaba/neug/pull/870/commits/11cac9ce9227d2d861d064812009a38344f96281 and we should rollback it.

Closes #945

### DIFF
--- a/src/execution/execute/ops/batch/batch_update_vertex.cc
+++ b/src/execution/execute/ops/batch/batch_update_vertex.cc
@@ -101,6 +101,10 @@ neug::result<ContextChunk> UpdateVertexOpr::eval_impl(
       int32_t col_id = std::distance(property_names.begin(), pos);
       assert(col_id >= 0 &&
              col_id < static_cast<int32_t>(property_names.size()));
+      if (value.IsNull()) {
+        THROW_NOT_SUPPORTED_EXCEPTION("Setting NULL for property " +
+                                      prop_name);
+      }
       if (property_types[col_id] != value.type()) {
         THROW_RUNTIME_ERROR("Property type mismatch for property " + prop_name);
       }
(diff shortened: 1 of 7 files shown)


src/execution/execute/ops/batch/batch_update_vertex.cc
```diff
@@ -101,9 +101,7 @@ neug::result<ContextChunk> UpdateVertexOpr::eval_impl(
       int32_t col_id = std::distance(property_names.begin(), pos);
       assert(col_id >= 0 &&
              col_id < static_cast<int32_t>(property_names.size()));
-      // SET may assign NULL without carrying the property's concrete type, so
-      // skip the type comparison for NULL values.
-      if (!value.IsNull() && property_types[col_id] != value.type()) {
+      if (property_types[col_id] != value.type()) {
```
zhanglei1949, line 107: Maybe we should raise unsupported error when value.IsNull()

zhanglei1949 (approved):
LGTM

PR #951, opened 2026-08-27
feat: add Node.js bindings for AP explicit transactions

zhanglei1949:
## Summary
- expose the embedded AP Connection transaction lifecycle through beginTransaction, commit, rollback, and hasActiveTransaction
- keep transaction state and COW ownership in the existing C++ Connection, with JavaScript only validating options and mapping Status errors
- document the Node.js API and cover lifecycle, rollback-only, read-only, close, and WAL recovery behavior

## Verification
- cmake --build build --target neug_node_bind -j4
- explicit transaction Node.js smoke test: passed
- node --test tools/nodejs_bind/tests/test_db_transaction.js: 14 passed, 1 skipped; 2 unrelated fixture failures because /tmp/modern_graph is empty, with the second caused by the first retaining its read lock
- git diff --check

Refs #779

### DIFF
--- a/tools/nodejs_bind/src/node_connection.cc
+++ b/tools/nodejs_bind/src/node_connection.cc
@@ -152,7 +152,15 @@ Napi::Value NodeConnection::HasActiveTransaction(
 }
 
 Napi::Value NodeConnection::GetSchema(const Napi::CallbackInfo& info) {
-  return Napi::String::New(info.Env(), conn_->GetSchema());
+  Napi::Env env = info.Env();
+  try {
+    return Napi::String::New(env, conn_->GetSchema());
+  } catch (const neug::exception::Exception& e) {
+    Napi::Error::New(env, e.what()).ThrowAsJavaScriptException();
+  } catch (const std::exception& e) {
+    Napi::Error::New(env, e.what()).ThrowAsJavaScriptException();
+  }
+  return env.Null();
 }
 
 Napi::Value NodeConnection::Close(const Napi::CallbackInfo& info) {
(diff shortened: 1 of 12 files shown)


tools/nodejs_bind/src/node_connection.cc
```diff
@@ -99,6 +119,38 @@ Napi::Value NodeConnection::Execute(const Napi::CallbackInfo& info) {
...
+  return StatusToJs(info.Env(), status);
+}
+
+Napi::Value NodeConnection::HasActiveTransaction(
+    const Napi::CallbackInfo& info) {
+  return Napi::Boolean::New(info.Env(), conn_ && conn_->HasActiveTransaction());
+}
+
 Napi::Value NodeConnection::GetSchema(const Napi::CallbackInfo& info) {
   return Napi::String::New(info.Env(), conn_->GetSchema());
```
Spockkk0225: When the transaction is rollback-only, `conn_->GetSchema()` throws a `TxStateConflictException`. This binding callback does not catch that C++ exception or convert it to a `Napi::Error`, so it may escape the N-API callback boundary and terminate the Node.js process instead of producing a catchable JavaScript error. Please catch and translate the exception here, and add a regression test that calls `getSchema()` after a statement has put an explicit transaction into rollback-only state.

zhanglei1949:
TODO: add tab to transaction doc

tools/nodejs_bind/src/node_connection.cc
zhanglei1949: resolved.

Spockkk0225 (approved):

Issue #122, opened 2026-03-25
Issues from graph testing benchmark


longbinlai:
As titled. This is for Tinghan to collect all issues exposed from the testing

PR #954, opened 2026-08-27
fix: propagate the PARALLEL option to the native CSV reader

johnmzz:
## Summary

Pass the CSV `PARALLEL` option throughout the native CSV loading path.

Previously, `parallel=false` was parsed but not propagated to the underlying `csv-parser`. As a result, the third-party reader continued using its default threaded mode, and NeuG's CSV row-counting pass could also create worker threads.

## Changes

- Add `use_threads` to `CsvReadConfig`.
- Propagate the CSV `PARALLEL` option through both CSV configuration paths.
- Pass `use_threads` to `csv::CSVFormat::threading()`.
- Make the CSV row-counting pass respect the same option.
- Keep CSV parallelism enabled by default.
- Add regression tests for the default and explicit `parallel=false` behavior.

## Behavior

- `parallel=true`:
  - enables adaptive parallel row counting in NeuG;
  - enables threading in the third-party CSV reader;
  - allows the third-party reader to choose speculative parallel parsing for sufficiently large local files.

- `parallel=false`:
  - uses single-threaded row counting;
  - makes the third-party CSV reader parse synchronously on the caller thread.

Related to #122.


src/storages/loader/loader_utils.cc
```diff
@@ -1271,6 +1276,19 @@ CsvReadConfig build_csv_read_config(
     config.double_quote = (value == "true" || value == "1" || value == "TRUE");
   }
 
+  if (csv_options.count("PARALLEL")) {
```
zhanglei1949: 目前这里写死的字符串，似乎与 loading_config.h里面定义的字面常量重复，考虑统一都定义在loading_config.h里面，避免在.cc里面使用字符串字面量

tests/utils/test_reader.cc
```diff
@@ -15,9 +15,43 @@
...
+  reader::CsvOptionsBuilder singleThreadedBuilder(singleThreadedState);
+  EXPECT_FALSE(singleThreadedBuilder.build().use_threads);
+
+  const auto filePath =
+      std::string(ARROW_READER_TEST_DIR) + "/parallel_options.csv";
+  const std::vector<DataType> nativeColumnTypes = {
+      DataType(DataTypeId::kInt64)};
+  EXPECT_TRUE(
+      build_csv_read_config(filePath, {}, nativeColumnTypes).use_threads);
+  EXPECT_FALSE(build_csv_read_config(filePath, {{"PARALLEL", "false"}},
```
zhanglei1949: 同上。避免"PARALLEL"

zhanglei1949:
We should move/remove the const char definition like: https://github.com/alibaba/neug/blob/54b3d507eeeaebea1b8bb27adafb85f16bbb82b4/include/neug/execution/execute/ops/batch/batch_update_utils.h#L41. 
Recommand moving to loading_config.h

zhanglei1949 (approved):
LGTM

PR #961, opened 2026-08-27
fix: restore Windows CI with nothrow GraphView swap

longbinlai:
## Summary

- **Restore the Windows build on main.** Every `NeuG Test (Windows)` run on main has failed since #887 (7286fdde) with `graph_snapshot_store.cc(347): error C2607: static assertion failed` on `std::is_nothrow_swappable_v<GraphView>`. Root cause: MSVC's `std::unordered_map` move constructor/assignment are **not** `noexcept` ([microsoft/STL#165](https://github.com/microsoft/STL/issues/165)), so `GraphView`'s defaulted move operations — and therefore `std::swap` on it — are not nothrow on MSVC, while libc++/libstdc++ mark them noexcept and pass.
- **Fix:** give `GraphView` a member-wise `swap` (3 pointers + `vector::swap` / `unordered_map::swap`, all `noexcept` per the standard) exposed via ADL, so the existing `static_assert` in `replaceCurrentSnapshotInPlace` holds on every implementation. No change to `graph_snapshot_store.cc`.
- **Opt-in Windows CI for PRs:** `Windows.yml` now also fires on `pull_request` (`labeled`/`synchronize`/`reopened`), but both jobs are gated on the new **`ci:windows`** label. Unlabeled PRs cost nothing (jobs skip); adding the label runs the full Windows pipeline for that PR, and new pushes to a labeled PR re-run it. This PR is labeled `ci:windows` to verify the fix on MSVC before merge. (`workflow_dispatch` on a branch remains available for one-off runs.)

## Verification

- Local clang++ `-std=c++20` syntax check of `graph_view.h` with three new asserts: ADL `swap(GraphView&, GraphView&)` is noexcept, `std::is_nothrow_swappable_v<GraphView>` is true, and the member `swap` is noexcept.
- Member-wise swap is noexcept on MSVC by the standard: `unordered_map::swap` is unconditionally `noexcept` (C++17), `vector::swap` is `noexcept` for `std::allocator`, pointer swaps are trivial.
- The real proof: the `ci:windows`-labeled run on this PR.

## Notes

- Windows CI has been red on main since 7286fdde (Aug 26); #914 (add653a4) was the last green run.


.github/workflows/Windows.yml
```diff
@@ -5,6 +5,14 @@ on:
   release:
     types:
       - created
+  # Opt-in only: Windows CI is slow, so PRs run it solely when a maintainer
+  # adds the `ci:windows` label. `synchronize` keeps it re-running on new
+  # pushes of labeled PRs; unlabeled PRs skip every job.
+  pull_request:
```
zhanglei1949, line 11: [P2] Ensure the opt-in Windows check can finish. The only completed ci:windows run for the previous head reached the existing 120-minute job timeout and was cancelled during the final Python test: graph_snapshot_store.cc had compiled successfully and all C++ tests passed, but the overall check never produced a success result. The new trigger therefore currently gives labeled PRs a cancelled check instead of usable Windows validation. Please increase timeout-minutes, or split or trim the job, so the labeled PR workflow can complete reliably.

zhanglei1949: https://github.com/alibaba/neug/actions/runs/33069180023/job/98506910761

longbinlai:
This will be closely investigated @lnfjpt

PR #970, opened 2026-08-28
fix: support TP transaction lookup on older libstdc++

### LINKED ISSUE #969: fix: support TP transaction lookup on older libstdc++
## Problem

`ServiceTransactionManager` stores transaction sessions in `std::unordered_map<std::string, ...>` and looks them up with `std::string_view`. The current transparent hash/equality configuration requires the C++20 heterogeneous `unordered_map::find` overload. Some supported libstdc++ versions do not provide that overload, causing the TP service build to fail.

## Expected behavior

The TP transaction manager should compile with supported standard library implementations while preserving transaction lookup and removal behavior.

## Proposed scope

Use the standard `std::string` map lookup overload in the two internal lookup sites and remove the now-redundant transparent hash helper.

zhanglei1949:
Closes #969

## Summary

- use the portable `std::string` lookup overload for TP transaction-session lookup and removal
- remove the redundant transparent hash helper

## Validation

- `cmake --build build --target test_db_svc`
- `MODERN_GRAPH_DATA_DIR="$PWD/example_dataset/modern_graph" build/tests/unittest/test_db_svc --gtest_filter=NeugDBServiceTest.ExplicitTransaction*`

### DIFF
--- a/src/server/service_transaction_manager.cc
+++ b/src/server/service_transaction_manager.cc
@@ -315,7 +315,7 @@ ServiceTransactionManager::LockEntry(std::string_view transaction_id) {
 ServiceTransactionManager::EntryPtr ServiceTransactionManager::Find(
     std::string_view transaction_id) const {
   std::lock_guard lock(mutex_);
-  const auto found = entries_.find(transaction_id);
+  const auto found = entries_.find(std::string(transaction_id));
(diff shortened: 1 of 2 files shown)


Spockkk0225 (approved):
LGTM

PR #972, opened 2026-08-28
fix: report actionable Python Session timeout errors

### LINKED ISSUE #971: Improve Python Session timeout error message
## Problem

When a remote query exceeds the Python `Session` timeout, `Session.execute()` catches `requests.Timeout` as a generic `RequestException` and raises:

```text
Could not execute query: MATCH (n) RETURN n, Error code: 2001
```

The message classifies the timeout as a generic network error (`ERR_NETWORK`) and omits the configured timeout and remediation guidance. This makes it difficult to distinguish a slow query from other connection failures.

The millisecond timeout form also cannot currently be parsed because values such as `250ms` match the `s` suffix before the `ms` suffix.

## Expected behavior

- Report query timeouts as `TimeoutError` with `ERR_QUERY_TIMEOUT` (`3003`).
- Include the effective and configured Session timeout in the message.
- Suggest increasing the Session timeout or optimizing the query, including a configuration example.
- Continue reporting non-timeout network failures as `ERR_NETWORK` (`2001`).
- Correctly support both second and millisecond timeout strings.

## Reproduction

Configure a short Session timeout and execute a query whose HTTP response takes longer than the configured limit:

```python
session = Session(endpoint, timeout="2s")
session.execute("MATCH (n) RETURN n")
```

A deterministic regression test can reproduce the behavior by injecting `requests.exceptions.ReadTimeout` into the Session HTTP POST.

shirly121:
## Summary

- distinguish HTTP query timeouts from generic network failures in the Python `Session` API
- raise `TimeoutError` with `ERR_QUERY_TIMEOUT` (3003), the configured timeout, endpoint, and actionable guidance
- fix parsing of millisecond timeout values such as `250ms`
- add deterministic regression coverage for second and millisecond timeout configurations

## Before

```text
Could not execute query: MATCH (n) RETURN n, Error code: 2001
```

## After

```text
Query timed out after 2 seconds (session timeout: 2s) while requesting http://localhost:10000/cypher. Consider increasing the Session timeout or optimizing the query. For example: Session(endpoint, timeout="30s"). Error code: 3003
```

## Testing

```text
python3.9 -m pytest -q tests/test_session_timeout.py
2 passed
```

Closes #971

### DIFF
--- a/tools/python_bind/neug/session.py
+++ b/tools/python_bind/neug/session.py
@@ -33,6 +33,7 @@ except ImportError as e:
         raise e
 
 from neug.proto.error_pb2 import ERR_NETWORK
+from neug.proto.error_pb2 import ERR_QUERY_TIMEOUT
 from neug.proto.error_pb2 import ERR_SESSION_CLOSED
 from neug.query_result import QueryResult
 from neug.utils import is_access_mode_valid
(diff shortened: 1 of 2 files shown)


longbinlai (approved):
LGTM

zhanglei1949 (approved):
LGTM

PR #974, opened 2026-08-28
fix: avoid materializing unused variable-length paths

### LINKED ISSUE #228: [BUG] Concurrent *2..2 path queries retain large thread-local scratch memory and drive RSS into the GB range
**Describe the bug**
We noticed a memory-retention issue around a specific concurrent `*2..2` path-query shape, and wanted to check whether you would consider this expected under the current implementation or a bug worth fixing.

When the query includes the special `*2..2` path shape with an inline predicate and is executed concurrently, RSS and private anonymous memory rise sharply and do not fall back after the workload stops. The effect becomes operationally significant under higher worker counts because the retained memory is mostly private anonymous memory.

Current source inspection suggests the special path uses `static thread_local` scratch containers that are cleared between iterations but do not release their enlarged capacity.

**Reproduction**
Minimal reproducer shape:

```bash
python reproduce.py --mode special --loops 3 --sample-every 1 --hot-neighbor-count 100000 --parallel-clients 64 --max-thread-num 64 --rss-stop-mb 1700 --output-dir artifacts/special
```

Control:

```bash
python reproduce.py --mode control --loops 3 --sample-every 1 --hot-neighbor-count 100000 --parallel-clients 64 --max-thread-num 64 --rss-stop-mb 1700 --output-dir artifacts/control
```

The reproducer bulk-loads a synthetic graph with one hot source vertex, many `Alice` intermediates, and one sink vertex, then compares:

- `special`: a `MATCH p = (a:person)-[e:knows*2..2 (r, n | WHERE n.fName = 'Alice')]->(b:person) ...` workload
- `control`: a structurally simpler query that does not trigger the same path specialization

I also uploaded a gist with the standalone reproducer and root-cause notes here:

https://gist.github.com/d1zzy126/efabe978ac518c180bfc0cc377aa9fcb

**Expected behavior**
Temporary scratch memory used for a path fast path should not remain inflated long after the query burst has ended. Some allocator retention is understandable, but repeated bursts should converge toward a stable steady-state band.

**Actual behavior**
On one confirmation run with `64` worker threads, `64` concurrent clients, and `100000` hot neighbors:

```text
baseline RSS ~= 115.14 MB
post-workload RSS ~= 1729.28 MB
after-idle RSS ~= 1729.32 MB
after-idle Anonymous ~= 1711840 KB
```

The retained memory stayed high after idle instead of returning near baseline.

**Suspected root cause**
The current code appears to keep large scratch capacity in `static thread_local` containers in the `*2..2` fast path. The containers are content-reset (`clear()` / element reset) but not shrunk, so a single large transient workload can leave per-thread capacity pinned for later reuse.

I would appreciate your view on two questions:

1. Is this retention expected under the current worker-thread / fast-path design?
2. If not, would you consider it a bug or an area for memory-reclamation improvement?

johnmzz:
## What does this PR do?

This PR avoids unnecessary path materialization for predicate-free variable-length walk queries when the path and relationship variables are not used by downstream operators.

For example:

```cypher
MATCH p = (a:person)-[e:knows*2..2]->(b:person)
WHERE a.ID = 0
RETURN count(*);
```

Previously, the planner selected `ALL_V` and materialized complete Path objects even though `p` and `e` were unused. This caused significantly higher memory usage under concurrent execution.

## Changes

- Add `LogicalRecursiveExtend` handling to the projection push-down optimizer.
- Select `END_V` when:
  - the query uses a variable-length walk;
  - the path/relationship result and its properties are unused;
  - no recursive node or relationship predicate is present.
- Preserve `resultOpt` when copying `LogicalRecursiveExtend`.
- Add a regression test verifying that the physical plan uses `END_V`.

Queries that consume the path result or contain recursive predicates continue to use `ALL_V`.

## Scope

This PR does not optimize variable-length paths containing inline node or relationship predicates, because the endpoint-only execution path does not currently support those predicates.

Related to #228.


tests/compiler/path_test.cpp
```diff
@@ -18,6 +18,8 @@
 #include <memory>
```
shirly121, line 18: 需要增加端到端测试，最好可以打印出 path 值并且确认为空，如果引擎还是保存这些点边的话这个优化就没有意义了

shirly121: 另外增加测试验证 opt = END_V 的情况下, length(path) 能返回正确值

shirly121 (approved):

PR #977, opened 2026-08-31
fix: resize incident CSR after bulk vertex copy

### LINKED ISSUE #936: Bulk COPY vertex growth leaves incident edge CSR stale after edges exist
## Bug

After a relationship triplet has been populated, a later persistent `COPY` into either endpoint node table can grow `VertexTable` past its reserved headroom without growing the incident edge CSR directory. A subsequent ordinary `CREATE` edge involving a newly copied vertex fails with `Source vertex id out of range`.

This is distinct from closed #807: that fix makes `EdgeTable::BatchAddEdges` use the node indexer capacity, so a later single-node write within the existing headroom works. This issue occurs when a later **bulk node COPY itself grows capacity**.

## Cause

`BulkCowGraphStorage::BatchAddVerticesImpl()` detaches the vertex table and indexes, then calls `PropertyGraph::BatchAddVertices()`. `VertexTable::insert_vertices()` may call its local `EnsureCapacity()`, but the bulk path does not use `CowGraphStorage::detachForResize()` followed by `PropertyGraph::EnsureCapacity()`. Consequently, the incident `EdgeTable` CSR remains sized to the old node capacity.

An ordinary edge insert only grows edge storage when its property capacity is exhausted; it can reach `MutableCsr::put_edge()` first and reject the new endpoint LID as out of range.

## Reproduction

1. Create node table `N` and relationship table `R (FROM N TO N)`.
2. `COPY` enough `N` rows to establish a capacity, then `COPY` at least one `R` row.
3. `COPY` additional `N` rows so the node table crosses its current 25% headroom and grows capacity.
4. `CREATE` an `R` edge from or to one of the newly copied nodes.

Expected: the edge insert succeeds.

Actual: the mutable CSR reports `Source vertex id out of range: <new_lid> >= <old_capacity>` (or the corresponding inbound-side failure).

## Scope

`tools/python_bind/tests/test_loader.py` is not currently affected because it copies all node tables before every relationship table. Any incremental loader that copies nodes after an incident edge table already exists is affected.

## Proposed direction

Make bulk node capacity growth follow the same COW resize protocol as ordinary `AddVertex`: detach all incident edge tables before growth, then propagate the new capacity through `PropertyGraph::EnsureCapacity()`. The solution must also handle suppliers whose actual chunk count exceeds/does not provide `RowNum()`, rather than relying only on a preflight estimate.

## Acceptance test

Add a regression covering the four-step sequence above, crossing the capacity boundary, then checkpoint/reopen and verify the new edge is queryable.

Regression introduced/exposed by [PR #887](https://github.com/alibaba/neug/pull/887).

zhanglei1949:
## Summary

- Preserve the existing `BatchAddVertices` interface and reconcile incident CSR capacity after persistent COPY actually grows a vertex table.
- Detach, resize, and mark incident EdgeTables dirty only on an actual capacity change, using the vertex table post-COPY capacity rather than `RowNum()` preflight.
- Add a regression that checkpoints and reopens immediately after node growth, before creating an edge from the newly added vertex.

Fixes #936.

### DIFF
--- a/include/neug/storages/graph/property_graph.h
+++ b/include/neug/storages/graph/property_graph.h
@@ -356,6 +356,7 @@ class NEUG_API PropertyGraph {
                               const DeleteEdgePropertiesParam& config);
 
   Status EnsureCapacity(label_t v_label, size_t capacity);
+  Status SyncIncidentEdgeCapacity(label_t v_label);
 
   Status EnsureCapacity(label_t src_label, label_t dst_label,
                         label_t edge_label, size_t capacity);
--- a/include/neug/transaction/cow_graph_storage.h
+++ b/include/neug/transaction/cow_graph_storage.h
@@ -134,6 +134,7 @@ class CowGraphStorage : public StorageUpdateInterface,
   Status detachEdgeColumn(uint32_t edge_triplet_id, int32_t col_id);
   Status detachAdjlists(uint32_t edge_triplet_id, vid_t src_lid, vid_t dst_lid,
                         Allocator& alloc);
+  Status detachIncidentEdgeTablesForResize(label_t label);
   Status detachForResize(label_t label, size_t capacity);
   Status detachForResize(label_t src_label, label_t dst_label,
                          label_t edge_label, size_t capacity);
--- a/src/storages/graph/property_graph.cc
+++ b/src/storages/graph/property_graph.cc
@@ -134,7 +134,17 @@ Status PropertyGraph::EnsureCapacity(label_t v_label, size_t capacity) {
     if (capacity <= old_cap) {
       return neug::Status::OK();
     }
-    auto v_new_cap = vertex_tables_[v_label].EnsureCapacity(capacity);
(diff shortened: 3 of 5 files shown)


liulx20 (commented):
The incident EdgeTables are resized here, but they do not appear to be marked dirty. `StorageUpdateInterface::BatchAddVertices()` only marks the vertex table dirty, while `DumpDirtyAndReopen()` persists/reopens EdgeTables based on `dirty_.IsEdgeDirty(index)`.

The new regression test masks this because it creates an edge before checkpointing; that CREATE marks the EdgeTable dirty and causes the resized CSR to be persisted. A sequence such as:

`COPY vertices -> CHECKPOINT/close -> reopen -> CREATE edge using a newly added vertex`

can therefore reopen the old CSR capacity and still hit the original problem.

Could we mark every incident EdgeTable affected by `SyncIncidentEdgeCapacity()` dirty, and adjust the regression to checkpoint/reopen immediately after the vertex COPY, before creating the edge?

liulx20 (approved):

PR #995, opened 2026-09-01
ci: support weekly wheel artifacts and manual build selection

### LINKED ISSUE #994: Support weekly wheel artifacts and manual build selection
## Upload artifacts for scheduled weekly tests

The scheduled `Build Weekly Wheels` run builds and tests wheels, but its artifacts are not currently retained. Scheduled weekly runs should upload the generated artifacts for the enabled platforms so they can be downloaded and inspected after CI finishes.

Artifact uploads should remain limited to the official `alibaba/neug` repository.

## Select platform and Python version for manual runs

Manual `workflow_dispatch` runs should provide selectors for both the target platform and Python version. Users should be able to build all supported targets or select one platform and one Python version for a focused run.

The default selections should preserve the existing full weekly build behavior.

Spockkk0225:
Closes #994

## Upload artifacts for scheduled weekly tests

Allow scheduled `Build Weekly Wheels` runs to upload the generated artifacts for the enabled Linux x86_64, Linux ARM64, and macOS ARM64 builds. The existing official-repository restriction remains unchanged.

## Select platform and Python version for manual runs

Add `workflow_dispatch` selectors for the target platform and Python version. Both selectors default to `all`, preserving the existing full weekly build behavior while allowing focused manual builds.


lnfjpt (approved):
LGTM

PR #996, opened 2026-09-02
fix: align vector column size with VID slots

### LINKED ISSUE #981: [Bug] HNSW index + alternating MERGE create/update crashes at column.h:488 set_any
## Bug Summary

With an HNSW index on a `FLOAT[32]` vector column, an alternating workload of `MERGE ... ON CREATE` (insert new row) and `MERGE ... ON MATCH SET` (update an existing row) crashes after a few thousand iterations with:

```
ERR_INTERNAL_ERROR: Runtime exception: Index out of range at .../include/neug/utils/property/column.h:488 func: set_any, Error Code: 1012
```

## Environment

- **NeuG version:** 0.1.3 (build-py312)
- **OS:** macOS
- **Python:** 3.12
- **Extension:** `vector_search`

## Minimal Reproduction

```python
#!/usr/bin/env python3
import neug

VEC = [0.1] * 32

db = neug.Database(":memory:", mode="w")
conn = db.connect()

try:
    conn.execute("LOAD vector_search;")
    conn.execute(
        "CREATE NODE TABLE IF NOT EXISTS T("
        "id STRING PRIMARY KEY, vector FLOAT[32], text VARCHAR(65535));"
    )
    conn.execute(
        "CREATE INDEX T_hnsw_idx ON T USING HNSW (vector) WITH (metric = 'cosine');"
    )

    for n in range(1, 4101):
        conn.execute(
            """
            MERGE (v:T {id: $id})
                ON CREATE SET v.vector = $v, v.text = $text
                ON MATCH  SET v.vector = $v, v.text = $text;
            """,
            parameters={"id": f"r-{n}", "v": VEC, "text": f"t{n}"},
        )
        conn.execute(
            """
            MERGE (v:T {id: $id})
                ON MATCH SET v.vector = $v, v.text = $text;
            """,
            parameters={"id": f"r-{n-20}", "v": VEC, "text": "updated"},
        )
finally:
    conn.close()
    db.close()
```

## Expected Behavior

The workload completes without error.

## Actual Behavior

Crashes at `n=4077` on the first `MERGE` (the `ON CREATE` path) of iteration 4077:

```
ERR_INTERNAL_ERROR: Runtime exception: Index out of range at .../include/neug/utils/property/column.h:488 func: set_any, Error Code: 1012
```

Peak RSS at crash is only ~580 MB and the system has several GB of free memory, so this is **not an OOM**.

## Isolation Tests

I ran four variant workloads to narrow down the trigger:

| Scenario | Load `vector_search` | HNSW index | Result |
|---|---|---|---|
| Insert-only (`CREATE` new rows) | Yes | Yes | ✅ 5000 rows OK |
| Update-only (`MERGE ... ON MATCH SET`) | Yes | Yes | ✅ 5000 updates OK |
| Vector column, **no HNSW index** | Yes | No | ✅ 4100 iterations OK |
| Plain `STRING/VARCHAR` table (no `vector_search`) | No | No | ✅ 10000 iterations OK |

Only the **HNSW index + alternating insert/update** combination triggers the crash.

## Notes

- The crash happens on the `ON CREATE` path of a `MERGE` after many prior updates, suggesting the HNSW index state becomes inconsistent when rows are updated.
- Removing the HNSW index prevents the crash while keeping the vector column, so the issue appears to be in the index maintenance path rather than the vector column storage itself.

shirly121:
## Summary

- define IndexIDAccessor::size() as the number of addressable VID mapping slots
- make VecColumn::size() return its accessor size so it follows the row-capacity semantics of other property columns
- expose VecColumn::buffer_size() for the separately growing index-ID payload buffer
- retain the existing Table resize and VertexTable capacity logic
- add a focused regression test for property-column growth after vector updates

## Root cause

VecColumn::size() previously returned the size of its versioned vector payload buffer. Repeated updates allocate new index IDs, so this buffer can grow beyond the vertex-table capacity. When the vector column is first, Table::size() then incorrectly reports that the table is already large enough and ordinary row-addressed columns are not resized before the next vertex insert.

The fix separates the two concepts: VecColumn::size() now reports VID mapping slots through IndexIDAccessor::size(), while buffer_size() reports the vector payload slots used by index IDs. The accessor continues to maintain its own capacity during upserts, and VecColumn::resize() retains its existing payload-buffer behavior.

## Validation

- NeuG Python binding and vector_search extension build completed successfully
- focused 4096-to-4097 capacity-boundary regression: passed
- original 4,100-iteration alternating MERGE create/update reproduction: passed
- passed the previous failure point at iteration 4,077
- completed with 4,120 vertices and no idx >= size_ exception

Fixes #981

### DIFF
--- a/include/neug/storages/index/index_id_accessor.h
+++ b/include/neug/storages/index/index_id_accessor.h
@@ -33,6 +33,8 @@ class IndexIDAccessor : public Module {
  public:
   ~IndexIDAccessor() override = default;
 
+  // Number of addressable VID slots in the VID-to-index-ID mapping.
+  virtual size_t size() const = 0;
   virtual index_id_t GetIndexIDByVID(vid_t vid) const = 0;
   virtual vid_t GetVIDByIndexID(index_id_t index_id) const = 0;
   virtual index_id_t GetNextIndexID() const = 0;
@@ -62,7 +64,7 @@ class DefaultIndexIDAccessor final : public IndexIDAccessor {
         next_index_id_(std::make_shared<std::atomic<index_id_t>>(0)) {}
   ~DefaultIndexIDAccessor() override = default;
 
-  size_t size() const {
+  size_t size() const override {
     return vid_to_index_id_
                ? vid_to_index_id_->GetDataSize() / sizeof(index_id_t)
                : 0;
@@ -122,6 +124,7 @@ class VecColumnBackedIndexIDAccessor final : public IndexIDAccessor {
   explicit VecColumnBackedIndexIDAccessor(IndexIDAccessor& offset_accessor)
       : offset_accessor_(offset_accessor) {}
 
+  size_t size() const override;
   index_id_t GetIndexIDByVID(vid_t vid) const override;
   vid_t GetVIDByIndexID(index_id_t index_id) const override;
   index_id_t GetNextIndexID() const override;
--- a/include/neug/utils/property/vec_column.h
+++ b/include/neug/utils/property/vec_column.h
@@ -92,7 +92,7 @@ class VecColumn : public ColumnBase {
   size_t size() const override;
   // Number of index-ID slots currently allocated in the vector payload
   // buffer. This is independent from size(), which reports VID slots.
-  size_t buffer_capacity() const;
+  size_t buffer_size() const;
   void resize(size_t size) override;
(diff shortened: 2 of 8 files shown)


zhanglei1949 (approved):
LGTM

PR #1013, opened 2026-09-07
fix: validate VARCHAR maximum length

### LINKED ISSUE #1006: [BUG] VARCHAR(65536) silently stores non-empty values as empty strings
## Environment

- NeuG Python package: `neug==0.2.0`
- Python: `3.12.11`
- OS: macOS 15.7.7, arm64
- Also reproduced through the released `@graphscope-neug/neug@0.2.0` Node.js package.

## Problem

The data type documentation says that `VARCHAR(max_length)` supports a maximum length of 65536. NeuG accepts `VARCHAR(65536)` in DDL, but any non-empty value written to that column is immediately read back as an empty string. No error or warning is returned.

`VARCHAR(65535)` works as expected with the same schema and value.

Documentation: https://github.com/alibaba/neug/blob/765008c35f3cf5dc68e636aad0c3851c9cedd502/doc/source/cypher_manual/data_types.md#L72-L75

## Minimal reproduction

```python
import tempfile
from pathlib import Path

import neug


def round_trip(root: Path, declared_length: int):
    db_path = root / str(declared_length)
    db = neug.Database(
        str(db_path), mode="read-write", checkpoint_on_close=False
    )
    conn = db.connect()
    conn.execute(
        f"CREATE NODE TABLE Item("
        f"id INT64 PRIMARY KEY, value VARCHAR({declared_length}));"
    )
    conn.execute("CREATE (:Item {id: 1, value: 'sentinel'});")
    before_checkpoint = list(
        conn.execute("MATCH (item:Item) RETURN item.value;")
    )[0][0]
    conn.execute("CHECKPOINT;")
    conn.close()
    db.close()

    db = neug.Database(
        str(db_path), mode="read-only", checkpoint_on_close=False
    )
    conn = db.connect()
    after_reopen = list(
        conn.execute("MATCH (item:Item) RETURN item.value;")
    )[0][0]
    conn.close()
    db.close()
    return before_checkpoint, after_reopen


with tempfile.TemporaryDirectory() as temporary_root:
    root = Path(temporary_root)
    print("65535:", round_trip(root, 65535))
    print("65536:", round_trip(root, 65536))
```

Output:

```text
65535: ('sentinel', 'sentinel')
65536: ('', '')
```

## Expected behavior

Either:

1. `VARCHAR(65536)` preserves values as documented, including across checkpoint/reopen; or
2. DDL rejects 65536 and the documented maximum is changed to 65535.

Accepting the schema and silently replacing a non-empty value with an empty string should not be possible.

## Suspected cause

This is an inference from the current source: `TypedColumn<std::string_view>` stores `width_` as `uint16_t`, restores it with `static_cast<uint16_t>`, and `string_item.length` is a 16-bit field. A declared width of 65536 therefore appears able to wrap to zero.

- https://github.com/alibaba/neug/blob/765008c35f3cf5dc68e636aad0c3851c9cedd502/include/neug/utils/property/column.h#L238-L262
- https://github.com/alibaba/neug/blob/765008c35f3cf5dc68e636aad0c3851c9cedd502/include/neug/utils/property/column.h#L598-L601

A regression test should cover the boundary before checkpoint and after read-only reopen, and ensure unsupported widths fail during DDL instead of during data storage.

shirly121:
## Summary

- reject explicit `VARCHAR` lengths outside the supported `1..65535` range during compiler type binding
- cover `STRING` and `VARCHAR` boundary behavior in CREATE/ALTER DDL, including LIST/ARRAY element types and checkpoint/reopen
- correct the documented maximum length from 65536 to 65535

## Testing

- `python3 -m pytest -q tests/test_schema.py` (37 passed)
- targeted string length and persistence tests (16 passed)
- compiler `gopt_test` suite (178 passed)

Fixes #1006

### DIFF
--- a/tools/python_bind/tests/test_schema.py
+++ b/tools/python_bind/tests/test_schema.py
@@ -140,6 +140,79 @@ def test_create_node_table_with_default_value(tmp_path):
     db.close()
 
 
+def test_string_type_length_boundaries(tmp_path):
+    db = Database(db_path=str(tmp_path), mode="w", checkpoint_on_close=False)
+    conn = db.connect()
+
+    conn.execute(
+        "CREATE NODE TABLE StringLengths("
+        "id INT64 PRIMARY KEY,"
+        "string_value STRING,"
+        "min_value VARCHAR(1),"
+        "max_value VARCHAR(65535));"
+    )
+    conn.execute("ALTER TABLE StringLengths ADD altered_string STRING;")
+    conn.execute("ALTER TABLE StringLengths ADD altered_min VARCHAR(1);")
+    conn.execute("ALTER TABLE StringLengths ADD altered_max VARCHAR(65535);")
+    conn.execute(
+        "CREATE (:StringLengths {"
+        "id: 1, string_value: 'string', min_value: 'm', "
+        "max_value: 'sentinel', altered_string: 'altered', "
+        "altered_min: 'a', altered_max: 'persisted'});"
+    )
+    query = (
+        "MATCH (n:StringLengths {id: 1}) "
+        "RETURN n.string_value, n.min_value, n.max_value, "
+        "n.altered_string, n.altered_min, n.altered_max;"
+    )
+    expected = [["string", "m", "sentinel", "altered", "a", "persisted"]]
+    assert list(conn.execute(query)) == expected
+
+    conn.execute("CHECKPOINT;")
+    conn.close()
+    db.close()
+
+    db = Database(db_path=str(tmp_path), mode="r", checkpoint_on_close=False)
+    conn = db.connect()
+    assert list(conn.execute(query)) == expected
+
+    conn.close()
+    db.close()
+
+
+@pytest.mark.parametrize("ddl_path", ["create", "alter"])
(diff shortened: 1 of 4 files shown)


tools/python_bind/tests/test_schema.py
```diff
@@ -140,6 +140,79 @@ def test_create_node_table_with_default_value(tmp_path):
...
 
+def test_string_type_length_boundaries(tmp_path):
+    db = Database(db_path=str(tmp_path), mode="w", checkpoint_on_close=False)
+    conn = db.connect()
+
+    conn.execute(
+        "CREATE NODE TABLE StringLengths("
+        "id INT64 PRIMARY KEY,"
+        "string_value STRING,"
+        "min_value VARCHAR(1),"
```
longbinlai, line 151: 这里的 min_value, max_value 是什么意思啊

shirly121: 就是 string 长度范围为 [1, 65535]；这里用两个边界值测试

longbinlai (approved):
LGTM

PR #1021, opened 2026-09-08
fix: use strict boolean predicates for three-valued logic

### LINKED ISSUE #1019: Fix three-valued AND/OR semantics with NULL
## Three-Valued Boolean Logic

In Boolean operations, `null` represents an unknown value and produces `null` when the result cannot be determined. However, when `false` participates in an `AND` operation or `true` participates in an `OR` operation, the result is deterministic and must be `false` or `true`, respectively.

| `a` | `b` | `a AND b` | `a OR b` | `a XOR b` |
|---|---|---|---|---|
| `false` | `null` | `false` | `null` | `null` |
| `true` | `null` | `null` | `true` | `null` |
| `null` | `null` | `null` | `null` | `null` |

In addition, `NOT null -> null`.

### Current Issue

```cypher
RETURN null AND false;
RETURN null OR false;
```

| Expression | Expected result | Current result |
|---|---|---|
| `null AND false` | `false` | `null` |
| `null OR false` | `null` | `false` |

Spockkk0225:
## Summary

- Correct `AND` and `OR` evaluation when either operand is `NULL`.
- Make the result independent of operand position while preserving deterministic outcomes for `false AND x` and `true OR x`.
- Apply the same three-valued logic to record, vertex, and edge expressions.

## Tests

- Added `NULL` cases to the existing parameterized Boolean-expression tests. Together with the existing cases, they now fully cover all combinations of `true`, `false`, and `NULL` for `AND` and `OR`, as well as `NOT NULL`.
- Added unit-test coverage for `Value::IsFalse()` with both `true` and `false` values.

## Docs

- The existing Logical Operators documentation already covers the complete three-valued truth tables for `AND`, `OR`, and `NOT`. No documentation change is needed.

Closes #1019


extension/parquet/CMakeLists.txt
```diff
@@ -156,6 +156,7 @@ if(BUILD_TEST)
     add_library(neug_parquet_carquet_impl STATIC
         ${CMAKE_CURRENT_SOURCE_DIR}/src/carquet/input_adapter.cc
         ${CMAKE_CURRENT_SOURCE_DIR}/src/carquet/output_adapter.cc)
+    add_dependencies(neug_parquet_carquet_impl neug_proto)
```
Spockkk0225: 这个修复和 https://github.com/alibaba/neug/pull/1023 一致，到时候会统一merge

shirly121 (dismissed):

extension/parquet/CMakeLists.txt
Spockkk0225: 已删除冗余的依赖

shirly121 (approved):

PR #1023, opened 2026-09-08
fix: enforce complete file predicates within readers

luoxiaojian:
File readers could reject or incorrectly evaluate LOAD predicates involving arithmetic, casts, NULLs or parameters. Pass the complete predicate through the existing reader interface and make each reader responsible for filtering before returning the requested columns.

CSV and JSON use the shared DataChunk filter, which now reuses NeuG's execution expressions instead of separate limited evaluators. The Arrow Parquet reader retains native expression filtering when conversion and binding succeed; otherwise it evaluates decoded chunks before projection, in both batch and full reads. CAST expressions use this fallback because the previous Arrow converter discarded the conversion itself. IO and evaluation errors still propagate.

Query parameters are collected from DataSource predicates, including nested expressions, and supplied on each execution without mutating cached plan state. Preserve unconsumed filters above table functions and correct Boolean NULL/short-circuit semantics. Shared chunk aggregation validates layouts and preserves nested values and NULLs; CSV now uses the same helpers.

The planner's original complete-predicate forwarding is preserved. There is no global pushdown whitelist or new capability-query interface. Carquet implementation, build, submodule and pkg-config handling are unchanged.

Regression coverage includes native Arrow arithmetic filtering, reader-local fallback, hidden filter columns, batch/full reads, empty results, conversion failures, repeated parameters, parameters inside lists, nested CAST/CASE/functions, Boolean truth tables, NaN and nested chunk aggregation. LOAD documentation describes the reader contract.

## Validation

- Release build with Python, Parquet and HTTPFS passed; original build configuration restored and verified.
- 67/67 related CTest entries and 114/114 execution tests passed. HTTPFS tests passed with loopback-server permissions.
- Final broad Python query/load/array/documentation/Parquet/export regressions: 299 passed, 22 skipped. Focused provider regressions passed 59/59.
- With all extensions disabled: build, 33/33 reader/expression CTest entries, 114/114 execution tests and 39/39 JSON/CSV Python regressions passed.
- clang-format 10, isort, Black, flake8 with project configuration and whitespace checks passed. Tests and doc/source changes satisfy the local PR-completeness rules.

The existing skips cover unavailable LDBC/comprehensive datasets, credential-dependent HTTPFS exports, an existing optional-match issue and PyArrow's NULL fixed-size-list limitation. No new regression is skipped. Remote CI has not been run.


zhanglei1949 (commented):
P1: The exact PR head fails `extension_tests_default`: `input_adapter.cc` and `output_adapter.cc` include `result.h` before `error.pb.h` has been generated in a clean parallel build. This static target depends only on `neug::carquet`, while `neug_utils` is the target that depends on `neug_proto`. Add a direct dependency on `neug_proto` (or an equivalent generated-header dependency) before merging.

tools/python_bind/tests/test_query.py
```diff
@@ -1449,3 +1449,97 @@ def test_multi_label2(tinysnb):
...
+    path.write_text('{"id":1}\n{"id":2}\n{"id":3}\n', encoding="utf-8")
+    source = f"LOAD FROM '{path}'" + (" WITH id" if with_clause else "")
+    query = f"{source} WHERE id + 1 > $minimum RETURN id ORDER BY id;"
+    for minimum, expected in [(3, [[3]]), (1, [[1], [2], [3]]), (4, [])]:
+        assert list(conn.execute(query, parameters={"minimum": minimum})) == expected
+
+
+@pytest.mark.parametrize("file_format", ["csv", "jsonl"])
+def test_load_reader_binds_parameters_inside_list(empty_db, tmp_path, file_format):
+    _, conn = empty_db
```
shirly121, line 1535: 这个测试在 debug 模式下报错：

  Assertion failed: (type.id() == DataTypeId::kList),
  function GetChildType, file types.cc, line 71.
  Fatal Python error: Aborted

执行路径：

1. 文件过滤器逐行计算表达式
     src/utils/io/read/common/row_expression_filter.cc:149

  2. 执行列表表达式 [$first, $last]
     src/execution/expression/exprs/struct_expr.cc:88

     const DataType& child = ListType::GetChildType(type_);

  3. GetChildType() 发现传入的 type_ 不是 kList，触发 Debug assertion
     src/common/types.cc:70

     const DataType& ListType::GetChildType(const DataType& type) {
       assert(type.id() == DataTypeId::kList);  // 此处 abort
     }

BingqingLyu:
Non-blocking question: in the fallback path `scanOptions->filter` stays `literal(true)` and projection widens to all entry-schema columns, so both row-group pruning and projection pushdown are lost — a remote `oss://`/`s3://` read transfers essentially the whole file before filtering locally. The trigger set is also broader than `CAST` (query params, `IS NULL`, `STARTS WITH`, `IN`, `CASE`, unbindable exprs). Reasonable correctness-first tradeoff for this PR — is there a follow-up plan to push some of these down natively (e.g. safely-castable `CAST`, params), or to push a convertible subset purely for row-group pruning?

tools/python_bind/tests/test_query.py
luoxiaojian: fixed

shirly121 (approved):

PR #1025, opened 2026-09-08
docs: improve C++ API reference

longbinlai:
## Summary

- add concrete descriptions for all C++ `QueryResult` typed getters
- state each getter’s required source column types and that unsupported types throw `neug::exception::RuntimeError`
- document the supported widening and boolean conversions separately from the type preconditions
- document both column-index and column-name overloads in the public header so regenerated API docs keep the descriptions
- align C++ API navigation as Database, Connection, Service, then Query Result, including the generator source configuration

## Validation

- `make -C doc doxygen`
- `make -C doc html-only` (succeeds with pre-existing documentation warnings)
- `git diff --cached --check`

### DIFF
--- a/doc/source/_scripts/cpp_key_classes.json
+++ b/doc/source/_scripts/cpp_key_classes.json
@@ -5,12 +5,12 @@
   "key_classes": [
     "neug::NeugDB",
     "neug::Connection",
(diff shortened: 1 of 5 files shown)


zhanglei1949 (approved):
LGTM

PR #1032, opened 2026-09-08
fix: handle NULL ordering and DURATION comparison semantics

### LINKED ISSUE #1020: Fix ordering and DURATION comparison semantics with NULL
## Binary Comparisons and Null Predicates

For every binary comparison operator (`=`, `<>`, `<`, `<=`, `>`, and `>=`), the result is `null` if either operand is `null`.

```cypher
RETURN null = null, null <> null,
       null = 1, null <> 1,
       1 < null, null >= 1;
// null, null, null, null, null, null
```

Therefore, binary comparison operators cannot be used to determine whether an expression is `null`. Use the dedicated `IS NULL` and `IS NOT NULL` predicates instead.

```cypher
RETURN null IS NULL, null IS NOT NULL;
// true, false
```

### Current Status

NeuG handles these semantics correctly.

## Null Ordering in `ORDER BY`

`ORDER BY` places `null` after every non-null value. Consequently, `null` appears last in ascending order and first in descending order.

```cypher
UNWIND [1, null, 2] AS value
RETURN value
ORDER BY value ASC;
// 1, 2, null
```

```cypher
UNWIND [1, null, 2] AS value
RETURN value
ORDER BY value DESC;
// null, 2, 1
```

### Current Issue

```cypher
UNWIND [3, 1, null, 4, 2] AS value
RETURN value ORDER BY value ASC;
```

- Expected result: `1, 2, 3, 4, null` (`null` sorts last in ascending order).
- Current result: `1, 3, null, 2, 4`.

Currently, `null` causes the values to be sorted in separate segments instead of producing the correct overall order.

## Duration Comparisons

The lengths of calendar units such as months and days are not intrinsically equivalent. Some duration semantics therefore do not define a direct ordering across these units and may return `null` for such comparisons.

### Current Status

NeuG intentionally follows DuckDB's fixed-base duration semantics. Duration units are normalized using fixed conversion factors, for example `1M = 30D`, and the normalized values are compared directly.

```cypher
RETURN duration('P1M') < duration('P30D');
// false
```

This definite result is NeuG's selected duration semantics and is not an implementation issue. The documentation should briefly note this choice and clarify that, although `DURATION` and `DATE` calculations are correct within their respective domains, combining a `DATE` with a `DURATION` does not guarantee calendar-aware equivalence; for example, adding `P1M` and `P30D` may produce different dates.

Spockkk0225:
## Summary

- Place `NULL` values last in ascending order and first in descending order.
- Keep sorting deterministic when multiple `NULL` values are present.
- Document NeuG's fixed-base `INTERVAL` normalization and its interaction with `DATE` arithmetic.

## Tests

- Added regression coverage for both ascending and descending `NULL` placement in a single test.

## Docs

- Documented the `ORDER BY` behavior for `NULL` values with ascending and descending examples.
- Documented the fixed conversion bases used for `INTERVAL` comparison and showed why equivalent intervals can produce different results when combined with a `DATE`.

Closes #1020


doc/source/cypher_manual/data_types.md
```diff
@@ -97,6 +97,26 @@ Alternatively, you can use STRING to specify the character type directly; STRING
...
+- 1 hour = 60 minutes
+- 1 minute = 60 seconds
+- 1 second = 1,000 milliseconds
+- 1 millisecond = 1,000 microseconds
+
+Fixed normalization ensures consistent calculations within the `INTERVAL` type,
+but calculations involving both `DATE` and `INTERVAL` can produce different
+results for otherwise equivalent intervals:
+
+```cypher
```
shirly121, line 121: 这里是不是也可以加一个 interval(1 year) = interval(12 * 30 * 24 hours) 比较的例子

tools/python_bind/tests/test_query.py
```diff
@@ -144,6 +144,22 @@ def test_aggregate_over_all_null_input(empty_db):
     assert list(result) == [[None, None, None, 0, []]]
```
shirly121, line 147: 测试里也只有 orderby null 的测试，没有 interval 和 date 比较？

doc/source/cypher_manual/data_types.md
Spockkk0225: 已补充
https://github.com/alibaba/neug/pull/1032/commits/ae3237b8f8e96a191c2cfe247b2851597d7fc6b3

tools/python_bind/tests/test_query.py
Spockkk0225: interval和date已经有c++的测试，现在补充了python测试
https://github.com/alibaba/neug/pull/1032/commits/ae3237b8f8e96a191c2cfe247b2851597d7fc6b3

shirly121 (approved):

PR #1033, opened 2026-09-08
fix: avoid allocating unused 1 GiB WAL files

### LINKED ISSUE #1000: Improve local WAL lifecycle for unused 1 GiB sparse files
## Problem

`LocalWalWriter` preallocates every new WAL file to `TRUNC_SIZE` (1 GiB) with `ftruncate`. The file is sparse on filesystems such as APFS, but `close()` leaves the full logical length in place, including when the writer appended no records. A later read-write database open selects the next versioned `thread_<slot>_<version>.wal` rather than reusing the empty file.

Repeated short-lived embedded read-write opens therefore accumulate 1 GiB sparse WAL files. In one small embedded workload, 41 reopen/query/validation cycles produced:

- 41 WAL files;
- approximately 41 GiB apparent size;
- approximately 208 KiB physical allocation.

The physical allocation is initially small, but the apparent size is misleading to users and tooling. Backup, archive, filesystem migration, or copy paths that do not preserve holes can materialize the unused ranges and turn the apparent size into real disk consumption.

This is independent of Service mode: it is reproducible with the embedded `Database` API opened repeatedly in read-write mode.

## Resolution scope and remaining limitation

PR #1033 addresses the reported preallocation problem by creating WAL files only on the first non-empty append and storing their actual durable contents instead of reserving 1 GiB. Repeated read-write opens without mutations therefore create no WAL files, while mutation-bearing WAL files have a logical size proportional to their records.

PR #1033 does not bound total WAL growth for a workload that keeps `checkpoint_on_close=false`, performs mutations, and never requests a manual checkpoint. All WAL versions in the current checkpoint epoch remain necessary for recovery and are reclaimed only after a newer checkpoint is durably published. #1037 tracks the follow-up design for triggering automatic checkpointing from accumulated WAL bytes and reclaiming the retired epoch through the existing checkpoint GC.

## Relevant implementation

- `include/neug/transaction/wal/local_wal_writer.h` fixes `TRUNC_SIZE` at `1ul << 30`.
- `src/transaction/wal/local_wal_writer.cc` calls `ftruncate(fd_, TRUNC_SIZE)` when constructing each writer.
- `LocalWalWriter::close()` closes the descriptor without removing an unused file or reducing its logical length to the bytes actually written.

## Requested lifecycle improvement

Please make retired local WAL files reflect their actual durable content without weakening recovery guarantees. Possible implementations include:

1. remove a newly created WAL file when no record was appended;
2. safely truncate a retired non-empty WAL file to its used length;
3. create or extend the file lazily instead of reserving 1 GiB at writer construction;
4. garbage-collect empty retired WAL versions after checkpoint publication.

The implementation should choose the option that remains correct across WAL epoch rotation, recovery, concurrent read-only processes, and crash interruption.

## Acceptance checks

- Repeated embedded read-write open/close operations with no mutations do not increase retained WAL file count or apparent WAL bytes without bound.
- A small committed mutation remains recoverable after close/reopen and leaves a WAL whose logical length is proportional to the data written rather than the 1 GiB reservation.
- Checkpoint and WAL epoch rotation still retain every file required by a live checkpoint reference and remove only files proven unnecessary.
- Sparse-file copy behavior remains covered, but lifecycle correctness does not depend on every downstream copy/archive tool preserving holes.
- Diagnostics or tests report both apparent bytes and allocated bytes so regressions are visible.

## Reproduction sketch

```js
for (let i = 0; i < 40; i += 1) {
  const db = new Database({
    databasePath,
    mode: 'w',
    checkpointOnClose: false,
  });
  const conn = db.connect();
  conn.execute('MATCH (n) RETURN count(n);').close();
  conn.close();
  db.close();
}
```

After the loop, inspect both the logical length and allocated blocks of `wal/*/*.wal`.

zhanglei1949:
## Related Issues
fix #1000

Follow-up: #1037

## What does this PR do?
Avoid creating and preallocating a 1 GiB local WAL file when a read-write database session performs no mutations. Local WAL records continue to use ordinary file APIs and preserve the existing per-append durability semantics and legacy WAL format.

## What changes in this PR?
- Delay local WAL file creation until the first non-empty append.
- Create versioned WAL files exclusively without overwriting existing files.
- Remove the old 1 GiB `ftruncate` preallocation and use complete positional write loops for the record and one zeroed terminator.
- Synchronize a newly created WAL directory entry before writing transaction data, then synchronize the WAL file on every successful append.
- Stop using a writer after any append I/O failure, so an interrupted tail cannot be overwritten from a stale offset.
- Keep recovery mmap read-only. Recovery now retains the previously committed prefix when the final append is interrupted, while still rejecting an illegal negative length.
- Add process-level fault coverage for interruption after a full record write, interruption during a partial record write, and writer reuse after an I/O failure.
- Add regression coverage for repeated read-write open/close sessions without mutations, exact WAL file size, WAL replay, version selection, and repeated mutation/recovery cycles without close-time checkpoints.

The old writer already used `write()` plus `ftruncate`; this PR does not replace a writable mmap path. The parser continues to use mmap only for read-only recovery.

## Performance
A local WAL append microbenchmark compares the PR path against the base implementation with 128-byte records and the same per-append `F_FULLFSYNC` durability. On the latest macOS run:

- Steady append: base 276.5/s, p50 3.89 ms, p99 6.01 ms; PR 264.3/s, p50 3.97 ms, p99 4.88 ms.
- First append with an empty directory: base 247.5/s, p50 4.01 ms, p99 5.12 ms; PR 243.4/s, p50 4.08 ms, p99 5.11 ms.
- First append with 1,000 existing WAL files: base 149.3/s, p50 6.51 ms, p99 8.07 ms; PR 172.9/s, p50 5.88 ms, p99 6.62 ms.

A `pwritev` variant did not improve steady append or first-write latency in repeated runs, and it requires an availability fallback for the project's macOS 10.15 deployment target, so it is not included.

## Remaining limitation
This PR reduces each WAL file to its actual durable contents, but it does not bound total WAL growth when `checkpoint_on_close=false` and a mutation workload never requests a manual checkpoint. WAL files in the current checkpoint epoch remain required for recovery and are reclaimed only after a newer checkpoint is durably published. #1037 tracks WAL-byte-based automatic checkpointing and subsequent retired-epoch garbage collection.

## Validation
- Built the `transaction_test` target successfully.
- WAL parser/replay tests: 25 passed, 1 skipped because the platform cannot lower `RLIMIT_AS`.
- Selected Python 3.13 WAL regressions: 4 passed.
- `make format-check` and `git diff --check` passed.
- `graphify .` was attempted, but semantic extraction could not run because no LLM API key is configured for the repository's non-code files.


src/transaction/wal/local_wal_writer.cc
```diff
@@ -53,119 +135,100 @@ LocalWalWriter::~LocalWalWriter() noexcept {
...
     file_used_ = 0;
 #ifdef _WIN32
-    if (_close(fd) != 0) {
+    const int close_result = _close(fd);
 #else
-    if (::close(fd) != 0) {
+    const int close_result = ::close(fd);
 #endif
+    if (close_result != 0) {
       THROW_IO_EXCEPTION("Failed to close file" + std::string(strerror(errno)));
```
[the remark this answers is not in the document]
zhanglei1949: Thanks, fixed locally. The diagnostic now reads `Failed to close WAL file: <system error>`, adding both the missing separator and WAL context. The updated target builds successfully; the full `test_db_svc` suite passes 42/42, and the focused WAL regression suite passes 13 tests with one platform-specific skip. This fix will be included in the next push.

liulx20:
Reviewed at e8cd7139f71fc09ba43f8fb56493796f18998603. Two concerns:

**[P1] A crash between the record write and the terminator write prevents recovery of previously committed records.**

In [local_wal_writer.cc:220–221](https://github.com/alibaba/neug/blob/e8cd7139f71fc09ba43f8fb56493796f18998603/src/transaction/wal/local_wal_writer.cc#L220-L221), the first `WriteAllAt` overwrites the old terminator, and the second writes its replacement. If the process exits between them, the file ends without a terminator. The new parser rejects the entire file with `truncated header or missing terminator`, so database opening fails even though earlier records were successfully appended and synchronized.

This was reproduced in an isolated C++ harness compiling this revision's actual writer/parser sources, with unrelated dependencies stubbed: append record 1 successfully, then intercept `pwrite` and call `_exit(73)` immediately after the write of record 2 completes, before its terminator is written. The parent process then opens the WAL with `LocalWalParser`:

```text
After acknowledged append: last_ts=1
Child exit=73
Recovery failed: ... truncated header or missing terminator
```

Please make the append/recovery protocol handle an interrupted tail while preserving recovery of the previously committed prefix, and add fault-injection coverage inside `append()` (including partial writes). The added process-exit test exits after a completed commit and before the next transaction commits, so it does not exercise this window. Merely combining the writes would not guarantee crash-atomic persistence.

**Performance: please add a before/after benchmark for small transactions.**

The normal append path now makes two write syscalls rather than one and extends the file on each append; the first append per writer also scans the directory and synchronizes it. These may affect small-transaction throughput, commit p50/p99, and first-write latency with many existing WAL files. Per-append file synchronization already existed, so this is a potential regression to measure, not a measured result.

Could the record and terminator be submitted together using `pwritev` on supported platforms, with correct short-write handling and a platform fallback? That would normally avoid the extra write syscall without copying the payload into a combined buffer.

Also, relative to the PR base, the old writer already uses `write()` plus `ftruncate`; recovery still uses read-only `mmap`. Please clarify the description's “replace mmap/ftruncate-based WAL writes” wording.

Validation here was source review plus the isolated fault-injection reproduction, not the full project test suite.

src/transaction/wal/local_wal_writer.cc
```diff
@@ -53,119 +131,101 @@ LocalWalWriter::~LocalWalWriter() noexcept {
...
-#else
-  // if (fsync(fd_) != 0) {
-  if (fdatasync(fd_) != 0) {
-    THROW_IO_EXCEPTION("Failed to fsync wal file " +
-                       std::string(strerror(errno)));
+
+  const std::array<char, sizeof(WalHeader)> terminator{};
+  // Preserve the legacy framing without the old zero-filled preallocation:
+  // each append overwrites the previous terminator, then writes a new one.
+  WriteAllAt(fd_, data, length, file_used_);
```
longbinlai: **合并前必须修：写 WAL 过程中宕机会导致数据库无法重新打开。** 现在先覆盖旧的结束标记，再写新的结束标记。如果进程在两次写之间退出，WAL 最后就没有完整结束标记；重启时 parser 会拒绝整个文件，连前面已经成功提交的数据也无法恢复。这个问题是取消零填充预分配后新增的。建议恢复时丢弃最后一条未写完整的记录，保留前面完整记录，并补一个中途退出的回归测试。

src/transaction/wal/local_wal_writer.cc
```diff
@@ -53,119 +131,101 @@ LocalWalWriter::~LocalWalWriter() noexcept {
...
+  WriteAllAt(fd_, terminator.data(), terminator.size(), file_used_ + length);
+  SyncFile(fd_);
+  if (file_created &&
```
longbinlai: **合并前必须修：目录同步失败时，事务可能报错后又在重启时出现。** 当前代码先把完整 WAL 记录同步到磁盘，再同步目录；但只有全部成功后才更新 `file_used_`。如果目录同步失败，调用方会认为事务失败，磁盘上却已经存在一条可恢复的完整记录。重启会把它恢复出来，继续运行则可能从旧位置覆盖它。建议在写事务记录前完成新文件的目录同步，或在失败后禁止继续使用该 writer，并补目录同步失败测试。

src/transaction/wal/local_wal_writer.cc
zhanglei1949: Fixed in fd784794. Recovery now holds the last complete record as pending until a following record header or the zero terminator proves it durable; an interrupted final record is dropped while the previously confirmed prefix is retained. The process-level regression interrupts append both after the complete record write and during a partial write, then verifies that only the prior committed record is recovered. The WAL parser/replay suite passes 25 tests with one platform-specific RLIMIT_AS skip.

src/transaction/wal/local_wal_writer.cc
zhanglei1949: Fixed in fd784794. After O_EXCL creates the WAL file, create_file() now synchronizes the directory before append() writes any transaction bytes. If that synchronization fails, the descriptor is closed and the transaction record is never written. In addition, any append I/O exception poisons the writer (opened_=false), preventing reuse of file_used_. The process-level short-write case catches the I/O failure and verifies that a subsequent append is rejected.

zhanglei1949:
Addressed the performance and description feedback in fd784794 and the updated PR body. I ran a local 128-byte WAL append microbenchmark with the same per-append F_FULLFSYNC policy. Latest run: steady base 276.5/s (p50 3.89 ms, p99 6.01 ms) vs PR 264.3/s (p50 3.97 ms, p99 4.88 ms); empty-directory first append base 4.01/5.12 ms p50/p99 vs PR 4.08/5.11 ms; first append with 1,000 existing files base 6.51/8.07 ms vs PR 5.88/6.62 ms. A pwritev variant was neutral or slower in repeated runs and is unavailable without a fallback at the project macOS 10.15 deployment target, so I did not add that complexity. The description now explicitly states that the old writer already used write() plus ftruncate and that recovery mmap remains read-only.

longbinlai (dismissed):
LGTM

zhanglei1949:
TODO: Run LDBC benchmark to evaluate the performance.

src/transaction/wal/local_wal_writer.cc
```diff
@@ -53,120 +151,134 @@ LocalWalWriter::~LocalWalWriter() noexcept {
...
+
+  try {
+    if (fd_ == -1) {
+      create_file();
+    }
+
+    const std::array<char, sizeof(WalHeader)> terminator{};
+    // The zero-filled tail remains the recovery terminator. Preallocation
+    // keeps ordinary commits from extending i_size and flushing inode metadata.
+    // TODO(W1): LocalWalParser currently uses EOF to identify an interrupted
```
zhanglei1949, line 265: [P1] 这里已经准确指出了当前的恢复正确性缺口，但它仍存在于生产默认路径中：`ensure_file_size()` 先把文件扩到 64 MiB，进程若只写完新 header 和部分 payload 就退出，parser 的 `length <= end - ptr` 会把后面的预分配零区当作剩余 payload，从而把未提交/不完整记录恢复出来。现有故障注入测试用 `initial_file_size=0, file_growth_size=1` 绕开了默认预分配，因此不能覆盖这个场景。请在合并前取消这种 framing 下的预分配，或引入可验证的 commit trailer/checksum，并用生产默认参数补回归测试。

src/transaction/wal/local_wal_writer.cc
```diff
@@ -53,120 +151,134 @@ LocalWalWriter::~LocalWalWriter() noexcept {
...
-    break;
+    if (fd_ != -1) {
+      // Make the directory entry durable before WAL contents can become
+      // recoverable. If this fails, append() must not write the transaction.
+      if (!file_utils::fsync_directory(prefix)) {
+        (void) CloseFd(std::exchange(fd_, -1));
+        THROW_IO_EXCEPTION("Failed to sync wal directory " + prefix);
+      }
+      try {
+        ensure_file_size(initial_file_size_);
```
zhanglei1949, line 190: [P1] 当前首次 mutation 会在这里固定扩到 64 MiB，`close()` 也不会缩短文件。这样确实把旧的 1 GiB reservation 降低了 16 倍，并且无 mutation 时不再创建文件，但仍不满足 #1000 中‘小 mutation 的 WAL 逻辑长度与实际记录成比例’的验收项，也与 PR 描述里的 ‘actual durable contents’ 不一致。请明确二选一：缩小 #1000/PR 的承诺，只解决无 mutation 的空 WAL；或者恢复真正的精确长度实现，并解决相应 framing/durability 问题。

src/transaction/wal/local_wal_writer.cc
```diff
@@ -53,120 +151,134 @@ LocalWalWriter::~LocalWalWriter() noexcept {
...
+                _S_IREAD | _S_IWRITE);
 #else
-    fd_ = ::open(path.c_str(), O_RDWR | O_CREAT | O_TRUNC, 0644);
+    fd_ = ::open(wal_path.c_str(), O_RDWR | O_CREAT | O_EXCL, 0644);
 #endif
-    break;
+    if (fd_ != -1) {
+      // Make the directory entry durable before WAL contents can become
+      // recoverable. If this fails, append() must not write the transaction.
+      if (!file_utils::fsync_directory(prefix)) {
```
zhanglei1949, line 185: [P2] `O_EXCL` 已经创建了新版本文件；如果目录同步失败，这里只关闭 fd，没有删除刚创建的空文件。后面的 `ensure_file_size()` 失败路径也有同样问题。反复失败/重试会泄漏空版本，并最终逼近 65536 版本上限，和本 PR 避免无用 WAL 的目标冲突。请保存新路径并在创建阶段失败时做明确的清理与错误处理测试。

zhanglei1949 (commented):
结论：当前 head 能解决 #1000 核心复现中的‘无 mutation 的 RW 会话仍创建 1 GiB WAL’，但还不能按现有描述视为完整修复，也不建议按当前形态合并。当前实现重新引入 64 MiB 预分配，既不满足 Issue 中小 mutation WAL 逻辑长度应与实际记录成比例的验收项，又使新的 EOF 恢复判定在生产默认配置下失效。建议拆分：PR A 只做首次非空 append 时懒创建 WAL，保持既有 framing/parser；后续 PR 单独设计带可验证 commit trailer/checksum 的 crash-safe framing。另请同步更新 PR 描述和测试名称，避免继续声称当前文件只保存实际内容。

PR #1036, opened 2026-09-08
fix: correct null handling in CASE expressions

### LINKED ISSUE #1028: Fix CASE constant classification and matching semantics with NULL
## `CASE` Semantics

A simple `CASE` returns `null` when no branch matches and no `ELSE` clause is provided.

```cypher
RETURN CASE 1
    WHEN 2 THEN 'matched'
END;
// null
```

A generic `CASE` behaves the same way when no branch matches and no `ELSE` clause is provided.

```cypher
RETURN CASE
    WHEN false THEN 'matched'
END;
// null
```

A generic `CASE` treats only a `WHEN` expression whose result is `true` as a match. A branch whose result is either `false` or `null` is skipped, and evaluation continues with subsequent branches, following semantics similar to `WHERE`.

```cypher
RETURN CASE
    WHEN null THEN 'unknown branch'
    WHEN true THEN 'matched branch'
END;
// "matched branch"
```

In a simple `CASE`, a `WHEN null` alternative is interpreted as an `IS NULL` check on the `CASE` expression.

```cypher
WITH null AS value
RETURN CASE value
    WHEN null THEN 'null value'
    ELSE 'not matched'
END;
// "null value"
```

Although Neo4j treats a simple `CASE value WHEN null` as an equality comparison equivalent to `WHERE value = null`, and therefore does not consider `null` equal to `null`, a `WHEN null` branch can never be selected under those semantics and has no practical meaning. NeuG therefore chooses to retain Kùzu's semantics and rewrites the branch as `WHERE value IS NULL`.

A generic `CASE` with `IS NULL` can also be used to match a null value explicitly.

```cypher
WITH null AS value
RETURN CASE
    WHEN value IS NULL THEN 'null value'
    ELSE 'not null'
END;
// "null value"
```

An extended simple `CASE` may specify the `IS NULL` comparison predicate explicitly.

```cypher
WITH null AS value
RETURN CASE value
    WHEN IS NULL THEN 'null value'
    ELSE 'not null'
END;
// "null value"
```

### Current Issue

```cypher
RETURN CASE 1
    WHEN null THEN 'null value'
    ELSE 'not matched'
END;
```

- Expected result: `"not matched"`.
- Current result: `"null value"`.

The incorrect result originates from an error in Kùzu's own simple-`CASE` rewrite inherited by NeuG. The rewrite checks whether the `WHEN null` expression itself is null, producing a condition equivalent to `null IS NULL`, instead of checking whether the `CASE` expression is null. The generated condition is therefore always true, even when the `CASE` expression is not null.

Two separate issues contribute to this behavior:

1. NeuG does not support constant folding for `CASE` expressions, but they are currently classified as constant expressions and sent through the constant-folding path. `CASE` should therefore be removed from the set of expressions considered constant.
2. The simple-`CASE` rewrite inherited from Kùzu uses the wrong operand for a null `WHEN` expression:

   ```cpp
   // rewrite "CASE a.age WHEN 1" as "CASE WHEN a.age = 1"
   if (ExpressionUtil::isNullLiteral(*boundWhen)) {
     boundWhen = bindNullOperatorExpression(ExpressionType::IS_NULL,
                                            expression_vector{boundWhen});  // Should use boundCase instead.
   } else {
     boundWhen = bindComparisonExpression(
         ExpressionType::EQUALS, expression_vector{boundCase, boundWhen});
   }
   ```

   It should apply `IS NULL` to the `CASE` expression instead.

Spockkk0225:
## Summary

- Stop classifying `CASE` expressions as constants because the constant evaluator does not support `CASE` evaluation.
- Fix a simple-`CASE` rewrite bug inherited from Kùzu by applying `IS NULL` to the `CASE` expression instead of the `WHEN NULL` expression itself.
- Ensure `CASE NULL WHEN NULL` selects the null branch while `CASE 1 WHEN NULL` falls through to `ELSE`.

## Tests

- Added regression coverage for both matching and non-matching simple `CASE` expressions with `WHEN NULL`.

## Docs

- Added documentation for simple and searched `CASE` expressions, `WHEN NULL` behavior, and optional `ELSE` clauses.
- Enabled the previously commented-out `CASE` entry in the expression documentation index.

Closes #1028


doc/source/cypher_manual/expression/case_expr.md
```diff
@@ -0,0 +1,70 @@
...
+END;
+```
+
+Only a condition that evaluates to `TRUE` matches. Conditions that evaluate to
+`FALSE` or `NULL` do not match. Use `IS NULL` to test for a `NULL` value
+explicitly:
+
+```cypher
+RETURN CASE
+    WHEN expression IS NULL THEN 'null value'
```
shirly121, line 63: 如果是
`RETURN CASE WHEN expression = NULL THEN .. ELSE`
expression = NULL 永远都是 false 走 else 分支吗？
建议确认这个行为并且在文档和测试中添加相关内容

Spockkk0225: 确认了，如果写成expression = NULL，那么这边会走三值逻辑得到NULL，导致不会匹配而走ELSE分支。
已补充测试&文档

shirly121 (approved):

PR #1040, opened 2026-09-08
fix: correct DISTINCT semantics for NULL values

### LINKED ISSUE #964: Multi-column DISTINCT and semi-join merge NULL with valid INT32 -1 (encode_value collision)
## Problem

`DISTINCT` on multiple columns merges a NULL key with a valid `INT32 -1` value — the same `encode_value()` collision pattern as #906, but in the dedup/join operators instead of group-by.

## Repro (verified on pre-#907 build)

```cypher
CREATE NODE TABLE source(id INT64, bucket INT32, PRIMARY KEY(id));
CREATE NODE TABLE target(id INT32, PRIMARY KEY(id));
CREATE REL TABLE links(FROM source TO target);
CREATE (:source {id: 1, bucket: 5}), (:source {id: 2, bucket: 5}), (:target {id: -1});
MATCH (source:source), (target:target) WHERE source.id = 1
CREATE (source)-[:links]->(target);

MATCH (source:source)
OPTIONAL MATCH (source)-[:links]->(target:target)
RETURN DISTINCT target.id, source.bucket;
```

Current result:

```text
-1 | 5
```

Expected result:

```text
-1   | 5
NULL | 5
```

## Root cause

`encode_value()` (src/common/types/value.cc:852) encodes NULL as `put_int(-1)` — byte-identical to a valid `INT32 -1` (also `put_int(-1)`). The multi-column dedup path (src/execution/common/operators/retrieve/dedup.cc:44) and the semi-join right-side build path (src/execution/common/operators/retrieve/join.cc:47) feed values straight into `encode_value()` with no validity handling, so NULL and `INT32 -1` produce identical encoded keys.

Notes:
- Single-column `DISTINCT` is not affected: it takes the `generate_dedup_offset()` fast path (dedup.cc:36), which respects the column's validity bitmap.
- `INT64 -1` does not collide with NULL (`put_long` vs `put_int` encodings differ in width); only `INT32` values collide.
- PR #907 fixes the group-by paths (GKey bitmap + optional-aware typed-key routing); the same treatment is needed here.

## Suggested fix

Mirror the #907 approach in `Dedup::dedup` and the join build path: check `Value::IsNull()` per component and encode an explicit validity marker alongside each value.

### LINKED ISSUE #1030: Fix DISTINCT semantics for projected rows and aggregate inputs with NULL
## `RETURN DISTINCT`

`RETURN DISTINCT` removes duplicate rows using all projected columns. Rows whose projected values are identical are retained only once, including rows with `null` in the same projected positions.

```cypher
UNWIND [null, null, 1, 1] AS value
RETURN DISTINCT value;
// null, 1
```

### Current Issue

```cypher
MATCH (source:source)
OPTIONAL MATCH (source)-[:links]->(target:target)
RETURN DISTINCT target.id, source.bucket;
```

In the reproduction dataset, the projected results contain two distinct rows, `(-1, 5)` and `(null, 5)`:

- Expected result: retain both `(-1, 5)` and `(null, 5)`.
- Current result: only `(-1, 5)` is returned.

`RETURN DISTINCT` incorrectly treats different projected rows as duplicates. `null` must not be confused with the actual value `-1`.

## `WITH DISTINCT`

`WITH DISTINCT` follows the same row-deduplication rules as `RETURN DISTINCT`, but performs deduplication in the middle of the query pipeline. Subsequent clauses receive only the deduplicated rows.

```cypher
UNWIND [null, null, 1, 1] AS value
WITH DISTINCT value
RETURN value;
// null, 1
```

## `DISTINCT` in Aggregate Arguments

`DISTINCT` in an aggregate argument deduplicates only the input values for that aggregate. After deduplication, the aggregate applies its own null semantics; for example, both `count(expr)` and `collect(expr)` exclude `null`.

```cypher
UNWIND [null, null, 1, 1, 2] AS value
RETURN count(DISTINCT value), collect(DISTINCT value),
       sum(DISTINCT value), avg(DISTINCT value);
// 2, [1, 2], 3, 1.5
```

The null values do not contribute to the aggregate results, and the duplicate value `1` contributes only once.

### Current Issue

```cypher
UNWIND [null, null, 1, 1, 2] AS value
RETURN sum(DISTINCT value), avg(DISTINCT value);
```

| Expression | Expected result | Current result |
|---|---|---|
| `sum(DISTINCT value)` | Error: not supported | `4` |
| `avg(DISTINCT value)` | Error: not supported | `1.3333333333333333` |

The current function architecture does not support applying the `DISTINCT` operator to `SUM` or `AVG`. These forms should return an explicit unsupported error instead of silently ignoring `DISTINCT` and producing an incorrect result.

Spockkk0225:
## Summary

- Distinguish NULL values from encoded non-NULL values when deduplicating projected rows.
- Route optional columns through the NULL-aware multi-column deduplication path so `RETURN DISTINCT` and `WITH DISTINCT` retain genuinely different rows.
- Reject unsupported `SUM(DISTINCT ...)` and `AVG(DISTINCT ...)` expressions explicitly instead of silently ignoring `DISTINCT` and returning incorrect results.

## Tests

- Added coverage for deduplicating projected rows containing both NULL and the real value `-1`, including single- and multi-column projections.
- Added parameterized coverage for rejecting distinct `SUM` and `AVG` inputs, both with and without NULL values.
- Built the Python binding and ran all five added test cases successfully.

## Docs

- The aggregate-function reference already marks `SUM` and `AVG` as not supporting `DISTINCT`, so no documentation change is needed.

Closes #1030
closes #964


tools/python_bind/tests/test_query.py
```diff
@@ -144,6 +144,73 @@ def test_aggregate_over_all_null_input(empty_db):
...
 
+@pytest.mark.parametrize("aggregate", ["sum", "avg"])
+@pytest.mark.parametrize(
+    "values",
+    [
+        "CAST([1, 1, 2], 'INT64[]')",
+        "CAST([CAST(null, 'INT64'), CAST(null, 'INT64'), 1, 1, 2], 'INT64[]')",
+    ],
+)
+def test_distinct_aggregate_not_supported(empty_db, aggregate, values):
```
shirly121: 建议加一组测试验证 count/collect/min/max 和 distinct 可以一起使用

tools/python_bind/tests/test_query.py
```diff
@@ -144,6 +144,73 @@ def test_aggregate_over_all_null_input(empty_db):
...
+    _, conn = empty_db
```
shirly121, line 149: 除了distinct null 之外，aggregate 函数 count/collect/min/max/avg/sum 对 NULL 值处理也需要明确一个标准，建议单独 issue 和 PR。

Spockkk0225: aggregation function对null的语义已经在这个issue统一：
https://github.com/alibaba/neug/issues/890

tools/python_bind/tests/test_query.py
Spockkk0225: 重新补齐了所有aggregation的测试

liulx20 (approved):

PR #1044, opened 2026-09-09
fix: order Carquet adapters after protobuf generation

liulx20:
## What do these changes do?

A clean parallel extension build can compile `neug_parquet_carquet_impl` before protobuf generation finishes. Its input/output stream headers include `neug/utils/result.h`, which requires the generated `neug/generated/proto/plan/error.pb.h`, causing a missing-header compilation failure.

Add an explicit dependency on `neug_proto` so generation completes before the Carquet adapters compile. This changes only one line in the Parquet CMake configuration.

### Validation

A focused clean CMake build using the adapter target definition and the actual adapter sources reproduced the missing-header failure without the dependency. With the dependency, it generated `error.pb.h` and compiled both adapter sources successfully with `-j4`. Full extension CI remains to be verified.

## Related issue number

Observed in [extension CI](https://github.com/alibaba/neug/actions/runs/34305275798/job/102322246010?pr=1039). The same missing dependency exists on main; this fix is submitted separately from #1039.


zhanglei1949 (approved):
LGTM

PR #573, opened 2026-06-18
docs: standardize table/label naming to LDBC/Neo4j convention and fix Cypher keyword casing

longbinlai:
## Summary

Fixes #572

This PR standardizes table/label naming and Cypher keyword casing across all NeuG documentation to eliminate inconsistencies that could confuse users.

## Naming Convention (LDBC/Neo4j Style)

| Element | Convention | Examples |
|---------|-----------|----------|
| **Node labels** | PascalCase | `Person`, `Software`, `Company`, `User` |
| **Relationship labels** | UPPERCASE_SNAKE | `KNOWS`, `CREATED`, `WORKS_FOR`, `FRIENDS` |
| **Cypher keywords** | ALL-CAPS | `MATCH`, `RETURN`, `WHERE`, `ORDER BY`, `LIMIT` |
| **DDL syntax** | Full keywords | `CREATE NODE TABLE`, `CREATE REL TABLE` |

### Label Naming (Node → PascalCase, Rel → UPPERCASE_SNAKE)

**Cypher Manual:**
- `match_clause.md`, `where_clause.md`, `return_clause.md`, `order_clause.md`, `limit_clause.md`, `with_clause.md`, `union_clause.md`, `merge_clause.md`, `dml_clause.md`, `index.md`
- `person` → `Person`, `software` → `Software`
- `knows` → `KNOWS`, `created` → `CREATED`
- Output tables updated (`_LABEL`, `_SRC_LABEL`, `_DST_LABEL`)

**Data I/O:**
- `import_data.md`: Unified all examples to `Person`/`KNOWS`/`User`/`FRIENDS`/`WORKS_FOR`
- `export_data.md`: `person` → `Person`, `knows` → `KNOWS` (including JSON output)
- `load_data.md`: Updated label references

**Other Docs:**
- `transaction.md`: `WorksAt` → `WORKS_AT`
- `data_types.md`: All label references in tables and examples
- `extensions/load_httpfs.md`: `Person` in COPY TO examples

**Python API Reference:**
- `database.md`, `connection.md`: `CREATE TABLE` → `CREATE NODE TABLE`/`CREATE REL TABLE`

**Tutorials:**
- `benchmark-neug-dual-mode.md`: Labels in descriptive text

### Cypher Keyword Casing (Title Case → ALL-CAPS)

[...]

PR #543, opened 2026-06-12
refactor: replace iterator with cursor-based HasNext/Next/Get API

liulx20:
Fixes

PR #1051, opened 2026-09-10
docs: fix stale #543 iterator-API residue in doc sources

zhanglei1949:
## What do these changes do?

Fix stale documentation **sources** left behind by the #543 iterator→cursor `QueryResult` refactor. These are doc sources only (header comments, the doc generator, its config, and a Python docstring) — no functional code changes, and no generated docs are committed (rationale below).

- `include/neug/main/connection.h`, `include/neug/main/neug_db.h`: replace the pre-#543 `for (auto& record : result.value())` examples with the current cursor API (`hasNext()` / `next()` / `GetString(...)` / `GetCurrentRowAsString()`). Also expand the `neug_db.h` file-locking note to match the already-published wording: a read-write open is exclusive, while multiple read-only processes/instances can share the directory.
- `doc/source/_scripts/generate_cpp_docs.py`: update the Quick Start example templates to the cursor API and fix a struct emission bug (`content += "}"` → `"};"`, which previously produced invalid C++ in `_meta.ts`).
- `doc/source/_scripts/cpp_key_classes.json`: drop the retired `neug::RecordLine` key class and its dangling `include_classes` reference; align the `QueryResult` description to "cursor-based access".
- `tools/python_bind/neug/session.py`: correct the `Session.execute()` docstring — the default `access_mode=""` infers the mode from the query text, so remove the inaccurate "(default)" on update mode and document the inference; move the transaction-behavior note out of the `:return:` block so it renders.

## Why no generated docs are included

The published docs under `doc/source/reference/` are **hand-curated and already correct** — several were fixed directly in #573/#892/#1003/#1025 without updating the sources. Re-running the generator (`make html` / `make api-docs`) produces a **mix of catch-up and regressions**, and the regressions would ship broken docs:

- flattens the hand-curated `cpp_api/query_result.md` — #1025's five sections (`Cursor Traversal` / `Typed Value Accessors` / `Metadata` / `Other Methods` / `Example`) collapse into a single `Public Methods` list;
- deletes `cpp_api/service.md`'s hand-written `ExecutionSlot` description ("runtime-neutral execution context…"), which lives in no header and therefore cannot be reproduced by generation;
- in the Python docs, reverts #573's label casing (`Person` → `person`), breaks examples into invalid Cypher (`-[:KNOWS]->` → `-[knows]->`), and injects raw Sphinx `:meth:` roles that don't render on the docs site.

So this PR deliberately fixes the **sources only** (a strict improvement, zero published-doc regression). Bringing the generated docs back in sync — correcting the stale Python docstrings and reconciling the hand-curated C++ pages so the generator reproduces the good output — is left as a separate follow-up.

## Related issue number

N/A — follow-up cleanup to #543.

### DIFF
--- a/doc/source/_scripts/cpp_key_classes.json
+++ b/doc/source/_scripts/cpp_key_classes.json
@@ -9,8 +9,7 @@
     "neug::ExecutionSlot",
     "neug::TpExecutionSlotPool",
     "neug::ExecutionSlotLease",
-    "neug::QueryResult",
-    "neug::RecordLine"
+    "neug::QueryResult"
   ],
   "class_files": {
     "neug::NeugDB": {
@@ -26,8 +25,7 @@
     "neug::QueryResult": {
       "filename": "query_result",
       "title": "QueryResult",
-      "description": "Container for query results with iterator access",
-      "include_classes": ["neug::RecordLine"]
+      "description": "Container for query results with cursor-based access"
     },
     "neug::NeugDBService": {
       "filename": "service",
--- a/doc/source/_scripts/generate_cpp_docs.py
+++ b/doc/source/_scripts/generate_cpp_docs.py
@@ -1302,8 +1302,10 @@ int main() {
 
   // Process results
   if (result.has_value()) {
-    for (auto& record : result.value()) {
-      std::cout << record.ToString() << std::endl;
+    auto& qr = result.value();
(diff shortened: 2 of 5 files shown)


liulx20 (approved):

longbinlai (approved):
LGTM

PR #1067, opened 2026-09-14
refactor: remove unused compiler Arrow code

### LINKED ISSUE #1063: [BUG] ArrowBuffer::reserveInternal does not check realloc/malloc (leak + null write)
**Describe the bug**

`ArrowBuffer::reserveInternal` (`include/neug/compiler/common/arrow/arrow_buffer.h:89-96`) never checks the allocator return value:

```cpp
if (dataptr) {
  dataptr = (uint8_t*) realloc(dataptr, bytes);   // L91, unchecked
} else {
  dataptr = (uint8_t*) malloc(bytes);             // L93, unchecked
}
capacity = bytes;                                  // L95, unconditional
```

- On `realloc` failure the original block stays allocated but `dataptr` is overwritten with `nullptr`; the destructor (L34-42) early-returns on `!dataptr`, so the old buffer leaks permanently.
- `capacity = bytes` runs unconditionally, so after a failure the state is `dataptr == nullptr && capacity == bytes`, and `resize(bytes, value)` (L72-78) then writes `dataptr[i] = value` through a null pointer.

Currently latent in-repo: outside the arrow module nothing constructs `ArrowRowBatch`/`ArrowBuffer` (only the forward/friend declarations at `include/neug/compiler/common/types/value/value.h:44,54`), but the code is compiled into `neug_common_arrow` and affects downstream consumers.

**Expected behavior**

On allocation failure keep the previous state (`dataptr`/`capacity` unchanged, no leak) and report the error.

**Suggested fix**

```cpp
uint8_t* new_ptr = dataptr ? (uint8_t*) realloc(dataptr, bytes)
                           : (uint8_t*) malloc(bytes);
if (new_ptr == nullptr) { throw std::bad_alloc(); }
dataptr = new_ptr;
capacity = bytes;
```

**Environment**

- NeuG version: master (`169970b`); platform-independent code defect

shirly121:
## Summary

- Remove the unreachable Kuzu-derived Arrow converter, row batch, buffer, and null-mask tree implementation.
- Remove the orphaned Arrow CMake configuration and stale forward/friend declarations.
- Retain `arrow.h`, which defines the Arrow C Data Interface structures used by the active Python `to_arrow()` export path.

## Testing

- `cmake --build build --target neug_py_bind -j$(sysctl -n hw.ncpu)`
- `python3.9 -m pytest -q tests/test_to_arrow.py` (11 passed, 3 skipped)
- `git diff --check`

Fixes #1063


longbinlai (approved):
LGTM

PR #1075, opened 2026-09-14
fix: preserve null vertices in column unions and reuse INVALID_VID

liulx20:
## What do these changes do?

Preserve valid vertices and NULL rows when `SLVertexColumn::union_col()` combines nullable columns. The same-label path previously reversed the left-hand NULL check, replacing valid vertices with NULL and passing NULL IDs to a builder that requires valid IDs. The multi-label path also passed NULL records directly to the non-null builder, which could assert or index the label bitmap with an invalid label.

Route NULL rows through `push_back_null()` in both paths and reuse the existing `constexpr INVALID_VID` throughout vertex/edge columns, the affected retrieval operators, and their tests. This introduces no new builder interface or sentinel value.

Add regression coverage for all left/right nullable combinations, including a single-label left column combined with different-label, multi-segment, and multi-label right columns. Check row order, valid IDs, NULL positions, labels, and optional-column metadata.

Validation:
- 19 vertex-column tests passed on the PR sources, including the new same-label and multi-label union cases. Compiled the column sources separately against existing support libraries; the full application test suite was not run.
- Compiler syntax checks passed for `edge_expand.cc`, `sink.cc`, and the vertex operator consuming `get_v.h`.
- Changed-line clang-format checks and `git diff --check` passed.

## Related issue number

No linked issue.


zhanglei1949 (approved):
LGTM

PR #1077, opened 2026-09-14
test: assert NULLS LAST for ascending order

### LINKED ISSUE #1076: Fix ORDER BY result validation in the existing Parquet list/array test
### Problem

The Parquet LIST/ARRAY regression test `test_parquet_list_and_array_preserve_null_elements` was made order-insensitive after its previous expected result assumed input order even though the query used `ORDER BY value`.

The current ascending sort behavior is deterministic: all non-NULL values are ordered first, and NULL values are placed at the end.

### Expected behavior

Ascending `ORDER BY` should consistently use NULLS LAST. For input values `[1, NULL, 3, 4, 5]`, `ORDER BY value` should return `[1, 3, 4, 5, NULL]`.

### Proposed change

Restore `ORDER BY value` in `test_parquet_list_and_array_preserve_null_elements` and assert the ordered result explicitly. This documents and protects the current ascending NULL placement semantics.

Spockkk0225:
## What do these changes do?

- Restore `ORDER BY value` in `test_parquet_list_and_array_preserve_null_elements`.
- Assert the actual ascending order: non-NULL values first and NULL at the end.
- Document and protect the current `ORDER BY ... ASC` NULLS LAST behavior.

## Related issue number

Fixes #1076

## Validation

- `python3 -m black .`
- `git diff --check`
- The Parquet extension test was not run locally because `third_party/carquet` is unavailable in this checkout; CI should validate it.


longbinlai (approved):
LGTM

PR #1078, opened 2026-09-14
fix: preserve input offsets when triangle candidates are empty

liulx20:
## What do these changes do?

Keep triangle-counting output offsets aligned with the original input rows when an input vertex has no qualifying first-hop neighbors. The bundled multiple-mutable CSR fast paths for both LT and GT skipped the manual `++idx` on `continue`. A match from the next input row therefore carried an earlier row's offset, causing `set_with_reshuffle()` to associate it with the wrong input values.

For example, if input row 0 has no candidates and row 1 produces a triangle, the output must use offset 1 rather than 0.

Use indexed loops in all four LT/GT traversal branches so the loop increment runs even when a vertex is skipped. Candidate filtering and triangle matching are unchanged.

Validation:
- Compiler syntax check of `tc_fuse.cc`, using the PR source and headers with the existing dependency build, passed.
- Changed-region clang-format check and `git diff --check` passed.
- Runtime regression tests and the full application suite were not run.

## Related issue number

No linked issue.


zhanglei1949 (approved):
LGTM

PR #1079, opened 2026-09-15
fix: build and repair release extensions with cibuildwheel

### LINKED ISSUE #1080: bug: NeuG 0.2.0 extensions fail to load due to missing libmimalloc.so.2
## Description

The extension packages published for NeuG 0.2.0 cannot be loaded in some wheel installations because their runtime dependencies are not packaged consistently with the NeuG wheel.

Loading the `vector_search`, `fts`, or `parquet` extension may fail because `libmimalloc.so.2` cannot be found.

## Affected version

- NeuG wheel: `0.2.0`
- Extensions: `vector_search`, `fts`, and `parquet`

## Error

```text
Library not loaded: libmimalloc.so.2
```

Depending on the build environment and platform, the failure may also appear as an undefined-symbol error or as a mismatch between the dependency names expected by the extension and those bundled in the wheel.

## Root cause

The `release-extension` workflow builds extension packages through a separate build script and environment. Unlike the NeuG wheel workflow, the generated extension packages do not pass through the same cibuildwheel build and repair process (`auditwheel` on Linux or `delocate` on macOS).

This creates two compatibility risks:

1. The extension and the NeuG wheel may be compiled in different images or toolchain environments, resulting in undefined symbols at load time.
2. Extension dependencies such as mimalloc may not be bundled or rewritten consistently, so the library names and runtime paths can differ from those in the wheel.

## Expected behavior

Extensions published for a NeuG release should load successfully with the wheel from the same release without requiring users to install or manually patch additional shared libraries.

## Required fix

- Build release extensions through cibuildwheel using environments compatible with the official NeuG wheel build.
- Run the resulting packages through the same platform repair process used by wheel releases.
- Extract and publish extension binaries only after repair.
- Validate the repaired `vector_search`, `fts`, and `parquet` extensions by loading them with the published NeuG 0.2.0 wheel on every supported platform.

shirly121:
## Summary

- build release extension packages through cibuildwheel so they use the same build and repair environment as NeuG wheels
- extract repaired extension binaries and checksums from the carrier wheel
- support Linux x86_64, Linux ARM64, and macOS ARM64 while dropping unsupported macOS x86_64
- avoid OSS uploads for ARM64 and macOS jobs, and publish only extension binaries and SHA256 files as workflow artifacts
- make the ARM runner setup and generated build environment robust

## Validation

- verified the release workflow YAML parses successfully
- verified repaired vector_search and fts extensions can be loaded with the NeuG 0.2.0 wheel
- merged the latest upstream main before opening this PR

Fixes #1080


longbinlai (approved):
LGTM

PR #1091, opened 2026-09-16
fix: disambiguate batch_size byte semantics with unified batch_rows row-count option

### LINKED ISSUE #846: [BUG] Ambiguous `batch_size` semantics: documented as bytes, but CSV/JSON readers treat it as a row count
**Describe the bug**

The generic read option `batch_size` (`ReadOptions::batch_size`, default `1 << 20`) has inconsistent semantics across readers and documentation:

1. **Documentation** (`doc/source/data_io/import_data.md`) describes it as an I/O batch size **in bytes** ("Batch size in bytes when `batch_read` is enabled", default 1 MiB).
2. **Original design** (commit `039d7a2a`, Arrow CSV reader) mapped it to Arrow's `block_size`, i.e. **bytes**. The Parquet extension still follows this: `batch_size` is passed to `ReaderProperties::set_buffer_size()` (bytes).
3. **Native CSV/JSON readers** (`src/utils/io/read/common/options.cc`) instead interpret it as a **row count**, deriving the row chunk size via `std::min(batch_size, 65536)`.

So the same option means *bytes* for Parquet and *rows* (clamped to 65536) for CSV/JSON, while the documentation only describes the byte interpretation. This inconsistency was discovered while fixing #839.

**Execution Logs**

No runtime error; the inconsistency is in code/docs. Relevant locations:

1. `doc/source/data_io/import_data.md`: documents `batch_size` as bytes, default `1048576`.
2. `extension/parquet/src/parquet_options.cc`: `reader_properties->set_buffer_size(readOpts.batch_size.get(options))` — bytes.
3. `src/utils/io/read/common/options.cc`: `config.chunk_size = std::min<int64_t>(batch_size, 65536)` — row count for CSV/JSON.

**Expected behavior**

`batch_size` should have one well-defined meaning across all readers and the documentation, e.g.:

- Option A: keep the documented byte semantics everywhere; CSV/JSON chunk sizing gets its own explicit row-count option.
- Option B: unify on row count across all readers and update the documentation accordingly.

Whichever direction is chosen, the documentation, default values, and reader implementations must agree, and the change should be called out in release notes since it alters user-visible behavior for at least one format.

**Error Message**

N/A (semantic/documentation inconsistency, not a runtime error).

BingqingLyu:
## Related Issues
fix #846

## What does this PR do?
Fixes the ambiguous `batch_size` semantics reported in #846: the option is documented as a byte count, but the CSV/JSON readers treated it as a row count. This PR adds a unified `batch_rows` option for row-count-based batching and keeps `batch_size` strictly as a bytes-only I/O option.

## What changes in this PR?
- Add a generic `batch_rows` option (number of rows per batch/chunk, default 65536) to `ReadOptions`, consumed by CSV, JSON and Parquet readers.
  - CSV/JSON: chunk row count, clamped to [1, 65536]; non-positive values fall back to 4096.
  - Parquet: Arrow record batch size (via `ArrowReaderProperties::batch_size()`).
- Keep `batch_size` (default 1 MiB) as a bytes-only option; it now only controls the Parquet buffered-stream buffer size, and the CSV/JSON readers no longer consume it.
- Keep `PARQUET_BATCH_ROWS` as a deprecated alias of `batch_rows` for backward compatibility. The fallback is implemented on the Parquet side (both the Arrow path and the carquet scan path), leaving the core `Option::get` API untouched; `batch_rows` takes precedence when both are set.
- Update documentation (`import_data.md`, `load_data.md`, `load_parquet.md`, `carquet_reader.md`) and tests (C++ reader/parquet tests, Python data I/O and load tests).


zhanglei1949 (approved):
LGTM

PR #1095, opened 2026-09-16
fix: register httpfs extension in SHOW_LOADED_EXTENSIONS

### LINKED ISSUE #1094: [BUG] httpfs extension missing from SHOW_LOADED_EXTENSIONS() after LOAD httpfs
**Describe the bug**

After `LOAD httpfs`, the httpfs extension does **not** appear in the result of
`CALL SHOW_LOADED_EXTENSIONS() RETURN *`, even though the extension is loaded
and fully functional (its `s3`/`oss`/`http`/`https` VFS providers are
registered). Loading other extensions such as `fts` (or `parquet`, `gds`,
`vector_search`, `pattern_matching`) does show them in the result.

**Root cause**

`CALL SHOW_LOADED_EXTENSIONS()` reads the global map
`ExtensionAPI::loaded_extensions_`, which is only populated when an extension's
`Init()` calls `ExtensionAPI::registerExtension(...)`.

Every other in-tree extension calls `registerExtension` in its `Init()`, but
`extension/httpfs/src/s3_extension.cc` does not — its `Init()` only initializes
libcurl and registers the VFS filesystem providers. As a result httpfs is never
added to `loaded_extensions_`, so it is absent from `SHOW_LOADED_EXTENSIONS()`.

| Extension | Calls `registerExtension` in `Init()` |
|-----------|----------------------------------------|
| fts | yes |
| parquet | yes |
| gds | yes |
| vector_search | yes |
| pattern_matching | yes |
| **httpfs** | **no** |

**Execution Logs**

Commands that reproduce the issue:

1. `LOAD httpfs;`
2. `CALL SHOW_LOADED_EXTENSIONS() RETURN *;`
   -> result does not contain a `httpfs` row.
3. For comparison: `LOAD fts;` then `CALL SHOW_LOADED_EXTENSIONS() RETURN *;`
   -> result contains `fts`.

**Expected behavior**

After `LOAD httpfs`, `CALL SHOW_LOADED_EXTENSIONS() RETURN *` should include a
row for `httpfs`, consistent with the other extensions.

**Error Message**

No exception is thrown — the symptom is a silently missing row in the
`SHOW_LOADED_EXTENSIONS()` output for httpfs.

**Proposed fix**

In `extension/httpfs/src/s3_extension.cc` `Init()`, after registering the VFS
providers, add:

```cpp
neug::extension::ExtensionAPI::registerExtension(
    neug::extension::ExtensionInfo{
        "httpfs",
        "Provides remote filesystem access for s3, oss, http, https schemes."});
```

and include `neug/compiler/extension/extension_api.h`. A regression test is
added to `extension/httpfs/tests/http_test.cc`.

BingqingLyu:
## Related Issues
fix #1094

## What does this PR do?

Fixes a bug where the `httpfs` extension did not appear in the result of
`CALL SHOW_LOADED_EXTENSIONS()` after `LOAD httpfs`, while other extensions
(`fts`, `parquet`, `gds`, `vector_search`, `pattern_matching`) did.

`SHOW_LOADED_EXTENSIONS()` reads the global `ExtensionAPI::loaded_extensions_`
map, which is only populated when an extension's `Init()` calls
`ExtensionAPI::registerExtension(...)`. The httpfs `Init()` only registered its
`s3`/`oss`/`http`/`https` VFS providers and never called `registerExtension`,
so httpfs was silently missing from the output even though it was loaded and
functional.

## What changes in this PR?

- `extension/httpfs/src/s3_extension.cc`
  - Add `#include "neug/compiler/extension/extension_api.h"`.
  - In `Init()`, after `RegisterHTTPProvider()`, call
    `ExtensionAPI::registerExtension({"httpfs", "Provides remote filesystem
    access for s3, oss, http, https schemes."})`.
- `tools/python_bind/tests/test_load.py`
  - Add an end-to-end regression test
    (`test_httpfs_appears_in_show_loaded_extensions`) that runs `LOAD httpfs`
    then `CALL SHOW_LOADED_EXTENSIONS() RETURN *` and asserts `httpfs` is
    present in the returned names. Marked `@extension_test` (offline, no
    network access required).

## Verification

- Built the Python binding and the `httpfs` extension locally.
- The new test **passes** with the fix applied and **fails** (`httpfs missing
  from SHOW_LOADED_EXTENSIONS(); got []`) when the `registerExtension` call is
  removed, confirming it captures the regression.
- `black`/`isort`/`flake8` clean on `test_load.py`; `clang-format` clean on
  `s3_extension.cc`.


lnfjpt (approved):
LGTM

PR #1097, opened 2026-09-17
fix: handle namespace predicates in variable-length paths

### LINKED ISSUE #1093: USE NAMESPACE with variable-length path causes SIGSEGV
## Summary

`USE NAMESPACE` combined with a variable-length path pattern causes the embedded Python process to segfault. The equivalent variable-length path without a namespace works, and fixed-length expansions inside the same namespace work.

## Environment

- NeuG `main`: `b7225d4e5ea5caf586655245603091e746a6134c`
- NeuG version reported by the build: `0.2.0`
- Release build
- macOS arm64
- Python 3.9

## Minimal reproduction

```python
import tempfile
from pathlib import Path

from neug import Database

tmp = tempfile.mkdtemp(prefix="neug_ns_varlen_")
db = Database(db_path=str(Path(tmp) / "db"), mode="w")
conn = db.connect()

conn.execute("CREATE NODE TABLE onto(uri STRING PRIMARY KEY, scope_id STRING)")
conn.execute("CREATE REL TABLE relate(FROM onto TO onto)")

for i in range(6):
    scope = "s0" if i < 4 else "s1"
    conn.execute(
        "CREATE (:onto {uri:$uri, scope_id:$scope})",
        parameters={"uri": f"u{i}", "scope": scope},
    )

for i in range(5):
    conn.execute(
        "MATCH (a:onto {uri:$a}), (b:onto {uri:$b}) "
        "CREATE (a)-[:relate]->(b)",
        parameters={"a": f"u{i}", "b": f"u{i + 1}"},
    )

conn.execute(
    "CALL project_graph("
    "'ns0', {'onto': 'n.scope_id = \"s0\"'}, "
    "['[onto, relate, onto]'])"
)

# SIGSEGV here.
list(
    conn.execute(
        "USE NAMESPACE ns0 "
        "MATCH (a:onto {uri:'u0'})-[:relate*1..2]-(x) "
        "RETURN DISTINCT x.uri"
    )
)
```

Running the reproducer directly exits with status 139. Running it through Python's `subprocess` reports return code `-11`. It reproduced in three consecutive runs.

## Expected behavior

The query should return the namespace-filtered endpoints (for this graph, `u0`, `u1`, and `u2`), or return a normal unsupported-query error if predicate-aware variable-length expansion is not supported. It should not terminate the host process.

## Actual behavior

The process terminates with `SIGSEGV` while executing the namespaced variable-length path query.

## Controls

| Query variant | Result |
| --- | --- |
| Variable-length `*1..2`, no namespace | succeeds: `u0`, `u1`, `u2` |
| `USE NAMESPACE ns0` + variable-length `*1..2` | **SIGSEGV** |
| Variable-length `*1..2` with explicit endpoint `scope_id` predicates | succeeds |
| `USE NAMESPACE ns0` + fixed one-hop pattern | succeeds: `u1` |
| `USE NAMESPACE ns0` + explicit fixed two-hop pattern | succeeds: `u0`, `u2` |

## Notes / suspected area

The namespace binder adds projected node/relationship predicates to MATCH patterns. The failure appears specific to passing those generated predicates through variable-length path expansion; ordinary path expansion and fixed-hop namespace expansion both work.

Relevant areas appear to include namespace predicate construction in `src/compiler/binder/bind/bind_graph_pattern.cpp` and `PathExpandOprWithPred` / `edge_expand_p_with_pred` in the path execution code.

## Workarounds

- Run the variable-length path against the base graph and add explicit scope predicates.
- Within the namespace, replace the variable-length pattern with fixed one-hop/two-hop patterns when the maximum depth is known.

shirly121:
## Summary

- bind projected node and relationship predicates to recursive path variables
- avoid redundant label predicates when table IDs already determine the concrete label
- deduplicate generated namespace filtering predicates
- preserve predicate expression identity after dependent-variable renaming

## Tests

- reproduce the USE NAMESPACE variable-length path crash from issue #1093
- cover filtered and unfiltered scan, expand, path expand, and get-v plans
- cover specified, unspecified, and partially specified labels
- cover fixed-length and variable-length OPTIONAL MATCH queries
- verify projected queries return expected rows and avoid standalone FilterOpr operators

Fixes #1093


longbinlai (approved):
LGTM

